PR #162, opened 2014-01-03
Kindle specific view for the Content Server

terwey:
Hi,

this is still a Work in Progress but I wanted to get some feedback on this before.
This is a Kindle specific view for the Calibre Content Server because the mobile view was sadly a bit of a mess on my Kindle.

I started by taking the normal mobile.py and duplicate it to kindle.py and work my way from there.

On top of that I've decided to filter away formats that the Kindle cannot show and just display the one's it can.
It starts by doing a `format:mobi OR format:azw3` search query and then afterwards on URL generation for the Download again only showing the relevant formats.

I've also made it possible to directly search for an Author, Tag or Series through the use of normal hyperlinks.
And because sometimes you want to see a Summary (but not always) this is Toggled through a simple Javascript command (which the Kindle supports! hooray!)

I'd love to hear some feedback, I haven't done Python in a while so don't go completely bonkers on me ;)

### DIFF
--- a/resources/content_server/kindle.css
+++ b/resources/content_server/kindle.css
@@ -63,6 +63,11 @@ div.navigation {
 #listing td.thumbnail {
     height: 60px;
     width: 60px;
+    vertical-align: top;
+}
+
+#listing td.download {
+    vertical-align: top;
 }
 
 #listing tr.lastTr {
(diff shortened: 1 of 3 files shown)


kovidgoyal:
The new server is more or less ready (currently it only has a cover grid view, but the views are modular so adding more views should be easy in the future)

calibre-debug --new-server

to see it in action

Closing this PR as it is no longer relevant.

PR #350, opened 2014-12-23
news: Extend BasicNewsRecipe with ability to skip articles which were downloaded in the past.

spacekpe:
This function has to be explicitly activated by
ignore_downloaded_articles = True in recipe definition.

When activated, the recipe stores hashes of downloaded articles in file named
<config_dir>/'recipes'/'recipe_storage'/<recipe.title>/<feed.title>.
Articles which already has hash stored in the file are not downloaded again.

The idea is taken from
http://www.mobileread.com/forums/showpost.php?p=1295505&postcount=10
but the code was changed significantly to improve integration.


kovidgoyal:
no followup

PR #352, opened 2015-01-02
Daily Cover from https://zeitung.sueddeutsche.de/webapp

lala-rob:
Now geting the daily Cover from zeitung.sueddeutsche.de


kovidgoyal:
This will not work if the person running the recipe is in a different time zone than the server. Instead you should use datetime.utcnow() + datetime.timedelta(hours=<whatever the offset is for germany>)

Of course, if germany uses daylight savings time, then it gets more complicated.

PR #361, opened 2015-01-29
Proposed solution for 1/2 of enhancement #1415714. Instead of making ...

cbhaley:
...quickview a pane, which I think would be very invasive, these changes make the quickview window always on top of calibre but not other application windows. In addition, minimizing calibre will minimize quickview and vice versa. I have tested this change on windows.

This is clearly a behavior change, but I think it is for the better.

Also note that I removed the minimize and maximize buttons. Minimize didn't really minimize anymore, and the window state was getting confused. This behavior change might trigger some unhappiness.


cbhaley:
Changes to remove "ignore" and call the superclass pushed.

PR #362, opened 2015-01-30
Another attempt at enhancement #1415714. The user has the choice of ...

cbhaley:
...putting quickview in its own window (original implementation) or as a splitter panel containing library view and the quickview panes.

### DIFF
--- a/src/calibre/customize/builtins.py
+++ b/src/calibre/customize/builtins.py
@@ -834,11 +834,6 @@ class ActionQuickview(InterfaceActionBase):
     actual_plugin = 'calibre.gui2.actions.show_quickview:ShowQuickviewAction'
     description = _('Show a list of related books quickly')
 
-class ActionQuickviewFocusTo(InterfaceActionBase):
-    name = 'Focus To Quickview'
-    actual_plugin = 'calibre.gui2.actions.show_quickview:FocusToQuickviewAction'
-    description = _('Move the focus to the Quickview pane/window')
-
 class ActionTemplateTester(InterfaceActionBase):
     name = 'Template Tester'
     actual_plugin = 'calibre.gui2.actions.show_template_tester:ShowTemplateTesterAction'
@@ -979,7 +974,7 @@ plugins += [ActionAdd, ActionFetchAnnotations, ActionGenerateCatalog,
         ActionAddToLibrary, ActionEditCollections, ActionMatchBooks, ActionChooseLibrary,
         ActionCopyToLibrary, ActionTweakEpub, ActionUnpackBook, ActionNextMatch, ActionStore,
         ActionPluginUpdater, ActionPickRandom, ActionEditToC, ActionSortBy,
-        ActionMarkBooks, ActionEmbed, ActionTemplateTester, ActionQuickviewFocusTo]
+        ActionMarkBooks, ActionEmbed, ActionTemplateTester]
 
 # }}}
 
--- a/src/calibre/customize/builtins.py
+++ b/src/calibre/customize/builtins.py
@@ -834,6 +834,16 @@ class ActionQuickview(InterfaceActionBase):
     actual_plugin = 'calibre.gui2.actions.show_quickview:ShowQuickviewAction'
     description = _('Show a list of related books quickly')
 
+class ActionQuickviewFocusTo(InterfaceActionBase):
+    name = 'Focus To Quickview'
+    actual_plugin = 'calibre.gui2.actions.show_quickview:FocusToQuickviewAction'
+    description = _('Move the focus to the Quickview pane/window')
+
+class ActionQuickviewFocusFrom(InterfaceActionBase):
+    name = 'Focus From Quickview'
+    actual_plugin = 'calibre.gui2.actions.show_quickview:FocusFromQuickviewAction'
+    description = _('Move the focus to the library pane')
+
 class ActionTemplateTester(InterfaceActionBase):
     name = 'Template Tester'
     actual_plugin = 'calibre.gui2.actions.show_template_tester:ShowTemplateTesterAction'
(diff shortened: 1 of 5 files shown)


kovidgoyal:
Looks fine to me. A couple of suggestions:

1) Have the quickview shortcut key toggle the panel visbility when it is embedded
2) Have the quickview window auto close and re-open itself when using the "Show as pane: checkbox, and make it a button instead of a check box.

cbhaley:
OK for the button instead of the checkbox.

As for the QV shortcut, right now it either opens the pane/window or moves the focus to QV's booklist. I did this because of users like BetterRed who want to avoid the mouse at almost all costs. Couple this with ESC moving the focus to the main library view and it becomes easy to switch back and forth. I think this behavior is important, but I am open for alternate implementations.

While speaking of focus, I am worried about the accelerators on the checkbox and buttons. When QV is docked the main calibre shortcut mechanism comes into play, meaning it is easy to overload an accelerator. I am thinking of removing all accelerators when docked, instead making TAB and SHIFT-TAB move between them. Do you have any thoughts on this?

kovidgoyal:
On Sun, Feb 01, 2015 at 12:24:32AM -0800, Charles Haley wrote:
> OK for the button instead of the checkbox.
> 
> As for the QV shortcut, right now it either opens the pane/window or moves the focus to QV's booklist. I did this because of users like BetterRed who want to avoid the mouse at almost all costs. Couple this with ESC moving the focus to the main library view and it becomes easy to switch back and forth. I think this behavior is important, but I am open for alternate implementations.

Funny, it was the same impulse that prompted my comment. Right now there
is no way to close the QV panel without using the mouse :)

Perhaps an additional shortcut, say Shift-Q to focus?

> 
> While speaking of focus, I am worried about the accelerators on the checkbox and buttons. When QV is docked the main calibre shortcut mechanism comes into play, meaning it is easy to overload an accelerator. I am thinking of removing all accelerators when docked, instead making TAB and SHIFT-TAB move between them. Do you have any thoughts on this?

Good point, there should indeed be no accelerators in a widget displayed
as part of the main window, otherwise the widget will break existing
users keyboard shortcuts.

Dont TAB/SHITF_TAB already move between controls?

cbhaley:
> Perhaps an additional shortcut, say Shift-Q to focus?

How do I add additional shortcuts? Do I make a new action, or can I have a second shortcut in an existing action. As far as I can tell from looking at the code, I will need to make another action.

> Dont TAB/SHITF_TAB already move between controls?

Yes, between *all* the controls on calibre's window. I want to 'trap' the tab key when the focus is in the QV pane, preventing TAB from moving out of the pane.

kovidgoyal:
On Sun, Feb 01, 2015 at 12:40:32AM -0800, Charles Haley wrote:
> 
> > Perhaps an additional shortcut, say Shift-Q to focus?
> 
> How do I add additional shortcuts? Do I make a new action, or can I have a second shortcut in an existing action. As far as I can tell from looking at the code, I will need to make another action.

You can assign multiple shortcuts to an action, use a tuple, like this:

action_spec = (..., ('Q', 'Shitft-Q'))

In the action handling code, you then check QApplcation.modifiers() to see
if shift is pressed and if so...

> 
> > Dont TAB/SHITF_TAB already move between controls?
> 
> Yes, between *all* the controls on calibre's window. I want to 'trap' the tab key when the focus is in the QV pane, preventing TAB from moving out of the pane.

Hmm, I'm somewhat ambivalent on that. On one hand, it would break
tabbing through the controls in the window, on the other hand, you
can't really tab through the controls anyway, since, the book list also
traps the tab key.

kovidgoyal:
Make that Shift+Q

cbhaley:
On 01/Feb/2015 09:02, Kovid Goyal <notifications@github.com> wrote:
> Make that Shift+Q

It seems that this technique "hard codes" the shift key, but there isn't any way in the shortcut editor to check that. In other words, the user can change the shortcuts to ('Q', 'R'), thus breaking it because the action won't see the SHIFT modifier.

I think adding a second action will be safer, unless you know some way of ensuring that the "alternate shortcut" always has the SHIFT modifier attached to it.

kovidgoyal:
You can simply, parse the shortcuts assigned to the action with QKeySequence. Then check which one keyboardModifiers() matches. The only way that would break is if the user chose to use two different primary keys for the shortcut and the alternate shortcut.

However, I am fine with using a second action. It's less fragile. I was just answering your question of whether one could assign multiple shortcuts :)

cbhaley:
Changes pushed.

kovidgoyal:
You dont need to create calibre level actions, instead in the genesis() method of the original QuickView action, just create a new QAction and register it, like this

```python
self.focus_action = QAction(self.gui)
self.gui.addAction(self.focus_action)
self.gui.keyboard.register_shortcut(...) # see keyboard.py for how to use register_shortcut
```

cbhaley:
Removed the calibre-level action.

One nit: I am not happy about how the focused item is shown in the QV widget. It is extremely hard to see (on my machine at least) which of the item box or the book list actually has the focus. The check box and buttons are easier to see. There is a slight change in the color, but it is so minor as to be indistinguishable. I tried to find a way to make it behave more like the library view (radical change of color), but I failed to figure it out. If you have the time to make a suggestion I will look further, otherwise it will have to stay like it is.

kovidgoyal:
Merged, with a few small fixes. Regarding focus highlighting, you mean you want to highlight the current cell that has focus in the table view? For that you would have to implement a custom delegate for the qheaderview and use setProperty('highlight_current_item', 150) on the table view.

cbhaley:
What I want to know is which widget (the items list or the table view) has the focus. As it is, there is almost no visual difference in either widget when the focus enters or leaves.

Perhaps highlighting the current cell (or row) will do the job. I will try that.

And sorry about leaving semicolons around. Comes with writing in both java and python. I also have no end of trouble remembering whether or not I should use camel case for identifiers. :)

kovidgoyal:
Dont worry, semi-colons are harmless. I have an automated tool that runs every time I save a file in my editor that nukes them for me, anyway.

PR #372, opened 2015-03-14
add bin/pdfimages from poppler to calibre packages

kfix:
Hello again Kovid!

Please consider adding the 48K binary [`pdfimages`](http://cgit.freedesktop.org/poppler/poppler/tree/utils/pdfimages.cc) from poppler to the calibre packages.
It has the same library requirements as the other poppler binaries being shipped.

I ask because my [plugin](http://github.com/kfix/calibre_plugin_djvumaker) uses it to list PDFs' image contents to decide whether they are scanned library books that should be further compressed to DJVU. The python functions in calibre to extract PDF images are way slower than `pdfimages`.

Currently, installing my plugin requires the user to install an external popper just to get `pdfimages`. Poppler has a few layers of dependencies (cairo, glib) which may be completely useless to the user (e.g. ones on Windows)

If `pdfimages` (which is already being built by the calibre CI bots) was included in the calibre package, the user would only have to install djvulibre and ghostscript, which have a smaller and shallow set of dependencies and more general value to any user.

Regards, joey

### DIFF
--- a/setup/installer/linux/freeze2.py
+++ b/setup/installer/linux/freeze2.py
@@ -67,7 +67,7 @@ arch = 'x86_64' if is64bit else 'i686'
 
 def binary_includes():
     return [
-    j(SW, 'bin', x) for x in ('pdftohtml', 'pdfinfo', 'pdftoppm')] + [
+    j(SW, 'bin', x) for x in ('pdftohtml', 'pdfinfo', 'pdftoppm', 'pdfimages')] + [
 
     j(SW, 'lib', 'lib' + x) for x in (
         'usb-1.0.so.0', 'mtp.so.9', 'expat.so.1', 'sqlite3.so.0',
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Wouldn't it be better to just add a function to the calibre poppler bindings to list the images in a PDF? That should have similar performance to pdfimages and avoid needing an extra binary. See podofo/doc.cpp. Or are you saying that poppler is much faster than podofo for this task?

kfix:
I do see a class [ImageExtractor](http://svn.code.sf.net/p/podofo/code/podofo/trunk/tools/podofoimgextract/ImageExtractor.cpp) that could be modified to only count images instead of extracting them. `podofoimgextract` appears to be as fast as `pdfimages` against a test PDF file.

I am not fluent in C++ or pyx so it will take me some time to write PDFDoc_image_count() and make another patch. Thanks for the pointer.

kovidgoyal:
Here you go: https://github.com/kovidgoyal/calibre/commit/bf152707b305359f69638f2c572cbba981077cff

kfix:
It seems to work, I just published next version of my plugin. Thanks!

PR #373, opened 2015-03-15
When editing on the spreadsheet, grow the edit box to the size of the ...

cbhaley:
...contents if possible.

Note that I am not particularly happy with this implementation, and especially with the magic numbers. However, I didn't find a better way to do it. In particular, updateItemGeometry doesn't seem suitable because it is called before the delegate has access to the text in the control so it doesn't know how wide to make the edit box.

### DIFF
--- a/src/calibre/gui2/library/delegates.py
+++ b/src/calibre/gui2/library/delegates.py
@@ -9,7 +9,8 @@ import sys
 
 from PyQt5.Qt import (Qt, QApplication, QStyle, QIcon,  QDoubleSpinBox, QStyleOptionViewItem,
         QSpinBox, QStyledItemDelegate, QComboBox, QTextDocument, QSize, QMenu, QKeySequence,
-        QAbstractTextDocumentLayout, QFont, QFontInfo, QDate, QDateTimeEdit, QDateTime)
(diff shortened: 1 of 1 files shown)


cbhaley:
BTW: see http://www.mobileread.com/forums/showthread.php?t=257283 for some explanation for why I attempted this.

cbhaley:
Comment so you see the push ...

cbhaley:
I prefer that the left edge of the editor widget be aligned with the left edge of the column. However, given that you don't and given that (I suspect) RtL language users would prefer it be aligned on the right edge, I will look at changing it.

kovidgoyal:
I agree that it is nice to have the edges aligned, and that should be the case when there is enough space to the right, but when there is not, I prefer to use all available space.

cbhaley:
New code pushed. See the comment in the commit.

kovidgoyal:
I have merged. One thing I noticed is that when editing a yes/no type column, the editor now appears to the side of the existing column decoration, whereas it used to cover it before. Is that deliberate, or did you just forget to handle that case? I think I prefer the old behavior, as having two possibly conflicting icons next to each other is confusing.

kovidgoyal:
FYI, I have tested the new code on both linux and OS X (not fully, only a couple of delegate types) it seems to work ok.

PR #375, opened 2015-03-17
Minor toolbar string fix: 1 book, 0 or > 1 - books

wlsch:
Just a little tooltip string fix for just one book it will be saying "1 book" instead "1 books".


kovidgoyal:
The "correct" way to handle this situation is to use the ngettext function, like this, 

ngettext('{0} [{1} book]', '{0} [{1} books]', count).format(...)

This is needed for languages that can have different plural forms for 1,2,3 and many

PR #377, opened 2015-03-28
Add ability to create additional empty formats to the "add empty book" ...

cbhaley:
...command. Add a new command to add an empty format to existing books.

The code to add an empty docx should be added to oeb/polish/create.py


cbhaley:
Added a new commit, putting the list of valid empty formats along side the function that creates empty formats.

kovidgoyal:
I have merged this PR, but i missed the last commit, which I implemented myself, in a slightly different fashion.

cbhaley:
FWIW: I don't think that the title should be added to the txt file. In at least one use case what is desired is an empty file to avoid taking up space on devices. If someone wants to keep notes and wants the title in the file then the person can add it.

kovidgoyal:
All the other formats have at least some metadata stored, so to be consistent I added the title to TXT. Its only a few bytes. With even a collection of ten thousand books, the required storage would be ~ 300KB. Also note that title is written only if it is not unknown, so in your use case which was adding from the add empty books dialog, it wont be written.

PR #378, opened 2015-04-01
Add an option to disable writing calibre_boookmarks.txt into epub file

gutenye:
There's a lot of posts saying about it, a quick Google search got below links, including your reponse:
                                                                                
http://www.mobileread.com/forums/showthread.php?t=85802
http://ebooks.stackexchange.com/questions/2603/how-to-delete-calibre-epub-bookmarks
https://code.google.com/p/epub-squeaker/wiki/RemoveCalibreBookmarks             
                                                                                
**USAGE**                                                                       
                                                                                
``` bash                                                                        
$ CALIBRE_NO_WRITE_BOOKMARKS=1 calibre                                          
```                                                                             
                                                                                
The enviornment variable name is a bit longer, maybe you can think of a better version.

### DIFF
--- a/src/calibre/ebooks/oeb/iterator/bookmarks.py
+++ b/src/calibre/ebooks/oeb/iterator/bookmarks.py
@@ -81,7 +81,7 @@ class BookmarksMixin(object):
         if bookmarks is None:
             bookmarks = self.bookmarks
         dat = self.serialize_bookmarks(bookmarks)
-        if os.path.splitext(self.pathtoebook)[1].lower() == '.epub' and \
+        if (not os.environ["CALIBRE_NO_WRITE_BOOKMARKS"]) and os.path.splitext(self.pathtoebook)[1].lower() == '.epub' and \
             os.access(self.pathtoebook, os.R_OK):
             try:
                 zf = open(self.pathtoebook, 'r+b')


kovidgoyal:
There is already such an option, in the Preferences->Miscellaneous section of the viewer. It prevents the viewer from remembering last read position/bookmarks.

gutenye:
@kovidgoyal 

Thanks for quick replying, but that's not the case, I want the `remember last read position` feature, and I also want calibre does not modify epub file. So put the `calibre_bookmarks.txt` file out of epub is the best option.

kovidgoyal:
Then you should make an actual option for the viewer in the Preferences->Miscellaneous section that controls that behavior. Not an env var. Something like "When vieweing EPUB files, store bookmarks in the file" and default it to True.

gutenye:
OK, I'll give it a try.

By disable `write calibre_booksmarks.txt to epub` [in line 84](https://github.com/kovidgoyal/calibre/blob/e4ac9c78c8820f38f37079cd2a874d0fa073b812/src/calibre/ebooks/oeb/iterator/bookmarks.py#L84) , it'll automatically fallback to  write it somewhere else, just like it did with the `.mobi` file, Am I understand the code correctly? 

If that is so, I just need to add an GUI option for it.

kovidgoyal:
Yes, that is correct.

kovidgoyal:
And note that your patch is insufficient. You also have to deal with reading bookmarks. Think about what happens if the user uses the option to turn off in file bookmarks, but opens an epub file with an previously created calibre_bookmarks.txt

gutenye:
Indeed, thanks for the heads up, I'll fix it on my next PR.

On Fri, Apr 3, 2015 at 11:36 AM, Kovid Goyal <notifications@github.com>
wrote:

> And note that your patch is insufficient. You also have to deal with
> reading bookmarks. Think about what happens if the user uses the option to
> turn off in file bookmarks, but opens an epub file with an previously
> created calibre_bookmarks.txt
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/kovidgoyal/calibre/pull/378#issuecomment-89148569>.
>



-- 
Linux, Vim, Ruby, Go and Javascript
- Guten.me  <http://GutenYe.com>
- GitHub.com/GutenYe <http://Github.com/GutenYe>

PR #379, opened 2015-04-03
Add a store_epub_bookmarks_in_the_file option for ebook-viewer

gutenye:
Follows #378

I'm not a python expert, and I want to test it by myself first, but I can't find out where to begin. I tried `$ python2 setup.py install --prefix=dist`, it does not work. And I use ArchLinux.

### DIFF
--- a/src/calibre/ebooks/oeb/iterator/bookmarks.py
+++ b/src/calibre/ebooks/oeb/iterator/bookmarks.py
@@ -66,7 +66,8 @@ class BookmarksMixin(object):
         self.bookmarks = []
         bmfile = os.path.join(self.base, 'META-INF', 'calibre_bookmarks.txt')
         raw = ''
-        if os.path.exists(bmfile):
+        if self.get_sore_epub_bookmarks_in_the_file_opt() and \
+            os.path.exists(bmfile):
             with open(bmfile, 'rb') as f:
                 raw = f.read()
         else:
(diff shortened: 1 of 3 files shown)


gutenye:
I've read http://manual.calibre-ebook.com/develop.html, nothing there saied how to build calibre  from the source.

kovidgoyal:
Instructions for setting up a calibre development environment are here: http://manual.calibre-ebook.com/develop.html

kovidgoyal:
You dont need to build calibre from source, simply setup the development environment and run calibre, your changes will be picked up automatically.

kovidgoyal:
I have committed an implementation of this, see above. It will be merged into master in a day or two. Feel free to post if you have any comments.

gutenye:
Thank you very much. 

I've tried a lot,  calibre distributed by Archlinux does not support `CALIBRE_DEVELOP_FROM`, and I have to download linux installer from calibre site, and .... 

You just saved me a lot of time :)

eli-schwartz:
`CALIBRE_DEVELOP_FROM` is only useful for a frozen calibre install (I suppose the theory is that if you install from source, you can update the install from the same source). And most distros have a habit of "fixing" program source code, so only the frozen binaries are officially supported.

Speaking of which, I don't recommend using the distro calibre, even on ArchLinux -- go check my still-open bugreport, they strip heavily-modified-from-upstream libs because Arch Purity.
So remind me, why do we use Arch again? Oh, right, because (partially): vanilla packages, unless changes are needed to compile/avoid breakage. Riiiiiiiiiiiiiiiiiiight.


Oh -- if you really want to, you can always use the AUR ;) they have a calibre-git package.
You *can* add patches and rebuild that (and the AUR package leaves modified libs alone, yay) and then you will have a patched calibre version that doesn't rely on source checkouts.

gutenye:
@eli-schwartz 

That's very nice, I totally forget about the `calibre-git` in the AUR. :+1:

PR #382, opened 2015-04-11
Adding The Codeless Code (thecodelesscode.com)

marumari:
This is the recipe for thecodelesscode.com.  Unlike most recipes, it creates more of an eBook than an article collection.


marumari:
Hah, I'm honored to rate a whole new API.  Can I ask why it is set to False by default?  I understand that it's not the current behavior, but it seems like all upside to me.

marumari:
I tried to have it work, and despite the dictionary containing:
(u'www.thecodelesscode.com', u'/case/62'): set(['feed_0/article_57/index.html'])

I still get:
Referenced file u'/case/62' not found

It could just be that something is screwed up with my build environment.  I'm on a Mac, and Calibre builds horribly there.  I'll wait for the next full release, and then build my recipe against that.  If you want to cancel out this pull request, go ahead.

marumari:
Ah ha!  Never mind; I figured it out.  The default resolve_internal_links should probably be smart enough to resolve them either by the path only or by the path + URI.  I had to update canonicalize_internal_url to remove the http://fqdn (to make it a relative URI), and that fixed it.

marumari:
Okay, the pull request should be ready to go.  :)

kovidgoyal:
It is off by default because there is no way to reliably map URLs/hrefs to their canonical form. That bit of functionality will need to be implemented and tested in each individual recipe.

kovidgoyal:
And incidentally you do not need to build calibre to run it from source, simply clone the source and set a single env var. http://manual.calibre-ebook.com/develop.html

PR #383, opened 2015-04-13
This comes from http://www.mobileread.com/forums/showthread.php?t=258781...

cbhaley:
... (row height problems).

It turns out that the bool icons were not scaled to the row height - 4. This patch fixes that.


kovidgoyal:
I implemented a different solution for two reasons:

1) Your approach resulted in icons that were much larger than before with the default settings (where there is considerable padding between rows). I dont know if this a good or bad thing, but people would notice and probably complain

2) I'm not wild about calling build_data_convertors in set_row_height. Of course this can be trivially fixed by just changing build_data_convertors to not use a local reference to the icons, which is onlya  slight performance penalty.

PR #401, opened 2015-05-26
Three changes for editing tags

cbhaley:
The first commit improves the category editor available from the tag browser. The second commit makes Shift F2 (or whatever the edit key is) on a tags-like column in the book list open the tags editor instead of the line editor. The third removes the edit tag button in single edit metadata because it alters the order of the tags.


kovidgoyal:
Merged. I have not looked at it, but it should not be too hard to have the tag editor optionally preserve order.

cbhaley:
Preserving order isn't too hard, but that is only part the problem. It would also need buttons to change the order, especially important when new people are added, which is rather more work.

kovidgoyal:
I usually use drag and drop for that, easier to implement and easier to use as well :)

PR #402, opened 2015-05-27
Handle custom is_names columns in the tag editor.

cbhaley:
You were right -- drag/drop was trivial. However, I didn't find a straightforward way to know that the order had been changed, so I changed the code to regenerate the applied tags list whenever necessary.

I thought about changing it to work on "authors" but that turned out to be more work than I wanted to do. :)

### DIFF
--- a/src/calibre/gui2/custom_column_widgets.py
+++ b/src/calibre/gui2/custom_column_widgets.py
@@ -317,9 +317,7 @@ class Text(Base):
             if self.sep['ui_to_list'] == '&':
                 w.set_space_before_sep(True)
                 w.set_add_separator(tweaks['authors_completer_append_separator'])
-                w.get_editor_button().setVisible(False)
-            else:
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Is there a reason for using SingleSelectionMode when enabling drag and drop?

cbhaley:
Yes. The behavior with multi-select and drag/drop (on WIndows) is strange. As far as I can tell there is no way to unselect everything. If I select a group of items then click on one of them to drag it, that item is unselected and the remain items are dragged. It is very hard to use.

I figured that is_name columns usually didn't contain 100s of items, so the usability gained was worth the loss of multi-select.

-- Charles

kovidgoyal:
That's because the .ui file sets selection mode to multiselect, I changed it back to extended select and it should be fine. Also the list setup code was not being run if the intialvalue of the field is empty, which, I think, is incorrect.

PR #414, opened 2015-07-22
Change default date format for news

stanh:
It's easier to organize news library using this [yyyy-mm-dd] format, especially when the library gets big

### DIFF
--- a/src/calibre/web/feeds/news.py
+++ b/src/calibre/web/feeds/news.py
@@ -91,7 +91,7 @@ class BasicNewsRecipe(Recipe):
 
     #: The format string for the date shown on the first page.
     #: By default: Day_Name, Day_Number Month_Name Year
-    timefmt                = ' [%a, %d %b %Y]'
+    timefmt                = ' [%Y-%m-%d]'
 
     #: List of feeds to download.
     #: Can be either ``[url1, url2, ...]`` or ``[('title1', url1), ('title2', url2),...]``


kovidgoyal:
It's harder to parse YYYY-mm-dd date format for humans. Don't rely on the date format to organize your library, use the date column, that's why it exists :)

stanh:
I don't understand why this is hard for human. For many of us this is quite natural. People are requesting this format back from 2011. For mobile / iPad / tablets users, there is no such a date column.

kovidgoyal:
It's simple, the the purpose of a date is two-fold. To enable sorting and to label an epoch. Humans mostly use dates for labelling. And the vast majority of humans do not  think of months and days in terms of numbers, so representing dates as numbers causes extra overhead for parsing them in humans. 

If you need to mangle titles for limited functionality devices, use the metadata plugboards functionality, that's what it is there for.

stanh:
Some news are published weekly, so the weekday is already irrelevant - what's important is which week.

Since calibre is so widely used, is it more reasonable to make this an option that we can customize, instead of hard-coded?

kovidgoyal:
It is customizable, you can customize it by customizing the recipe and adding your own timefmt to the customized version.

stanh:
Sorry someone says impossible in the forum. Could u advice how to do that please? Didn't get much on this from Google

kovidgoyal:
See http://manual.calibre-ebook.com/news.html

instead of creating a new recipe from scratch, simply use the "Customize builtin recipe" button in the dialog.

stanh:
Wow not entirely straightforward but finally solved my headaches!! thanks!

PR #416, opened 2015-08-05
Write the "real" series sort value to the existing column in the …

cbhaley:
…series table. Reason: other than composite columns, series_sort is the only value that cannot be extracted from a metadata.db.

This scheme change will not prevent reinstalling an older version (within reason) because the title_sort db function has existed for years.

Currently no one uses the series table sort column because it doesn't contain useful information. This commit changes that. That said, there is no reason to change any existing code because it should be computing the same value being put into the column, and if for some reason it computes a different value then we want to keep the current behavior.

### DIFF
--- a/src/calibre/db/schema_upgrades.py
+++ b/src/calibre/db/schema_upgrades.py
@@ -615,3 +615,28 @@ class SchemaUpgrade(object):
         self.db.execute(script)
 
 
+    def upgrade_version_21(self):
+        '''
+        Write the series sort into the existing sort column in the series table
+        '''
+
+        script = '''
+        DROP TRIGGER series_insert_trg;
+        DROP TRIGGER series_update_trg;
+
+        UPDATE series SET sort=title_sort(name);
+
+        CREATE TRIGGER series_insert_trg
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Hmm, db schema changes make me nervous. Shouldn't that be DROP TRIGGER IF EXISTS, just in case there are dbs out there where people have removed the trigger?

Also, while I haven't looked at the code to be sure, I have a vague memory that calibre calculates series sort based on the current language of the book. I know that is the case for title sort for certain, not sure for series sort.

cbhaley:
I pushed changes to do IF EXISTS.

Series sort is a bit of a mess. When you sort the book list the "most common language" is used to generate the series sort, which means that the sort can change from time to time depending on the books in the library. As far as I can tell that is the only place where language is used when generating series sort. For example, none of the tag browser, the smart_device driver, or the content server device compatible JSON function specify a language. Save_to_disk does not use a language when it writes the series sort into the OPF.

The title_sort() method uses the interface language and the tweaks, so even without the "most common language" stuff it gets close. It is certainly better for people who directly use the database than having nothing.

PR #420, opened 2015-08-24
Fix for bug 1487949: "contamination" of the user-defined template …

cbhaley:
…function cache.

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -310,7 +310,8 @@ class DB(object):
     # Initialize database {{{
 
     def __init__(self, library_path, default_prefs=None, read_only=False,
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Not going to spend the time getting hunspell built in the old build
system, it's just duplicated effort. But you can probably use the patch
directly in your calibre package if you like. Check the next commit
also, it had some minor build relaed things, not sure if you will need
them or not.

cbhaley:
Kovid, is your comment on the right pull request? I don't use hunspell and the commit seems unrelated to it.

kovidgoyal:
Yeah, sorry, went to the wrong place by mistake, pls ignore.

PR #424, opened 2015-09-09
When updating a book file on the device, send the cover as a tuple …

cbhaley:
…instead of a simple byte string, which is how a cover is sent using the normal path.


cbhaley:
The problem I was trying to solve was that books sent via the sync mechanism had no thumbnail. Unfortunately I had my head somewhere where it shouldn't be. The fix I provided yesterday masked the problem but actually fixed nothing, and for the reasons you provided made things worse. Teach me to code when I am tired. (sigh)

I have pushed the correct fix.

kovidgoyal:
I cherry picked the last commit so as to avoid having the various extra commits this pull request picked up. So you will likely need to rebase on master.

PR #425, opened 2015-09-15
Updates for Instapaper recipe (HTTPS, cleanups, ...)

Mno-hime:


kovidgoyal:
I am happy to merge some of those changes (the first two commits). But the latter two are likely to be controversial, and as such, you should get them signed off by the original recipe author. (https://bitbucket.org/khromov/calibre-instapaper)

khromov:
It looks like the article order change was added fairly recently, in 93b6d3997401733cc7c651b759d6796766e347eb

Personally, I am against that changeset, since:
* The article order is not as on Instapaper.com (Like @Mno-hime points out)
* If you have many articles, you won't get the latest ones, you'll get the oldest ones. That's not a very good user experience.

So I am definitely for merging that changeset to go back to the previous chronological order.

I see the point in removing the Starred articles feed since they are more used for sharing and preserving for a longer period of time, so it doesn't make a lot of sense to keep downloading them every time.

A GUI for recipe options are at the top of my wishlist for Calibre. :smile:

PR #429, opened 2015-10-04
Make the keyboard shortcut for the tag browser "find" text box use …

cbhaley:
…the general keyboard mechanism. Add a shortcut to "click" the current item in the tag browser, which is most useful after a "find". Also add shortcuts for the collapse button and the Alter Tag Browser button.

I didn't change the shortcut for the tag browser Find button. It currently uses the "& in translated string" method, so changing it to a fixed shortcut would change functionality. If this is OK I can make the change.


kovidgoyal:
I'm OK with changing the Find button to use a configurable shortcut. Mixing translated and configurable shortcuts in the same UI is not a good idea anyway.

PR #432, opened 2015-10-15
Make SHIFT+F2 clear the date when launching the editor on a date…

cbhaley:
…in the book list. This doesn't steal a shortcut because F2 with any modifier launches the delegate.


cbhaley:
Kovid: ignore this pull request for the moment. I didn't do what BetterRed asked for. :(

kovidgoyal:
You should probably have Shift+F2 clear the contents of any cell before editing it, not just date cells. Otherwise, sooner or later someone will open a bug report requesting that :)

kovidgoyal:
Sorry, got you second message too late.

PR #433, opened 2015-10-15
Keyboard modifiers (again) for delegates. Make CTRL+F2 clear the…

cbhaley:
…field's value for all fields. SHIFT+F2 sets date fields to now().


kovidgoyal:
You should not use modifiers() == Qt.Ctrl_Modifier because on some platforms, modifiers might include other flags like a numlock, special keys, etc. in which case it will never match. Instead check int(modifiers() & Qt.Ctrl_Modifier)

cbhaley:
Fixed to ignore all modifiers other than CTRL and SHIFT, which prevents ambiguity if both CTRL and SHIFT are pressed.

PR #442, opened 2015-11-05
Load system tray icon from the current theme

Arkq:
Loading the tray icon from the theme allows the user to customize the
desktop in greater extend. If the icon is not present in the icon set,
then the bundled one is loaded, so the previous behavior is preserved.
The themed icon name is consistent with the application name - calibre.

### DIFF
--- a/src/calibre/gui2/ui.py
+++ b/src/calibre/gui2/ui.py
@@ -21,7 +21,8 @@ from PyQt5.Qt import (
     QApplication, QSystemTrayIcon)
 
 from calibre import prints, force_unicode, detect_ncpus
-from calibre.constants import __appname__, isosx, filesystem_encoding, DEBUG
+from calibre.constants import (
+        __appname__, isosx, iswindows, filesystem_encoding, DEBUG)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
The name for the calibre icon should be calibre-gui not calibre. This is the name used by the calibre binary installer when installing icons into the (default) hicolor theme. Also since icon themes are a  linux only thing, that code should be conditional 

```py
from calibre.constants import isosx, iswindows

if not (iswindows or isosx):
   ....
```

Arkq:
As for the icon name, I will change that. However, I'm wondering why the linux condition is necessary. This code (fromTheme) should work (I've never done anything in the Python binding before, so I might be wrong) on all platforms, or am I missing something?

kovidgoyal:
It means I dont have to spend time making sure that it will work on those platforms.

Arkq:
I've rebased the code according to your comment. I was trying to make this branching (platform condition) without much maintenance penalty (tray icon is not displayed until the visibility is set to true, so this code should work for Linux as fast as before), but if you like more normalized approach, please let me know.

PR #447, opened 2015-11-20
Fix tag order tweak so that user categories are always at the end. …

cbhaley:
…Most of the complexity comes from the default node order putting search after the categories.

### DIFF
--- a/src/calibre/spell/dictionary.py
+++ b/src/calibre/spell/dictionary.py
@@ -243,13 +243,13 @@ class Dictionaries(object):
 
     def add_user_words(self, words, langcode):
         for d in self.dictionaries.itervalues():
-            if getattr(d.primary_locale, 'langcode', None) == langcode:
+            if d and getattr(d.primary_locale, 'langcode', None) == langcode:
                 for word in words:
                     d.obj.add(word)
 
     def remove_user_words(self, words, langcode):
         for d in self.dictionaries.itervalues():
-            if d.primary_locale.langcode == langcode:
+            if d and d.primary_locale.langcode == langcode:
                 for word in words:
                     d.obj.remove(word)
 


kovidgoyal:
My first, probably stupid, idea is, why not sort the top level nodes of
the tree after building the tree. It should not be a big performace hit,
since there are a relatively small number of top level nodes.

I think one would have to be a little careful with handling
hierarchical user categories, since IIRC they are sorted also
by the initial sort?

Kovid.

On Fri, Nov 20, 2015 at 07:13:01AM -0800, Charles Haley wrote:
> …Most of the complexity comes from the default node order putting search after the categories.
> 
>   https://github.com/kovidgoyal/calibre/pull/447
> 
> -- Commit Summary --
> 
>   * Fix tag order tweak so that user categories are always at the end. Most of the complexity comes from the default node order putting search after the categories.
> 
> -- File Changes --
> 
>     M src/calibre/gui2/tag_browser/model.py (21)
> 
> -- Patch Links --
> 
> https://github.com/kovidgoyal/calibre/pull/447.patch
> https://github.com/kovidgoyal/calibre/pull/447.diff
> 
> ---
> https://github.com/kovidgoyal/calibre/pull/447

-- 

Dr. Kovid Goyal 
http://www.kovidgoyal.net
http://calibre-ebook.com

kovidgoyal:
Another, perhaps easier, one is to change the loop at line 593 to iterate over nodes in a custom order instead of the sorted order (i.e. something like this, in pseudo code)

```py
last = []
for category in self.category_nodes:
    if category.is_user_category:
        last.append(category)
    process_one_node(category)
for uc in last:
    process_one_node(uc)
```

kovidgoyal:
Also, unrelated to this particular issue, currently the categories are sorted by the order in which they appear in field metadata, (modulo the tweak is present) should that perhaps be changed to be alphabetic? This would be a fairly noticeable behavior change, however.

cbhaley:
It turns on that things are a bit more complicated. Things don't work correctly if a user category contains a hierarchical item from some other category that is currently hidden, because the tree isn't built. I am thinking that the right approach is something like you suggested, but altered to build the complete tree then copy the top level nodes into the displayed list. I will look at that.

IIRC the order in field metadata was chosen because that was the display order in the tag browser. In any event, I don't think it is a good idea to change the default order. There is no pressing reason to do so and there will be screams of anguish from those who don't like the change.

-- Charles

PR #448, opened 2015-11-21
Build the node tree in the tag browser in the necessary order …

cbhaley:
…(user categories last), then hide and reorder the tree as needed.

Somehow I deleted the last pull request and all its comments. Sigh ...

### DIFF
--- a/src/calibre/gui2/tag_browser/model.py
+++ b/src/calibre/gui2/tag_browser/model.py
@@ -560,19 +560,24 @@ class TagsModel(QAbstractItemModel):  # {{{
                                 '5state' if tag.category != 'search' else '3state'
                         else:
                             if i < len(components)-1:
-                                t = copy.copy(tag)
-                                t.original_name = '.'.join(components[:i+1])
-                                t.count = 0
-                                if key != 'search':
-                                    # This 'manufactured' intermediate node can
-                                    # be searched, but cannot be edited.
-                                    t.is_editable = False
-                                else:
-                                    t.is_searchable = t.is_editable = False
+                                original_name = '.'.join(components[:i+1])
+                                t = self.intermediate_nodes.get((original_name, tag.category), None)
+                                if t is None:
+                                    t = copy.copy(tag)
+                                    t.original_name = original_name
+                                    t.count = 0
+                                    if key != 'search':
+                                        # This 'manufactured' intermediate node can
+                                        # be searched, but cannot be edited.
+                                        t.is_editable = False
+                                    else:
+                                        t.is_searchable = t.is_editable = False
+                                    self.intermediate_nodes[(original_name, tag.category)] = t
                             else:
                                 t = tag
                                 if not in_uc:
                                     t.original_name = t.name
+                                self.intermediate_nodes[(t.original_name, tag.category)] = t
                             t.is_hierarchical = \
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Before I merge this, I'd like to understand process_one_node() a little better. In case you missed my email in the kerfuffle:

I've been trying to follow process_one_node() in detail and I dont see
where user categories reference nodes created in previous categories.
process_one_node() seems to unconditionally call create_node() for all
categories, user or otherwise. The only thing that looks conditional is category_child_map and that is local to process_one_node(). What am I missing?

cbhaley:
I misspoke myself a bit in a previous comment.

Building user categories does not reuse the node tree. It can't because then opening one node would play havoc wherever else that node appeared. What actually happens is that for leaf nodes the item data is reused (the "data" in the TagTreeItem). This allows marking a leaf in one place to mark it everywhere.

The reason that user categories must be last is that the processing depends on the tag data being already marked as hierarchical or not. For example, if a hierarchical tag appears in a user category then that tag must be hierarchical in the UC even though the UC is not itself hierarchical. The way for the UC to know this is that the shared tag data has already been so marked.

Referring to the source as of my commit: line 533 is the main gateway. Processing will go to line 536 if the category is not hierarchical or the number of components == 1 (amongst other things). Assume that "tags" is a hierarchical category and the the item name is AA.BB. In this case processing will go to line 549 where the hierarchical node tree is generated and the tag is marked as hierarchical. In a subsequent call to process_one_node when a non-hierarchical user category is processed, the tag will have been marked so processing will go to line 549, building the intermediate node trees.

Perhaps changing the test on lines 528 and 535 to check the base category if a user category is being processed could eliminate the problem. The goal would be to detect whether an item is in a hierarchical category even if the current category is not hierarchical.

It is also probably the case that the "data" for the generated intermediate nodes could be reused. As it is, clicking on a generated node does the correct search, but the equivalent generated nodes in other categories are not marked. Marking across categories only works for leaf nodes; nodes returned by get_categories. I will take a look at this, mostly out of curiosity.

cbhaley:
Adding reusing the intermediate (generated) node data was easy.

kovidgoyal:
I've merged, but while playing around with it, I noticed that
while hierarchical items in normal user categories do indeed show up as
hierarchical, those in user categories created from grouped search terms
do not. Is there some reason for that?

On Sat, Nov 21, 2015 at 06:12:36AM -0800, Charles Haley wrote:
> Adding reusing the intermediate (generated) node data was easy.
> 
> ---
> https://github.com/kovidgoyal/calibre/pull/448#issuecomment-158645385

-- 

Dr. Kovid Goyal 
http://www.kovidgoyal.net
http://calibre-ebook.com

cbhaley:
Two reasons:

1) GST nodes are clones of the nodes in the source categories, not references to the original nodes. A node can represent more than one category, e.g., the same series name appearing in two columns. It isn't clear what one should do if one source is hierarchical and another is not. It is perhaps worth noting that although this merging was intended, it didn't work until a few days ago,

2) I wanted GSTs to "feel" more like a column. As part of that, they can be made hierarchical in settings, while "real" user categories cannot. The user adds hierarchy explicitly to a UC by creating sub-UCs, something that cannot be done for a GST.

PR #462, opened 2015-12-08
Responsive UI for the Content Server

RobertoGonzalez:
Hey Kovid,
I did a quick redesign of the content server, since the current one is not responsive and difficult to use on mobile. It's not exactly production ready (I built it for personal use), but I wanted to see if you are interested in this feature to finish it properly.

Here's a screenshot for reference (though it's better if you try it on iOS/Android device to see the difference in usability):

calibre


kovidgoyal:
The content server is currently in the process of being redesigned. 

Run the new server like this:

calibre-debug --new-server /path/to/calibre/library

You are welcome to contribute to the new server's code/design. 

With regard to your design, the new server has support for multiple, responsive renderings of the book list -- currently I have only implemented a cover grid, but other views will be fairly easy to implement in the future. 

Also note that the current server has a dedicated mobile view you can access via /mobile that is fairly useable on small screens.

PR #464, opened 2015-12-15
Change Name=calibre to Name=Calibre for .desktop file entry

foolsh:
English conventions standardize the capitalization of names.

### DIFF
--- a/src/calibre/linux.py
+++ b/src/calibre/linux.py
@@ -1019,7 +1019,7 @@ GUI = '''\
 [Desktop Entry]
 Version=1.0
 Type=Application
-Name=calibre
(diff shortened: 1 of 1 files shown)


eli-schwartz:
calibre conventions standardize the non-capitalization of the application name. Presumably Kovid had his reasons (and if it is deliberate, then it's probably not going to be changed...).

Regardless, it seems somewhat odd to capitalize it in the linux desktop file but nowhere in the UI or the manual or in the OSX/Windows installers.

kovidgoyal:
It is deliberate and is not going to change. It comes from calibre's roots as a command line utility.

PR #467, opened 2015-12-26
convert the text for Semitic Languages

mushony:
convert the text for Semitic Languages - for example:
text like this:
"author:   someone"
Looks like this, in Semitic Languages :
"someone  :author"
its necessary for good display to these languages

### DIFF
--- a/src/calibre/gui2/book_details.py
+++ b/src/calibre/gui2/book_details.py
@@ -75,6 +75,12 @@ def render_html(mi, css, vertical, widget, all_fields=False, render_data_func=No
         ans = templ%(u'<table><tr><td valign="top" '
             'style="padding-right:2em; width:40%%">%s</td><td valign="top">%s</td></tr></table>'
                 % (table, right_pane))
+    
+    # Layout of text to Semitic languages display
+    from calibre.adjustments_lang import html_text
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm somewhat confused by this pull request. What are you trying to achieve? You want labels to be the second column in the book details panel when the calibre user inteface language is a RTL langugae like Hebrew? 

The correct way to do that is to change the the mi_to_html() function to generate the table with the desired layout when passed a parameter rtl=True.

You can detect when to set rtl=True by checking the value of get_lang() from calibre.utils.localization. You would need to chech for both hebrew and arabic language codes.

kovidgoyal:
See https://github.com/kovidgoyal/calibre/commit/c33017583e4ac0663405276efd0f66b255af340c

mushony:
Very well.
But I found a small bug in the final display of book details.
This can be remedied only a in the file "book_details.py".
I'll upload the fixed code.
By the way, where I can convert the main window of the library (RTL), like you did here?

kovidgoyal:
You call QApplication::setLayoutDirection() However I would not recommend doing that, as it has very radical effects that are likely to break many things.

PR #473, opened 2016-01-01
 Book details panel: Display data in right-to-left format

mushony:
set to a right-to-left language such as Hebrew or Arabic

### DIFF
--- a/src/calibre/gui2/book_details.py
+++ b/src/calibre/gui2/book_details.py
@@ -45,6 +45,14 @@ def render_html(mi, css, vertical, widget, all_fields=False, render_data_func=No
 
     c = color_to_string(QApplication.palette().color(QPalette.Normal,
                     QPalette.WindowText))
+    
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Changing the entire html data to right to left unconditionally is not correct. There is no guarantee that the text being displayed is actually right to left. The rest of the book details panel display metadata from the book, which could be in any language. You need to add that code to the HTML in the individual book comment to have it display RTL if appropriate.

mushony:
You can not display the text correctly without using this code.
Although the text display in the right way, but not in the right place.
So that the text look like:


Instead of:

mushony:
By the way,
for LTR languages, the text is still displayed LTR (style="text-align:left")

kovidgoyal:
If you want it aligned right then add the align right to the table not the overall html. Adding it to the overall html will also affect the comments, whose language is not known.

PR #474, opened 2016-01-03
Fixing an exception if the CSS contains "@media all;"

sarnau:
Fixing an exception if the CSS contains "@media all;" and not the excepted "@media ... {"


kovidgoyal:
@media\s.*?[{;] will need the dotall flag. And @media\s[^{]*?[{;] will fail for empty media rules, for example it will match

@media all;  @media screen {

PR #483, opened 2016-02-15
Add Al-Masry Alyoum (Arabic) recipe.

hazrpg:


hazrpg:
This is for the Arabic version of "Egypt Independent" (Al-Masry Alyoum). The version currently in the master branch with the same name is in English (Egypt).

The current version points to: egyptindependent.com and this PR points to: almasryalyoum.com

As a side note, it appears the English version is broken due to the recent change the newspaper did to their domains. Previously the English articles were published on almasryalyoum.com but all the old English RSS feed urls now point to the home directory of egyptindependent.com - causing the feeds to pull the wrong data.

hazrpg:
Thank you for accepting the pull.

PR #484, opened 2016-03-10
Don't block UI when testing email with wizard

worr:
The test_email dialog in the wizard runs the email test in the same
thread as the UI. In the case where there's a timeout trying to send the
email, this can block the UI for up to thirty seconds.

This patch runs the email test in a new thread, and updates the
QTPlainTextEdit field via a slot rather than manually.

### DIFF
--- a/src/calibre/gui2/wizard/send_email.py
+++ b/src/calibre/gui2/wizard/send_email.py
@@ -9,6 +9,7 @@ __docformat__ = 'restructuredtext en'
 import cStringIO, sys
 from binascii import hexlify, unhexlify
 from functools import partial
+from threading import Thread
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
There are a couple of problems with this:

1) There needs to be some indication to the user that something is happening. Set the busy cursor and output some text in the log area saying, Sending mail, please wait...

2) You have to handle the case that the dialog is closed while the thread is running. You might be lucky and nothing bad will happen, but if one of the Qt objects involved int he signal/slot interaction is deleted, you may get an unhandled exception. So better to override the reject() method of the dialog and prevent the dialog from closing while the test thread is working.

kovidgoyal:
I modernized that dialog, which takes care of the ui blocking as well.

PR #487, opened 2016-03-25
Write the value of composite columns to the database when the OPF …

cbhaley:
…backup is written.

The rationale: several programs/applications read the calibre database to get information about a book. One problem: the values of composite columns isn't there because those values are computed as needed. This patch writes the values to the db that are correct at the moment the metadata.opf is written. That means that an app can use the dirtied table to check if the values might be out of date.

Without the mark_dirty change the book is backed up over and over.

### DIFF
--- a/src/calibre/db/backup.py
+++ b/src/calibre/db/backup.py
@@ -83,6 +83,17 @@ class MetadataBackup(Thread):
             self.db.clear_dirtied(book_id, sequence)
             return
 
+        cc_metadata = mi.get_all_user_metadata(False)
+        for key,custom_col in cc_metadata.iteritems():
+            try:
+                if custom_col.get('datatype') == 'composite':
+                    self.db.set_field(key, {book_id: custom_col.get('#value#')},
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I'm afraid I dont like this change. It imposes a fair bit of work on calibre for no benefit to calibre itself. While I am generally willing to accommodate the needs of third party programs, I draw the line at changes that impose significant costs on core calibre.

Also, the idea of storing ephemeral data in the database just smells wrong to me. It increases database size, it makes backups slower/more resource intensive and there is no guarantee that the value stored at backup time is the correct value/current value, since backup is a async process. It is true that one could use the dirty table to check integrity, but that is not guaranteed to actually work -- for example if a composite column uses data from outside the db -- library name or existence of format files on the file system, etc.

Finally making changes in the guts of the db layer requires fairly intensive review/testing -- I am particularly concerned about the changes in db.write. I am not sure that there is no code anywhere that relies on custom columns not touching the db even when set() is called.

PR #489, opened 2016-03-28
Add a LibraryClosed plugin type that is called when the library is …

cbhaley:
…closed (new_api). Add a way to provide a message during shutdown.

The shutdown message stuff turned out to be mildly complicated. Using timers broke the rest of shutdown (messages about unexpected threads), as did permitting garbage collection of the existing central widget.

### DIFF
--- a/src/calibre/customize/__init__.py
+++ b/src/calibre/customize/__init__.py
@@ -789,8 +789,8 @@ class LibraryClosedPlugin(Plugin):  # {{{
     '''
     type = _('Library Closed')
 
-    # minimum version 2 because it requires the new db
-    minimum_calibre_version = (2, 0, 0)
+    # minimum version 2.54 because that is when support was added
+    minimum_calibre_version = (2, 54, 0)
 
     def run(self, db):
         '''
--- a/src/calibre/customize/__init__.py
+++ b/src/calibre/customize/__init__.py
@@ -781,3 +781,25 @@ class EditBookToolPlugin(Plugin):  # {{{
 
 # }}}
 
+class LibraryClosedPlugin(Plugin):  # {{{
+    '''
+    LibraryClosedPlugins are run when a library is closed, either at shutdown,
+    when the library is changed, or when a library is used in some other way.
+    At the moment these plugins won't be called by the CLI functions.
+    '''
+    type = _('Library Closed')
+
+    # minimum version 2 because it requires the new db
+    minimum_calibre_version = (2, 0, 0)
+
+    def run(self, db):
+        '''
+        The db will be a reference to the new_api (db.cache.py).
+
+        The plugin must run to completion. It must not use the GUI, threads, or
+        any signals.
+        '''
+        raise NotImplementedError('LibraryClosedPlugin '
+                'run method must be overridden in subclass')
+# }}}
+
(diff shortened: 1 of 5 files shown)


kovidgoyal:
Causes a random crash for me on linux with this traceback:
```
  File "/home/kovid/work/calibre/src/calibre/gui2/cover_flow.py", line 391, in cover_flow_do_sync
    if (self.is_cover_browser_visible() and self.cf_last_updated_at is not None and time.time() - self.cf_last_updated_at > 0.5):
  File "/home/kovid/work/calibre/src/calibre/gui2/cover_flow.py", line 365, in is_cover_browser_visible
    return not self.cb_splitter.is_side_index_hidden
  File "/home/kovid/work/calibre/src/calibre/gui2/widgets.py", line 1023, in is_side_index_hidden
    sizes = list(self.sizes())
RuntimeError: wrapped C/C++ object of type LibraryWidget has been deleted
```

 Thinking about it, I dont mind changing the plugin semantics from run on close to run on backup of book. The plugin can store the composite field data in its external location, on every backup. That imposes no overhead for people who dont install the plugin and for people that do, it does not pollute the metadata.db. It also does not involve major changes to core calibre code. And it gets rid of the shutdown complications. 

I apologise for the run around, I should have thought of this before you did all the shutdown message work.

cbhaley:
That crash looks like another garbage collection problem. I can fix that, as at one point I had a working implementation that didn't replace the main widget.

I prefer the work-on-close method over the backup method because it has much less chance of stale data. Unless you object, I am going to redo my other implementation so that no widgets are GCed.

BTW: it would be so much nicer if I could put the composite value tables directly into metadata.db instead of in a lookaside DB. Calibre would be under no requirement to maintain the table or even keep it (e.g., restore). The table wouldn't have any foreign constraints so it couldn't affect calibre operation. One table could be named 'composite_column_values' and would have three columns (book_id, column_key, column_value). The other could be names 'composite_column_timestamps' and would have two columns (book_id, last_modified). Clearly since this work is done in a non-calibre plugin I could add the tables without asking, but I would rather discuss it instead of just doing it. :)

-- Charles

kovidgoyal:
I'm ok with keeping it at shutdown. But I'd really rather not have it stored in metadata.db. At some point I plan to implement support for remote libraries in calibre. That will involve syncing metadata.db over the network -- unnecessary (from calibre perspective) data in there is not something I want to encourage.

I suggest not showing the close notification at all if no shutdown plugins exist, that will minimize the impact if there remain crash possibilities.

cbhaley:
I left in the messages for two reasons:
1) I want to know if the changes fix the crash on Linux
2) The content server shutdown has bothered me in the past. It can take minutes. The message helps, I think.

I am not married to having them, should you want to remove them.

kovidgoyal:
Does not crash for me. Some notes:

1) The minimum required version should be 2.54.0 not 2.0.0 after all, before that calibre release the plugins will not run, even if they are installed

2) In db.cache.close() use except Exception: rather than except: 

3) if available_library_closed_plugins(): should be if tuple(available_library_closed_plugins()): or better create a dedicated has_library_shutdown_plugins() method in customize.ui that does not incur the performance cost of initializing the plugins. 

The content server delay is indeed annoying -- it no longer exists in the new server :)

cbhaley:
Changes pushed.

I am not sure I did what you wanted with the has_... method. You said that having the method would not incur the performance of initialization, but the plugin in already on the _initialized_plugin list. The change does avoid the customization stuff.

PR #490, opened 2016-03-28
FreeBSD hal requires an extra step to get the correct path...

madpilot78:
… to the device to be mounted.

As stated in the subject, after testing I discovered that when mounting USB devices The FreeBSD hal has a different structure from what is expected.

This patch fixes it.


kovidgoyal:
While I dont know anything about freebsd, I suspect that whether you have to lookup the parents paths or the device's paths varies per device. So it might be better if the code looked at the union of the paths from the device and those from the parent.

madpilot78:
I just modified the code to search both in the originating device of the device itself and in the parent.

Is this what you asked for?

PR #495, opened 2016-04-16
Pass the new ID URL patterns to devices while maintaining …

cbhaley:
…compatibility with the old interface.

### DIFF
--- a/src/calibre/devices/interface.py
+++ b/src/calibre/devices/interface.py
@@ -597,21 +597,11 @@ class DevicePlugin(Plugin):
         '''
         device_prefs.set_overrides()
 
-    # A driver should implement at most one of the set_library_info methods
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Since the wireless driver is currently the only consumer of set_library_info() I'd prefer changing the signature of that method even though it is backwards incompatible. IMO, the small risk of breaking a third party driver is outweighed by the benefits of not complicating the API.

cbhaley:
If you don't mind me adding an import of "msprefs" into the wireless driver then I don't need to change the API at all. As it (msprefs) is global and is currently used by worker processes, doing the import in a device driver doesn't seem to be out of line.

cbhaley:
A commit just in case you accept the "do it in the wireless driver" idea. I used a revert commit to make it easier in the case you don't accept it.

PR #497, opened 2016-04-26
Remove woblink store

t3d:
This is related to planned removal of QtWebKit.
May or may not come back in the future...


kovidgoyal:
I suggest waiting on removing it until you have decided whether you want to refactor to work without jsbrowser. Otherwise it will cause confusion if it disappears from some calibre releases and the re-appears.

kovidgoyal:
I refactored the plugin to no longer use Qt WebKit. Take a look.

PR #498, opened 2016-04-29
Fix for 1576742: (try to) Ensure that a date that is missing the day …

cbhaley:
…of the month uses a day in the middle of every month. It seems that the base default is the last day of the month, which breaks things for February dates for non leap years when the current year is a leap year. This is a guess ...

### DIFF
--- a/src/calibre/utils/date.py
+++ b/src/calibre/utils/date.py
@@ -129,7 +129,7 @@ def parse_date(date_string, assume_utc=False, as_utc=True, default=None):
         return UNDEFINED_DATE
     if default is None:
         func = datetime.utcnow if assume_utc else datetime.now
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Hmm, I'm not entirely convinced by this. It breaks the documented behavior of parse_date() which states that default is assumed to be the current date. While I dont know if any code actually relies on that behavior, I'd rather avoid changing it, if possible. As an alternative, you could change the template functions to use parse_only_date() when the string being parsed contains only date information. However, I dont know if it is possible to do that. If not, then change the docstring of parse() to match the new behavior and I will merge.

kovidgoyal:
After investigating, I decided to implement your fix. I could get the error to happen by setting day=31 in default and parsing the date '2016-04'. So presumably, dateutil does not fix invalid values in the default, at all.

PR #501, opened 2016-05-09
Save/reuse ViewLog size plus Saved Search Changes

JimmXinu:
Save/reuse ViewLog size from proceed_question when called with log_viewer_unique_name.

Plugins use proceed_question() to safely update the database.  I'd like to have the ViewLog dialog remember its size when called from my plugin.

This change allows that, but only if proceed_question() is called with parameter log_viewer_unique_name to preserve existing behavior everywhere else.

(I considered using the existing log_viewer_title instead of a new parameter, but thought that was too likely to not be unique.)

### DIFF
--- a/src/calibre/gui2/dialogs/message_box.py
+++ b/src/calibre/gui2/dialogs/message_box.py
@@ -186,10 +186,7 @@ class ViewLog(QDialog):  # {{{
         self.copy_button.clicked.connect(self.copy_to_clipboard)
         l.addWidget(self.bb)
 
-        if unique_name:
-            self.unique_name = "ViewLog_size_"+unique_name
-        else:
-            self.unique_name = None
+        self.unique_name = unique_name or 'view-log-dialog'
         self.geom = gprefs.get(self.unique_name, None)
         self.finished.connect(self.dialog_closing)
         self.resize_dialog()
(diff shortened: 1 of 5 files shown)


kovidgoyal:
I'm actually OK with just having that dialog remember its size always.

JimmXinu:
It was not my intention to also lump these Saved Search changes in the same Pull Request, but apparently that's how github works.  Sorry about that.

Saved Search Changes:
- Add an empty default entry to the Saved Search Editor drop down.  Otherwise, to see/edit the first saved search you have to select a different one and then select the first again.
- Change Saved Search Editor add button tooltip: From the old tooltip ('Add the new saved search'), I expected to enter the new name and search and then hit 'Add'.  That would change the existing saved search (if one was selected), then create a new, empty one. 'Add a new, empty saved search with the name entered' makes that clear.
- Fix for Saved Search Rename - would add new name, but not remove old name.
- Reset Saved Search combo tooltip when search bars cleared.

JimmXinu:
> I'm actually OK with just having that dialog remember its size always. 

One size for all ViewLog calls, or one size per log_viewer_title?

kovidgoyal:
One size for all.

kovidgoyal:
If you want to use potentially separate names, then I suggest modifying your original patch to have a fallback unique name if the unique name specified by the caller is None. Something like

self.unique_name = unique_name or 'view-log-dialog'

JimmXinu:
> self.unique_name = unique_name or 'view-log-dialog'

Done.

PR #504, opened 2016-05-16
remove duplicate Economist recipe

mourner:
There are two Economist recipes in Calibre — `economist` and `conomist_free`, with _identical code_. There's no reason to have two. cc @kovidgoyal


kovidgoyal:
The reason they are there is because in the past they were different. Deleting one means that people that have subscribed to it will have their subscriptions stop working on calibre update. Which is why they are not merged.

PR #508, opened 2016-05-28
Correct spelling of tolino shine and vision devices

probonopd:
According to the manual, the tolino shine and vision names are spelled in lowercase.


kovidgoyal:
You can change the gui_name but not the name variable (as the latter is used for device configuration).

PR #509, opened 2016-06-06
Fix for composite column sorting problem …

cbhaley:
http://www.mobileread.com/forums/showthread.php?t=274903, and in particular post 23.

The fix might be better made in python's locale module, but given that the only current use of atof is in fields.py I am not sure that signing up for maintaining the python module is worth it.

### DIFF
--- a/src/calibre/db/fields.py
+++ b/src/calibre/db/fields.py
@@ -8,7 +8,7 @@ __license__   = 'GPL v3'
 __copyright__ = '2011, Kovid Goyal <kovid@kovidgoyal.net>'
 __docformat__ = 'restructuredtext en'
 
-from locale import atof
+from locale import localeconv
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Assuming the char encoding of the separator character is latin1 is probably wrong. More appropriate would be mbcs -- however, I just committed a function that gets the thousands_sep and decimal_point characters as unicode objects with no need to guess at encodings. You should probably use that in your patch -- more robust.

kovidgoyal:
I went ahead an implemented a atof() based on my code -- please test it on your windows system

cbhaley:
Your changes work on Win10/64 with both UK (dot, comma) and French (comma, non-break-space) locales. I also tested numbers of the form 111.11 (dot) when in the French locale and it worked.

PR #514, opened 2016-06-21
New tabbed configuration for the KoboTouch driver

davidfor:
This implements a new tabbed configuration dialog for drivers. The dialog uses the common configuration for the first tab and allows device specific tabs to be added as needed. Group boxes are used to group options together and make dependencies easier. For backwards compatibility, it displays the "extra configuration" options as a separate tab.

The tabbed configuration has been implemented for the KoboTouch driver. This has a lot of options with dependencies between the options. 

There is some more details about the changes in http://www.mobileread.com/forums/showthread.php?t=273129. This includes screenshots and tests by other users.

### DIFF
--- a/src/calibre/debug.py
+++ b/src/calibre/debug.py
@@ -182,7 +182,7 @@ def print_basic_debug_info(out=None):
         pass
     from calibre.customize.ui import has_external_plugins, initialized_plugins
     if has_external_plugins():
-        names = (p.name for p in initialized_plugins() if getattr(p, 'plugin_path', None) is not None)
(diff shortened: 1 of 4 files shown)


kovidgoyal:
You might want to have the config dialog remember its size or at least have a larger default size.

davidfor:
The configuration widget is contained inside the device configuration dialog. I have never noticed before that the size of this is not stored when it is opened from the "Configure this device" option. Should that be storing the size?

For a default size, I'll have to test on a lower resolution machine to see what's best.

kovidgoyal:
IIRC, the default config dialog is simple enough that I never bothered implementing size storing for it. You'd need to implement it in there. No harm having it for all device config dialog types.

kovidgoyal:
IIRC, the default config dialog is simple enough that I never bothered implementing size storing for it. You'd need to implement it in there. No harm having it for all device config dialog types.

PR #522, opened 2016-07-11
Fixes Network rename/move issues with windows and servers

NathanaelA:
I realize you don't officially "support" network locations; and that the current edition of Calibre is single user.   But for those of us who are willing to run this as a single user hitting a network share; this fixes the most common reason for the error -13 permissions issue when changing a Author/Title on a network share from a windows machine.

### DIFF
--- a/src/calibre/utils/filenames.py
+++ b/src/calibre/utils/filenames.py
@@ -404,8 +404,10 @@ class WindowsAtomicFolderMove(object):
                         ' operation was started'%path)
             else:
                 raise ValueError(u'The file %r does not exist'%path)
+
+        hardlink = get_hardlink_function(path, dest)
         try:
-            windows_hardlink(path, dest)
+            hardlink(path, dest)
             return
         except:
             pass


kovidgoyal:
I'm afraid I'm an not going to merge this as it has performance implications for every calibre user to support a use case that is not actually supported.

NathanaelA:
@kovidgoyal If I benchmark it and create a patch that is either equal or faster; would that be acceptable?

kovidgoyal:
Sure, a patch that has no negative side effects is acceptable. But, be aware that this is an extremely critical code path, so the bar for demonstrating that a patch has n negative side-effects is very high.

kovidgoyal:
FYI, there is a calibre test suite, that you can run to check your patch for regressions, 

python setup.py test --test-module db

kovidgoyal:
Or if you are running using CALIBRE_DEVELOP_FROM, the tests can be run as

calibre-debug -c "from calibre.db.tests.main import *; run_tests(find_tests)"

NathanaelA:
Sweet; thanks for the pointers.   Yeah; I'm using CALIBRE_DEVELOP_FROM on my local box; because to me even if there is a speed hit here, it is totally worth it as I can now rename and fix books on my server, where before Calibre was totally corrupting any of the actual files involved.

NathanaelA:
I totally forgot to follow up last year; this patch has about a 30-40% speed hit while running all the tests.   So you are correct it does slow things down some; but I am willing to live with the speed hit and not have all my files corrupted when they get renamed.  ;-)     To me data safety is way more important. 

When I have spare time I might see if I can revamp that whole part of the code base as it has several things that are kinda fragile and can cause data corruption.

kovidgoyal:
Let me note that on any filesystem that is not broken calibre is 100% data safe. There is nothing in the code base that is either fragile or that can cause data corruption, on a well behaved filesystem. 

Network filesystems are broken. Putting lipstick on that pig is not going to change its pucker. Filesystem semantics and network semantics simply do not fit together well.

And note that calibre 3.0 is just around the corner which contains a rewritten content server that will be read/write (though the write part will not be turned on immediately for calibre 3.0) so you can stop relying on networked filesystems for syncing calibre libraries. Your efforts would be better spent contributing to that, rather than trying to make network filesystems work.

Having said all that, I'll re-iterate that I wont refuse a patch that has no negative side-effects on the standard use case but does improve the NFS use case.

PR #523, opened 2016-07-14
Bug #1602908: csv catalog not formatting is_multiple custom columns …

cbhaley:
…correctly


cbhaley:
Note that the fix removes the current bracketing (the [] around the values). I did this because it is what the module does for tags.

cbhaley:
Also note that is_names will be separated by commas, because (again) that is what the module does for the built-in authors column.

kovidgoyal:
I have pushed a commit to use & for authors

PR #527, opened 2016-07-17
Clear search_restriction combobox when 'Close VL' button pressed.

JimmXinu:
'Close VL' button removes restriction, but doesn't reset the search_restriction combobox.

View Manager PI compares the search_restriction combobox current value and if it hasn't been cleared, the PI doesn't behave correctly.


JimmXinu:
This is literally a one line change.  Is there some objection to it?

So sorry about all the extra commits listed--I didn't realize it would do that.  If you want, I can make a different PR with just the one change in it.

kovidgoyal:
It is already merged. GitHub probably didn't pick up that it is merged because I rebased it to get rid of all those extra commits.

JimmXinu:
Okay, thanks for explaining.  I'll try to avoid the extra commits in future.

PR #532, opened 2016-07-23
Add a boolean "show on left in book details" to the comments-type …

cbhaley:
…custom column

Kovid, this was easy enough to do that I went ahead with it, which removes any ambiguity about what I was talking about. No problem if you think it is a good idea.

### DIFF
--- a/src/calibre/ebooks/metadata/book/render.py
+++ b/src/calibre/ebooks/metadata/book/render.py
@@ -98,9 +98,16 @@ def mi_to_html(mi, field_list=None, default_author_link=None, use_roman_numbers=
                         val = markdown(val)
                     else:
                         val = comments_to_html(val)
+                    add_to_comment = True
                     if disp.get('show_heading'):
-                        val = '<h3 class="comments-heading">%s</h3>%s' % (p(name), val)
-                    comment_fields.append('<div id="%s" class="comments">%s</div>' % (field.replace('#', '_'), val))
+                        if disp.get('show_on_left'):
+                            ans.append((field, row % (name, val)))
+                            add_to_comment = False
+                        else:
+                            val = '<h3 class="comments-heading">%s</h3>%s' % (p(name), val)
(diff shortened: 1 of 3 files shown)


cbhaley:
BTW: I removed the inheritance of create_custom_column_ui but I did not remove the file. I *think* it should go, but I am not sure.

kovidgoyal:
I implemented it slightly differently (i found show_on_left to be rather confusing). Also, the server logic was incorrect (the comments are pre-processed to always be HTML unless they are of type short-text). Finally, I removed the auto change of the checkboxes in the preferences dialog.

cbhaley:
Three comments:

1) Given that only 3 of the possible 4 combinations of show_on_left and show_heading are meaningful, perhaps show_heading should be a 3-valued string set using a combo-box?

2) As it is, short-text implies show_heading. Is this correct/necessary? I don't see any problem with showing short-text on the right if that is what the user wants to do. The new server behaves this way, showing short-text in the comments section if show_heading is not set.

3) Another new-server issue: if a short-text is shown in the comments section then any html contained in that short-text is interpreted when it should not be. I think that the server-side preprocessor should do the same thing with short-text that it does for long-text. If this is done then the is_html parameter on add_row (line 205) would be set to True. I tested these two changes and the server seems to work correctly.

Let me know if you want me to make these changes and push them with a new PR.

kovidgoyal:
(1) Sounds good. make show_heading one of: 'above', 'side', and 'hide' and get rid of show_on_left

(2) No not necessary, just easier in the original implementation. I have no problems with having show_heading gain exclusive control of heading position. 

(3) I'm ok with having the server also pre-process short-text, but thinking about it for both short and long text we should probably not use <pre>, but instead <pre style="white-space:pre-wrap"> for long text and just a <span> for short-text. The idea being that we want line wrapping since book details can be displayed in fairly narrow spaces. 

Go ahead, and make the changes.

PR #535, opened 2016-08-07
Append final replace function output to marked text (#1609111)

Mymei2:
Append final user replace regex-function output to marked text (see #1609111)

Introduces a new flag "append_final_output_to_marked".
Depends also on existing flag "call_after_last_match".

Following two lines at the end of a user replace function will cause,
that the return value of the last additional call (after all replacements
are completed) will be appended to the marked text:
replace.call_after_last_match = True
replace.append_final_output_to_marked = True

TODO: Documentation update is required.
Part: Editing books
Chapter: Function Mode for Search & Replace in the Editor
Section: Having your function called an extra time after the last match is found

### DIFF
--- a/src/calibre/gui2/tweak_book/editor/text.py
+++ b/src/calibre/gui2/tweak_book/editor/text.py
@@ -313,7 +313,12 @@ class TextEdit(PlainTextEdit):
             raw, count = pat.subn(template, raw)
             if repl_is_func:
                 from calibre.gui2.tweak_book.search import show_function_debug_output
-                template.end()
+                if getattr(template.func, 'append_final_output_to_marked', False):
+                    retval = template.end()
+                    if retval:
+                        raw += unicode(retval)
+                else:
+                    template.end()
                 show_function_debug_output(template)
             if count > 0:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Looks fine, but you should add anote about it in edit.rst, otherwise no on will know it exists.

Mymei2:
Well, this is actually the TODO I've mentined above. I didn't know that I can do it myself. :-)

I've found the file in the "manual" folder and I see that it's actually the calibre online manual content.
However, I do not know this particular formatting syntax. What is it? Is there somewhere some documentation for it?

It doesn't seem to be terrribly complicated, so I'll to try my best, but I'd like to double-check before publishing it. :-)

kovidgoyal:
Its restructured text. THe format used for python's documentation.

Mymei2:
Thanks. I'll plan to do it in the next two days. However, only in the english version. What about the other languages? How is that stuff maintained? Should I open some TODO somewhere?

Mymei2:
Oops! I've just found in repository that you've already documented it, it's simply not online yet. Thanks once more. Now I can move to the second change. :-)

PR #538, opened 2016-08-31
Fix ignore collection deleting and bump firmware version.

davidfor:
I forgot to test the ignore collections and delete empty collection option together. This fixes that.

Plus, have confirmed the firmware for the new Aura One in the 3.20.x series.


kovidgoyal:
I made a small change in how the value of ignore_collections_names is tested so that it does not blow up for None

PR #539, opened 2016-09-02
Kobo decided the new firmware should be 4.0.

davidfor:
I think the title says it all.


kovidgoyal:
Do you know when this firmware/devices with thi firmware will become generally available?

davidfor:
It became available late yesterday. Probably about the time you were building 2.66. But, it is only being released to the two new devices. It might come out to the other devices, but I think Kobo will do another build in a month or so.

PR #541, opened 2016-09-08
metadata: Do not raise an exception if a PDF file has empty metadata

aroig:
Check explicitly for `info is None`, since an empty dict also evaluates to false. Otherwise it raises an exception when there is no metadata, and prevents from generating the cover.

This has started to happen recently with some arxiv PDF's, like this one: https://arxiv.org/pdf/1605.02391v2

### DIFF
--- a/src/calibre/ebooks/metadata/pdf.py
+++ b/src/calibre/ebooks/metadata/pdf.py
@@ -95,7 +95,7 @@ def get_metadata(stream, cover=True):
             raw = f.read().strip()
             if raw:
                 prints(raw)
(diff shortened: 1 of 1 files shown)


aroig:
I've just realized that it may have nothing to do with arxiv pdfs. The timing coincides with an upgrade to poppler 0.47. pdfinfo may have changed its output.

Anyway, thanks!

kovidgoyal:
Yes poppler >= 0.47 changed the output from pdfinfo in a non-backwards comaptible way, sigh.

PR #545, opened 2016-09-13
Don't clear the user-defined function list until after the DB is …

cbhaley:
…actually closed. This order prevents problems in plugins that run just before closing the db, e.g., library closed plugins.

It is possible that a better fix is to move the two lines of code into db.cache.close(). I looked for unwanted side effects and didn't see any, but I didnt want to make the more invasive change.

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -31,7 +31,8 @@ from calibre.utils.filenames import (
     WindowsAtomicFolderMove, atomic_rename, remove_dir_if_empty,
     copytree_using_links, copyfile_using_links)
 from calibre.utils.img import save_cover_data_to
-from calibre.utils.formatter_functions import load_user_template_functions
+from calibre.utils.formatter_functions import (load_user_template_functions,
+                                               unload_user_template_functions)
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Where are they loaded? If they are laoded in the db layer I see no problem with moving unloading intot he db layer as well.

cbhaley:
They are loaded in three places:

- db.backend.init as long as load_user_formatter_functions is true

- in gui2.preferences when template functions are defined or changed

- in gui2.save (Saver). I am not sure why this is required. I would have thought they would be loaded when the db was opened but I don't know this code path.

I am reasonably sure that moving the unload to close() is OK. One question is "Which close", cache.close() or backend.close(). As they are loaded in backend it seems reasonable to unload them there as well.

BTW: there is no problem if the functions get unloaded more than once.

kovidgoyal:
gui2.save is used for save to disk, presumably the functions from the current db are passed to that code. I am OK with doing the close in backend.py

cbhaley:
I pushed the changes.

I am also working on BetterRed's bugs. I won't push those until this pull request is merged, to avoid mixing up the commits.

PR #547, opened 2016-09-13
In bulk and single edit, make the initial "up value" operation on an …

cbhaley:
…undefined numeric field set the value to zero instead of -999999


cbhaley:
This fixes bug #1622757

PR #563, opened 2016-10-07
Fixed various recipes that had broken feed links.

CoderAllan:
Hi Kovid
I have searched through all the recipes and alot of them has broken/dead feed links.
So this is the first batch of fixed recipes. More will follow over the next weeks.
/Allan


kovidgoyal:
Cool, thanks. It is likely though that some of the recipes will in general need more work than just replacing feeds. News websites often change their markup. And recipes that do not use auto_cleanup will likely need adjustment.

PR #571, opened 2016-10-12
Stop cover metadata merge from dying on invalid img URLs

kemayo:
If an ebook contains a bad img tag, whose src looks enough like an IPv6 URL to
trip that codepath, but which isn't actually an IPv6 URL, it'll throw a
ValueError.

Fixes launchpad 1628969.


kovidgoyal:
IMO, code to remove old cover images should not have the side-effect of also removing imgs with invalid URLs in the src attribute. SO I pushed a slightly different fix for this.

PR #581, opened 2016-10-30
lwn_weekly: Prevents fail if article title is not defined.

marvil07:
More resilient behavior if article title contains non-html-encoded
characters. E.g. "<Programming>" string in the title start is recognized
by beatiful soup as html tag, but it is not, making the title a
NoneType, and raising an exception on Article creation.

This is actually a bug in the news page itself, but it anyways breaks
the recipe; hence, adding here a fixed string if for any reason the
article title end up with a NoneType.


marvil07:
Thank you for adding this upstream!

kovidgoyal:
You're welcome :)

PR #583, opened 2016-11-04
calibre: port to html5lib-python 0.999999999

MingcongBai:
And yes, that is nine zeros in total.

A quick port according to changelog found [here](https://pypi.python.org/pypi/html5lib/0.999999999).

### DIFF
--- a/src/calibre/ebooks/oeb/polish/parsing.py
+++ b/src/calibre/ebooks/oeb/polish/parsing.py
@@ -13,8 +13,8 @@ from bisect import bisect
 from lxml.etree import ElementBase, XMLParser, ElementDefaultClassLookup, CommentBase, fromstring, Element as LxmlElement
 
 from html5lib.constants import namespaces, tableInsertModeElements, EOF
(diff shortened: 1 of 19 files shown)


eli-schwartz:
What is the point of this? It only changes the import paths...

If you are trying to build linux distro packages which use the latest version of upstream html5lib, don't. The version that calibre bundles was forked in order to apply certain fixes and changes that upstream html5lib didn't want (in addition to a couple things they [recently merged](https://github.com/html5lib/html5lib-python/pull/245)). But in the current state of affairs, trying to use upstream html5lib in calibre leads to failing ebook conversions (which is kind of one of the main reasons to use calibre...).

For context and linked examples, a bug report I filed in Arch (which still hasn't upgraded python-html5lib because it "breaks calibre"): https://bugs.archlinux.org/task/43382

kovidgoyal:
upstream html5lib is not suitable for calibre

MingcongBai:
Well if that's the case this is the end of my argument, sorry for the trouble.

PR #591, opened 2016-12-05
Minor setup fixes

jamesbroadhead:
- make `setup.py check` a little more portable
- create target dirs in `setup.py get_translations` for use in a fresh
  clone


kovidgoyal:
This PR is breaking setup.py check on travis. https://travis-ci.org/kovidgoyal/calibre/jobs/181298343

Probably you need to do from distutils.spawm import find_executable (as is done in build_environment.py

kovidgoyal:
Closing this as both the pot and check commands are meant to be run only by me as part of releasing calibre -- so there is no real need to make them more general.

PR #592, opened 2016-12-05
Run, fix calibre.library tests

jamesbroadhead:
Run the calibre.library tests as part of `setup.py test`. 
Fix testSorting so that it passes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .bzr
 .bzrignore
 .build-cache
+.cache
 src/calibre/plugins
 resources/images.qrc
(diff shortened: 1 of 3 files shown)


kovidgoyal:
The calibre.library module is obsolete the db part is replaced by calibre.db and the server part by calibre.srv. It only exists because it is used by the tests in calibre.db to test for regressions against the old db API. As such, it's own test shoud not be run, in fact you can delete calibre.library.test instead of running it.

jamesbroadhead:
Ok, will include with my other PR

PR #593, opened 2016-12-05
Use the csv module in calibre.library.cli for check_library

jamesbroadhead:
- use csv.writer for the output of check_libary call in calibre.libary.cli
- tests before & after this change
- add the new test suite to test discovery
- delete c.library.test (deprecated, per #592)
- adds a new dep, the 'mock' library


kovidgoyal:
You'll need to install mock explicitly in .travis.yml (and probably appveyor.yml as well) otherwise the CI builds will fail. If I were you, I'd just add the bits of mock that you actually use to the calibre testing infrastructure, but it's up to you which approach you prefer. I'm OK with depedning on mock for the test suite.

jamesbroadhead:
How can I trigger an appveyor build?

PR #598, opened 2017-01-02
Updating for new site layout & combining Sunday ed

bobbysteel:


bobbysteel:
I suggest also deleting the Sunday Times recipe as that's no longer required. It may be preferable to include the Sunday times Magazine as a separate recipe, which I will do if desirable.

PR #600, opened 2017-01-15
missing Canada option for amazon download plugin

botmtl:
From what I saw of the rest of the code, amazon_ca is all good and ready to go, it's only missing this option in the UI. 
http://i.imgur.com/GKqtDLp.png

### DIFF
--- a/src/calibre/ebooks/metadata/sources/amazon.py
+++ b/src/calibre/ebooks/metadata/sources/amazon.py
@@ -800,6 +800,7 @@ class Amazon(Source):
             'br': _('Brazil'),
             'nl': _('Netherlands'),
             'cn': _('China'),
+            'ca': _('Canada'),
     }
 
     options = (


kovidgoyal:
Does it actually work? and if so could you add a couple of books to test it with at the bottom of the file, like there are for all the others.

botmtl:
I tought everything was perfect but the tests reveal that the series information, when available in the canadian counterpart, is not picked up.  I saw that /gui2/store/amazon_ca_plugin.py has existed for a while, any relation to what I'm trying to do here?

kovidgoyal:
No, those are for get books, not for metadata download. Series parsing happens in parse_series in amazon.py, you probably just need to adjust it a bit to handle whatever variation of markup amazon.ca uses.

botmtl:
I cannot find a single book on amazon.ca that has series information in it, so closing for now.

kovidgoyal:
It's fine to not have series information if the site does not have it. Many of the country specific sites dont have series info.

PR #604, opened 2017-01-19
Adapt to html5lib 0.99999999/1.0b9

kleink:
When updating html5lib in [pkgsrc](http://cdn.netbsd.org/pub/pkgsrc/current/pkgsrc/misc/calibre/) recently, we observed the calibre build being broken due to changes in the former's module naming:

* moved treebuilders._base to .base to clarify its status as public, and
* moved and ihatexml to _ihatexml to clarify its status as private.

Since there's no (version) dependency recorded in setup.py, I added a local patch to deal with these gracefully using the try: new/except: old pattern.

### DIFF
--- a/src/calibre/ebooks/oeb/polish/parsing.py
+++ b/src/calibre/ebooks/oeb/polish/parsing.py
@@ -13,8 +13,14 @@ from bisect import bisect
 from lxml.etree import ElementBase, XMLParser, ElementDefaultClassLookup, CommentBase, fromstring, Element as LxmlElement
 
 from html5lib.constants import namespaces, tableInsertModeElements, EOF
-from html5lib.treebuilders._base import TreeBuilder as BaseTreeBuilder
-from html5lib.ihatexml import InfosetFilter, DataLossWarning
+try:
+    from html5lib.treebuilders.base import TreeBuilder as BaseTreeBuilder
+except:
+    from html5lib.treebuilders._base import TreeBuilder as BaseTreeBuilder
+try:
+    from html5lib._ihatexml import InfosetFilter, DataLossWarning
+except:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
calibre has its own fork of html5lib (see src/html5lib), dont use the upstream version for it -- there is an issue in the html5lib bug tracker detailing the reasons for the fork.

kleink:
Thank you. I'll check why the external package got picked up during our build.

eli-schwartz:
#583 was there before you (rejected for the same reason).

Someone probably tried to overzealously unbundle vendored dependencies, see:
```
# Let's depend on the ones we have in pkgsrc instead of the bundled ones.
pre-configure:
	cd ${WRKSRC} && rm -rf src/cherrypy src/pyPdf src/html5lib src/chardet
```

Similarly, cherrypy upstream has open bug tickets (ancient, uncommented ones) for:
e298f23781ffb0544bfe11a32ada02c906f1d0ca and later commits dealing with the same problem
b117148241c0279c95b07637c3b6d7a27523b894

pyPdf was removed from calibre in b69fb230c5966f03a87058550574b102e18410ac (4.5 years ago)

kleink:
@eli-schwartz thank you for elaborating. That's the result of pkgsrc's policy having been applied very generously, which (among other things) aims at having to fix possible security issues once and prevent maintainers having to take care of each and every bundled instance of an issue. That being said, calibre obviously makes a case for such bundling.

I'll sort these sort these out as time permits (as time permits; I'm not actually affiliated with the pkg), e.g. https://github.com/jsonn/pkgsrc/commit/a202978616045e1eb18e5c1b4369f989860a3a34.

PR #611, opened 2017-01-23
Appledaily patch 20170123

Kennyl:
- Better translation on nav bar
- Better Cover Image
- New Download Logic
- Change Font Size


kovidgoyal:
Note that setting base font size in conversion_options is not a good idea as it makes it impossible for the user to override. If you need to rescale the font sizes, use extra_css instead.

Kennyl:
Okay , in case of this, I will await users' feedback for any rollback/ enhancement.

PR #612, opened 2017-01-23
Fix Video Player Space

Kennyl:
Fix flow-player parent div to display none


Kennyl:
remove_tags won't work for remove div with style criteria, so plaintext replace is done.

Kennyl:
Thanks for prompted action.

PR #620, opened 2017-01-26
Calibre: Haiku support

miqlas:
Haiku is a special snowflake, so i had to change some sources here and there.

Calibre compiles and install fine already, the gui starts but bails out because there is no netifaces yet, i have to port it.
The locking mechanism isn't my world, so i let the fun part for somebody else. MAybe we can piggyback on other implementation, but i had no time to fight with them. The islinux path surely doesn't works.

Let me know if you have better ideas than me.

Proof:
calibre


kovidgoyal:
For the locking you should try the default version, it might work on haiku. IIRC linux uses abstract network sockets which are not widely available outside of linux land. The default version uses fcntl file locking, I'd be surprised if haiku does not support that.

miqlas:
@kovidgoyal : i'm sure there is a proper way to do this, but as i'm not a developer, i cannot provide much better code.
Sorry :S

miqlas:
Now with disabled mtp and usb it starts fine, but the ebook-viewer crashing at start. I'll provide more info later.

kovidgoyal:
This PR is causing the building of calibre to fail, see https://travis-ci.org/kovidgoyal/calibre/jobs/195599049

presumably one of the commits changed something in the way pkg-config is searched for.

setup/build_environment.py
```diff
@@ -37,7 +37,7 @@
 
 PKGCONFIG = find_executable('pkg-config')
 PKGCONFIG = os.environ.get('PKG_CONFIG', PKGCONFIG)
-if islinux and not PKGCONFIG:
+if islinux or ishaiku and not PKGCONFIG:
```
kovidgoyal: You are missing brackets around islinux or ishaiku here

miqlas:
@kovidgoyal : thanks for your help. I already pushed the changes, and some other brackets too, as i'm not sure if they needed actually or not. If not, we can easily revert the last commit.
Let me know if you need any changes.

kovidgoyal:
Those are not needed. Basically you only need brackets if you are combing different boolean operators like and/or/not in one expression.

miqlas:
@kovidgoyal : Thank you.

PR #622, opened 2017-02-01
remove video captions

jfhutson:
This removes the captions from the videos which are stripped out.


kovidgoyal:
There are actually two nytimes recipes, nytimes.recipe and nytimes_sub.recipe, the changes should be made to both.

PR #624, opened 2017-02-03
remove 'masthead-social' tag

jfhutson:
this shows up on blogs like "The Upshot". See the source of https://www.nytimes.com/2017/02/01/upshot/strife-over-immigrants-can-california-foretell-nations-future.html?ref=todayspaper. Without the fix "The Upshot" always has "Follow Us" and then three blank bullet points at the top then "Get the Upshot in your Inbox".


kovidgoyal:
Again should be made to nytimes_sub.recipe as well

PR #634, opened 2017-02-22
Improve KFX metadata handling

j-howell:
Support getting metadata from sideloaded KFX files by treating the main KFX file of a book as the actual book path instead of the 'metadata.kfx' file. Read metadata.kfx only if the main book file is encrypted.

Also improve KFX metadata reading in general.

### DIFF
--- a/src/calibre/devices/kindle/driver.py
+++ b/src/calibre/devices/kindle/driver.py
@@ -38,10 +38,6 @@ file metadata.
 '''
 
 
-def get_kfx_path(path):
-    return os.path.dirname(os.path.dirname(path)).rpartition('.')[0] + '.kfx'
-
-
 class KINDLE(USBMS):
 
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I have merged this, but I have a couple of questions:

1) How are non-sideloaded kfx books still detected?
2) Since you removed delete_single_book are non-sideloaded kfx books still deleted?

j-howell:
1. Non-sideloaded books are detected by looking for .kfx files that are not in a .sdr directory. The metadata is retrieved for them by locating the associated metadata.kfx file if the main book file is found to be encrypted.

2. Non-sideloaded KFX books are now handled the same as any other kindle book. After the main book file is deleted the matching .sdr folder is also deleted by delete_extra_book_files in usbms/driver.py.

kovidgoyal:
OK thanks.

PR #635, opened 2017-02-22
Add Message-Id header to outgoing emails.

ThisIsADogHello:
Some MTAs/spam filters/etc seem more likely to treat an email as spam if it has no Message ID. 

This patch should hopefully make it less likely for an email to be dropped .

### DIFF
--- a/src/calibre/utils/smtp.py
+++ b/src/calibre/utils/smtp.py
@@ -18,16 +18,23 @@ def create_mail(from_, to, subject, text=None, attachment_data=None,
     assert text or attachment_data
 
     from email.mime.multipart import MIMEMultipart
-    from email.utils import formatdate, make_msgid
+    from email.utils import formatdate
     from email import encoders
+    import uuid
 
     outer = MIMEMultipart()
     outer['Subject'] = subject
     outer['To'] = to
     outer['From'] = from_
     outer['Date'] = formatdate(localtime=True)
-    outer['Message-Id'] = make_msgid()
     outer.preamble = 'You will not see this in a MIME-aware mail reader.\n'
+
+    # generate a Message-Id for this email
+    msgid_domain = from_.partition("@")[2]
+    if not msgid_domain:
+        # from address didn't provide a domain, let's make a best guess
+        msgid_domain = safe_localhost()
+    outer['Message-Id'] = "<{0}@{1}>".format(uuid.uuid4(), msgid_domain)
 
     if text is not None:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On my system that results in message-ids of the form
'<148781614594.22889.14281208660896828881@localhost.localdomain>'

I would not be surprised if spam filters purposely reject mails with message-id domains of localhost. Might be better to replace it with a fake domain in that case. Something like
```py
re.sub(r'@localhost[^>]+', '@{}.com'.format(random_string_of_ascii_letters), msg_id)
```

kovidgoyal:
Actually, rather than a random string, probably better to use the calibre installation uuid, like this
```py
from calibre.utils.config_base import prefs
hostname = prefs['installation_uuid'].replace('-', '')
```


That way it will stay the same for all emails sent from that calibre installation.

kovidgoyal:
Another option is to use the domain specified in from_

ThisIsADogHello:
Using the domain from from_ seems like the best option.

Looks like the way that that Thunderbird (the mail client that I've been using) generates its message IDs is with a UUID@[from domain].  Using a UUID there seems like the easiest way to make sure that the ID is unique, too.

kovidgoyal:
UUID@from domain sounds fine to me. Just be careful when parsing out the from domain, IIRC that can be  of the form "user name" <address> or just address or anything really. So good to have a fallback for when parsing fails to UUID@installation_id.com

ThisIsADogHello:
Alright, this should do a lot better for generating message IDs.  When the from address has a domain in it, they look more or less like any other legit mails', and it's using the safe_localhost function used for figuring out what to HELO with when it has nothing else to go on for a domain.

Although when I was testing with sending emails from a username only, it was also breaking SPF and DKIM, so hopefully a less than ideal message ID in those situations would be the least of a user's problems.

kovidgoyal:
Maybe use https://docs.python.org/2/library/email.util.html#email.utils.getaddresses to get more robust extraction of domain embedded in from_?

ThisIsADogHello:
I ended up using parseaddr instead of getaddresses, but it seems to be pretty robust now at parsing out the domain from some malformed inputs for From.  As long as it can make out a domain on the end, it does a pretty good job.  Had to handroll some fixes for it though, as getaddresses/parseaddr only split out the real name and email address parts of a domain, and couldn't find anything that would reliably pull out the domain itself.

PR #638, opened 2017-02-26
Use QSystemTrayIcon as a fallback

Arkq:
If one is using Linux and StatusNotifier spec is not available, use
QSystemTrayIcon as a fallback.

### DIFF
--- a/src/calibre/gui2/dbus_export/widgets.py
+++ b/src/calibre/gui2/dbus_export/widgets.py
@@ -238,8 +238,7 @@ class Factory(QObject):
             self.prune_dead_refs()
             self.status_notifiers.append(weakref.ref(ans))
             return ans
-        if iswindows or isosx:
-            return QSystemTrayIcon(parent)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
The whole reason that I went to the trouble of creating my own system tray icon implementation for linux is because the Qt system tray icon implementation was buggy -- it would lead to missing or hidden or non-functional tray icons or crashes in various circumstances. https://bugreports.qt.io/browse/QTBUG-31762 or https://bugreports.qt.io/browse/QTBUG-55922 or many others, google QSystemTrayIcon and qtbug

Arkq:
OK, so basically it means, that there is a requirement for the StatusNotifier spec, and if it's not available (some light, Linux-based desktop) it is not possible to keep calibre on the tray (if System Tray is supported)?

kovidgoyal:
Yes. Though there are plenty of light linux desktops that support statusnotifier. For example, I use one myself, qtlie.

PR #652, opened 2017-05-04
Don't raise exception if there are no feeds

Sophist-UK:


kovidgoyal:
If you want to specialize the handling of get_feeds in your recipe simply override it. The rest of the new download system relies on that method raising NotImplementedError when no feeds are available.

PR #653, opened 2017-05-05
Switch author link to Goodreads.

Sophist-UK:


kovidgoyal:
Hmm I'm somewhat sceptical about changing the default link from a public resource like wikipedia to a corporate website like Goodreads. What's the motivation for the change? Does goodreads have much more author information than wikipedia?

Sophist-UK:
Goodreads has some commercial sides, but it is not Amazon.

I would class Goodreads as being to books what MusicBrainz is to music - i.e. a definitive source of information about authors and their books.

In my experience, Wikipedia is missing the majority of authors - perhaps it has the most famous (say) 20%, but Goodreads has the vast majority of authors. Why not run from source and give it a try against some better and lesser known authors.

kovidgoyal:
Yeah and if someday amazon goes under or decides to monetise goodreads, or decides that maintaing it is no longer profitable all that data will disappear. I am pretty scpetical of the longevity of commercially backed data.

Still I am not adamantly opposed to changing the default template -- as for trying it out there is no need to run from source for that, simply change the tempate in Preferences->Look & Feel ->Book details. 

In my experience wikipedia covers over 80% of the authors I read -- perhaps I dont read sufficiently obscure authors :)

Sophist-UK:
1. Every large web site has costs - Goodreads is no different, so it shows paid adverts. That is a world of different cf. Amazon which is a sales site pure and simple. Of course things could change in the future. Goodreads might become more commercial, but then again so might Calibre. Amazon could be bought out by Wikipedia and turned into an information site about books and music - very unlikely perhaps, but it could possibly happen.

I don't personally see the issue with linking to them.

2. However I was unaware that you could change the link through user preferences - so please feel free to close this PR.

kovidgoyal:
1. The purpose of wikipedia is not too maximize its profits, neither is that the purpose of calibre. That is however, the purpose of Amazon (as it is of every publicly traded company)

2. As I said I am not adamantly opposed to changing the default, I just need more evidence that one is actually better than the other. I will check them out over the next few days and decide if making the change is worthwhile.

Regarding my longevity concern, it is not a big deal in this context, since if goodreads does go under, the default can always be changed.

PR #662, opened 2017-05-12
Updating for reduced image sizes

bobbysteel:
Changes to the website have greatly increased image sizes, so implementing in-built filesize reduction through image compression.

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -90,6 +90,9 @@ class Economist(BasicNewsRecipe):
     delay = 1
 
     needs_subscription = False
+    useHighResImages = False
+    compress_news_images = True
+    compress_news_images_auto_size = 5
 
     def get_browser(self):
         br = BasicNewsRecipe.get_browser(self)


kovidgoyal:
I dont like the idea of compressing news images unconditionally in recipes. As time passes, screens become higher resolution. So having higher resolution images in the source is a good thing. About the only legitimate use for post-download image compression I can think of, is amazon's arbitrary file size limits for emailing books. Given that emailing news files also destroys periodical formatting, I am not particularly sympathetic to this use case.

bobbysteel:
Well right now the current recipe delivers a 60MB file which is too big for
sending to Kindle in any circumstance.

bobbysteel:
And the new file w/ these flags looks entirely the same as before. No
changes in result.

kovidgoyal:
What nonsense, the Kindle can handle 60MB files perfectly well. Just send it using a USB connection.

bobbysteel:
Yes it can handle but who uses a wired cable to load files? Their wireless
sending is the key feature.

kovidgoyal:
Who uses email delivery -- its a horrible feature that is limited in size, accepts only old style mobi files, causes periodical formatting to be destroyed and fails randomly because of amazons spam controls. 

You want wireless delivery use the content server the kindle browser can download mobi files from it just fine.

bobbysteel:
Haha, tell that to Amazon. They accept via email or their SendToKindle
uploader which caps at 50MB. It works flawlessly and has for me and many
friends for many years. I suspect many of your users for news would use the
email-to-kindle versus setting up their own content server. A content
server is needless for almost any Kindle user.

You seem to be taking this very personally. It's an Amazon construct, not
dumb users.

kovidgoyal:
You seem to be confused. I have told you why I dont like the idea of forcing everyone to use lower quality images to accomodate amazon's brokenness, in the very first post. You are responding with telling me that it's amazon's fault. I know that already. Once again,

I dont want to make all calibre users suffer for something that is amazon's fault. 

And there is no "setup" for a content server. It's a single button to click in calibre.

kovidgoyal:
What I am willing to do is turn on the compression conditionally if the output profile is set to a kindle profile.

bobbysteel:
i take your point but a content server requires at a minimum a PC on 24/7
and a firewall configuration.

shrug. doesn't really matter as i have my own custom recipe, but i expected
this would help many more of your users that are on Kindle. i'll just go
away now.

bobbysteel:
that's very sensible.

bobbysteel:
i'm happy to code and submit a PR for that if you can point me to where the
conditional logic for this is similarly used. :)

kovidgoyal:
Dont worry, I'll take care of it :)

bobbysteel:
Thanks Kovid. As always I (and many others) highly appreciate all your work!

PR #663, opened 2017-05-12
Duplicate of economist.recipe

bobbysteel:
No need for this - at one time there was a need for the "free" recipe but this has not been the case for a while. This recipe seems to be totally duplicate and at best should be renamed as currently there are two identical recipes with the same official title.


kovidgoyal:
It remains because removing it will break automated news download for people who have subscribed to that recipe in the past.

bobbysteel:
can I suggest renaming one of them then? it's v confusing to have two identical recipes. i'd propose naming this one "The Economist (Free)" for consistency w/ origins.

kovidgoyal:
Since they produce identical output, it would be more misleading to change their titles, since if you name one (Free) and not the other, it would lead people to expect that the other is not free -- which is currently not true.

bobbysteel:
Don't you think it's ridiculous though to have duplicate names that require
code duplication?

kovidgoyal:
No, I dont. I have explained why above.

PR #664, opened 2017-05-13
Add a language field to the "get ebooks" dialog.

Steap:


kovidgoyal:
There's a whole bunch of other amazon plugins you should make similar changes to. See the parse_language() function in sources/amazon.py 

Also I thing the language field should contain cacnonicalized langauge names whereever possible, so that the displayed language is the same from all stores. See the get_langauge() and canonicalize_lang() functions in localization.py

The plugins should return ISO lang codes and the GUI should display them as localized language names.

kovidgoyal:
You should not need to maintain your own mapping of german to english language names, it should be there already, see langnames_to_langcodes() in localization.py

Unless of course, libri.de uses different langnames than the ones in the german translation of calibre.

kovidgoyal:
Though I suppose if you want it to work with non-german calibre, you have to maintain your own mapping. So feel free to ignore my previous comment :)

Steap:
OK, thanks for the tips! I'm going to drop the non-English stores for now, I'm more interested in pushing the new feature. I'll update the PR with a fix for all English Amazon stores.

kovidgoyal:
Unfortunately, you cannot re-use code in that fashion for the amazon plugins. The reason is that store plugins are auto-updated and reloaded dynamically (to keep up with changes to websites). That means they cannot depend on each other. So you'd have to duplicate the code across all amazon plugins -- it's annoying, but...

Volkthehunter:
Where is code that makes some text able to jump through pages when clicked? Thank you!

snow212-cn:
Is the coronavirus in India still not serious enough to kill the author?

snow212-cn:
Now is 12739. It'll double day by day and may you be gone soon.

Rockster1501 (approved):
Go go go

Rockster1501 (approved):

breakndream:
For this bug: Amazon meta data plugin should use current locale by default(https://bugs.launchpad.net/calibre/+bug/1256374), can we just change line 176 in the localization.py file to  `lang = get_system_locale()`? 
If not, I am curious how to implement that feature since from other methods, there were various ways to receive user language and that's the dead end..

PR #666, opened 2017-05-15
Store and access user template functions on a per-library basis. This…

cbhaley:
… is used when the DB evaluates composites, avoiding using the global copy.

The ability to pre-compile templates was removed because the one-true list of available functions is known only at execution time. The compiled copy depended on the global state being accurate. And in any event, I was never totally convinced that the compiled version was significantly faster than the pre-lexed version.

Includes a bug fix: the booklist must be refreshed if a user-defined function changed.


kovidgoyal:
I merged it with a slight change to simply the set_global_state function

PR #667, opened 2017-05-16
Two changes: format_number and improve performance of per-library template functions

cbhaley:
Two commits:

1) Improve the format_number template function, permitting leaving off the {} characters when using basic templates. This allows format_number to be used in Template Program Mode, where the { } characters are not permitted.

2) Improve performance of the recent changes to support per-library user template functions. The original implementation created a copy of the template functions dict every time a template was evaluated. The new implementation creates that dict when the library is opened and passes it to the formatter.

FWIW: It seems that the changes you (Kovid) made to simplify my original code have disappeared. Am I imagining things?

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -318,14 +318,9 @@ class Connection(apsw.Connection):  # {{{
 # }}}
 
 
-def set_global_state(backend, precompiled_user_functions=None):
-    if precompiled_user_functions:
-        load_user_template_functions(backend.library_id,
-                             [],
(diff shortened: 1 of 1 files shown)


kovidgoyal:
vs2015 is periodically rebased on master -- I think the rebasing process nukes merge commits, something I forgot about. https://stackoverflow.com/questions/15915430/what-exactly-does-gits-rebase-preserve-merges-do-and-why

I've recreated the changes as their own commit.

This pull request appears to be causing lots of test failures, see for instance, https://travis-ci.org/kovidgoyal/calibre/jobs/232732775

PR #672, opened 2017-05-27
Fix #1690391: Correct KoboTouch swapping of main memory and SD card o…

davidfor:
…n OSX (https://bugs.launchpad.net/calibre/+bug/1690391)

Tested as discussed in the bug report. Plus another user using VMs that was able to create both possible cases. Also added some extra debug output under OSX just in case.


kovidgoyal:
Dont you also need to bump the max firmware version?

davidfor:
Not really. The driver is only testing the major and minor versions and ignoring the build number. I'm updating the build number mainly as a reminder of when some things changed.

PR #674, opened 2017-05-29
Optimized assets + more

majouji:
• Optimized SVG and PNG assets (with SVGO and ImageOptim)
• Reduced the complexity of certain SVG assets by removing unecessary points, groups, clipping groups, and other useless parameters
• Updated icons for a few news sources (Instapaper, La Presse, CNN…)
• Changed the title for the Cyberpresse recipe to "La Presse" (the online version of the newspaper now has the same name as the printed edition and the name Cyberpresse has been officially retired).


kovidgoyal:
The SVG assets are source files and should not be optimized (they are not included in calibre binary distributions). The PNG files are already optimized by optipng in the script that is used to generate them from the SVG files. See imgsrc/generate.py

If you can make a separate pull request for the changes in the recipes folder, then I'll be happy to merge that.

majouji:
Ah gotcha - I can do that. FWIW I think it would be worth it to separate the source files from the SVGs you use in the app bundles, but your call. Thanks!

kovidgoyal:
As I said, the svgs are not included in binary calibre releases.

PR #678, opened 2017-06-05
Adds somewhat graceful handling for Ebook split errors.

brendoncrawford:
It is difficult to reproduce exactly, but some feeds that I have used
will cause a SplitError.  This will occur no matter which preferences
are changed, such as split point, disabling split in the UI, etc.

So, this acts as a last resort to prevent the split error from halting
all processing of the ebook.

### DIFF
--- a/src/calibre/ebooks/conversion/plugins/epub_output.py
+++ b/src/calibre/ebooks/conversion/plugins/epub_output.py
@@ -191,11 +191,15 @@ class EPUBOutput(OutputFormatPlugin):
         from calibre.ebooks.oeb.transforms.rescale import RescaleImages
         RescaleImages(check_colorspaces=True)(oeb, opts)
 
-        from calibre.ebooks.oeb.transforms.split import Split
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I dont understand the point of this. Simply set the max split size to zero and you will never get split errors. IMO when splitting is enabled it is better to fail at conversion time rather than create a file that may not work on the reader device.

brendoncrawford:
I did try to set max split size to 0, and the errors still occurred.  That is the reason I submitted this.  If you think this pull request is not a good idea, I understand.  I was just trying to scratch my own itch with this.

kovidgoyal:
If it fails with max split size set to zero then it is a bug and should be fixed. If you can create a reproducer and open a bug report, I will fix it. 

I dont think simply ignoring split faliures is a good idea, as having overlarge files in the EPUB can cause the EPUB to not work on the device and it is very hard to know why a device is failing to open a file, since they typically have no error messages. One of the goals of calibre conversion is to get to a file that will work on as many devices as possible, automatically.

PR #684, opened 2017-06-11
Account for the other two ways a saved search reference can be …

cbhaley:
…written. Also change the text of the menu line.


cbhaley:
Actually, there are three more ways one can write the saved search reference. The five (!) ways are:
search:="fff", search:"=fff", search:fff, search:=fff, search:"fff"
The pushed fix handles all 5.

PR #686, opened 2017-06-11
Process conversion options

Sophist-UK:
Conversion options were introduced by 2e0ad5d1e082ef7339c4e744ed3156c208f84cf0 in 2009, but somehow the processing of this dict has been lost.

This change reintroduces the processing of conversion_options.


kovidgoyal:
Conversion options should not be serialized with recipes. They are applied in the recipe input plugin, see line 122 recipe_input.py

They are read from the instantiated recipe class that is actually used to do the downloading, not from a static cache.

Sophist-UK:
Thanks - I missed that. I will see how to fix the issue I have another way.

PR #697, opened 2017-06-16
Enhancement #1698331: add set buttons for bool columns in single …

cbhaley:
…metadata.


kovidgoyal:
@cbhaley I've merged, but given that there are three buttons does there also need to be a dropdown? You could have a tick mark on the current value button, or a simple label showing the current value.

PR #700, opened 2017-06-18
Add option in preferences / Look & Feel to display the layout buttons…

cbhaley:
… directly on the status bar. Default is False.

Although I prefer the buttons directly on the status bar, I don't feel strongly enough to argue. And I certainly won't listen to the zillions of people who will kibbitz ad nauseam.

### DIFF
--- a/src/calibre/gui2/init.py
+++ b/src/calibre/gui2/init.py
@@ -588,9 +588,7 @@ class LayoutMixin(object):  # {{{
                 ''')
             self.status_bar.addPermanentWidget(button)
         if gprefs['show_layout_buttons']:
-            from calibre.utils.icu import primary_sort_key
-            for b in sorted(self.layout_buttons, key=lambda b: primary_sort_key(b.label)):
-                print(b.label)
+            for b in self.layout_buttons:
                 b.setVisible(True)
                 self.status_bar.addPermanentWidget(b)
         else:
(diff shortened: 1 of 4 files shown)


src/calibre/gui2/init.py
```diff
@@ -587,15 +587,22 @@ def init_layout_mixin(self):
...
-        b.setToolButtonStyle(Qt.ToolButtonTextBesideIcon)
-        b.setText(_('Layout')), b.setIcon(QIcon(I('config.png')))
-        b.setMenu(LayoutMenu(self))
-        b.setToolTip(_(
-            'Show and hide various parts of the calibre main window'))
-        self.status_bar.addPermanentWidget(b)
+        if gprefs['show_layout_buttons']:
+            from calibre.utils.icu import primary_sort_key
+            for b in sorted(self.layout_buttons, key=lambda b: primary_sort_key(b.label)):
+                print(b.label)
```
kovidgoyal: Left in a print() statement. Also I thinnk originally the buttons were sorted according to the layout mode (wide or narrow) so that they correspond more closely to the position of the widgets. Use button_orrder for that.

cbhaley:
Sorry about the print statement. I managed to remove the other 2.

I originally used the button order, but I changed it to sorted the list because that is what you did for the layout button menu. Happy to remove the sort.

kovidgoyal:
I should actually use button_order int he layout menu as well, I just used sort when it was a vertical menu, since then it made more sense, but I forgot to revert that when converting it to a horizontal menu

PR #702, opened 2017-06-19
Remove junk from redesigned wall street journal

bobbysteel:
Separately, i've noticed the current recipe includes the "whats news" section which is a compendium of prior days' top posts. This doesn't feel like a fit. I'd propose that a new variable be set, defaulting to False, like include_whats_news that a user could then enable if desired. Thoughts?

### DIFF
--- a/recipes/wsj.recipe
+++ b/recipes/wsj.recipe
@@ -69,9 +69,16 @@ class WSJ(BasicNewsRecipe):
         dict(id='right-rail'),
         dict(id='narrator-nav'),
         dict(name='div', id='ad_and_popular'),
-        dict(classes('comments-count-container')),
-        classes(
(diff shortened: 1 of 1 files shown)


kiulkiel:
Yes user should have the ability to choose

kovidgoyal:
Dont replace the use of the classes() function. That matches elements that have more than one class also, unlike using attrs={'class':[]}.

ALso I am fine with including an option to turn of what's news, but it has to default to on. It is always better to have extra content than less content by default. And that section was included at user request.

bobbysteel:
fair enough. it just feels like a weird section to have in there since it's old news and the new formatting makes the article titles unreadable (it just gives have sentences vs a proper title now on the site itself). i don't need to commit this since i can just edit it out of my own local recipe, but i was hoping others might prefer to not include that section by default....

PR #703, opened 2017-06-19
Ensure that the correct keyboard shortcut is shown in the layout buttons

cbhaley:


cbhaley:
Pushed a small change to fix a potential problem with subclasses of LayoutButton

kovidgoyal:
Wouldn't it be easier to just connect to the changed() signal of the QAction and update the tooltip?

cbhaley:
Very possibly. I didn't know that the action changed signal was emitted when shortcuts changed. I will investigate.

kovidgoyal:
I pushed a patch using the changed signal, I haven't actually tested it though

PR #713, opened 2017-07-04
Quickview improvements and changes

cbhaley:
See the individual commits for more detail.


kovidgoyal:
Couple of preliminary comments:

1) It is now my policy to insist on having SVG versions of all icons (see the imgsrc directory). PNG icons are generated from them using imgsrc/generate.py

2) At least in wide layout, the layout button should probably be to the right of the cover grid, the rationale being that QV shows up in the cetral panel just like the cover grid and cover browser

kovidgoyal:
You should probably also update the quickview section in the user manual (gui.rst) since it seems to be a little out-of-date

cbhaley:
See the commit comments.

kovidgoyal:
No idea, why this code is from time immemorial :)

cbhaley:
Small change to protect against the Quickview plugin being disabled.

cbhaley:
My apologies for the last two commits. I think I am done now.

PR #715, opened 2017-07-07
Fix calibre startup crash caused by Quickview when the applied …

cbhaley:
…virtual library is empty (has no books).


cbhaley:
Push commit to make the UI deal with switching to empty libraries or VLs

PR #719, opened 2017-07-09
Quickview shortcut changes and a margin change

cbhaley:
See the commit comments.


cbhaley:
Pushed a commit to make the focus-to-booklist action work when undocked. Also changed its shortcut to make it more consistent with the focus-to-quickview default shortcut

PR #720, opened 2017-07-09
Insert _( ) to add missing translation

Serized:
I don't know if I need to do something for the translation to be added to transifex?


kovidgoyal:
No, that's all that needs to be done, the rest is handles automatically.

PR #725, opened 2017-07-15
Update price detection for Amazon get books plugins

Serized:
Update price detection for non US kindle stores


kovidgoyal:
You have to also increment the 

store_version

at the top of each plugin

Serized:
Sorry, I forgot to do it...
I also changed the detection a bit since it showed a price of 0.00 followed with the real price when the book was available for kindle unlimited (it doesn't happen for french in which I did my testing)

PR #731, opened 2017-07-25
Update appstream xml files to current appstream format

norbusan:
Changes:
* replace root node "application" with "component"
* move the "type" attribute from the "id" node to the "component"
* put screenshot urls into "image" nodes
* change installation location to .../metainfo

### DIFF
--- a/src/calibre/linux.py
+++ b/src/calibre/linux.py
@@ -841,7 +841,7 @@ class PostInstall:
                 f.close()
                 des = ('calibre-gui.desktop', 'calibre-lrfviewer.desktop',
                         'calibre-ebook-viewer.desktop', 'calibre-ebook-edit.desktop')
-                appdata = os.path.join(os.path.dirname(self.opts.staging_sharedir), 'appdata')
+                appdata = os.path.join(os.path.dirname(self.opts.staging_sharedir), 'metainfo')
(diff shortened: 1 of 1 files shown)


norbusan:
Thanks for merging.
Checking with `appstreamcli validate ...` I still get one warning about the naming of files:
`````
W - calibre-gui.appdata.xml:calibre-gui.desktop:3
    The component ID is not a reverse domain-name. Please update the ID and that of 
    the accompanying .desktop file to follow the latest version of the Desktop-Entry 
    and AppStream specifications and avoid future issues.
`````
Do you want a pull request for this, too?

Norbert

kovidgoyal:
No, changing the filename for the desktop file is fairly disruptive and not worth the cost.

norbusan:
I thought so and thus refrained from it. Thanks for the quick answer.

petterreinholdtsen:
Note, the statement "changing the filename for the desktop file is fairly disruptive" seem to indicate that it is believed that changing the file name and ID used in the appstream metadata format also forces a rename of the desktop file.  This is not true.  It is perfectly possible to have a org.reverse.domain.xml appstream file and a something.desktop desktop file by including a `<launchable type="desktop-id">something.desktop</launchable>` tag in the XML file.

PR #739, opened 2017-08-05
Suggestion: Add "calibre" to desktop entry names.

djh101:
If I happen to have multiple E-book viewers, it's nice to know which one I am opening (currently I have to go back and modify the .desktop files for Gedit, Evince, and Calibre whenever there is an update). Not sure if this has been discussed before.

### DIFF
--- a/src/calibre/linux.py
+++ b/src/calibre/linux.py
@@ -990,7 +990,7 @@ VIEWER = '''\
 [Desktop Entry]
 Version=1.0
 Type=Application
-Name=LRF Viewer
+Name=calibre LRF Viewer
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm not aware of any other software that could call itself an e-book viewer on linux with a straight face. There are various bits of software that have varying levels of buggy support for EPUB, but that's about it. And since calibre has had an e-book viewer for over a decade, I think it is up to any such software to distinguish itself from calibre.

djh101:
FBReader comes to mind. Fair enough, though (though I'll still be editing my own .desktop files; I like having the brand name in my application names, if only for consistency).

PR #759, opened 2017-10-18
Add an icon for the restart and similar books actions

Serized:
The icon looks better when the action is placed in one of the main
toolbars


Serized:
If you think another color would be better, let me know.

PR #765, opened 2017-10-31
Remove FT email nag

bobbysteel:


kovidgoyal:
Whyare there two apparently identical pull requests? Also isn't unconditionally removing all <aside> tags likely to remove some usefull content as well?

bobbysteel:
one is for US and one is for UK recipe. Per your feedback ages ago they're duplicated recipes but kept for historical reasons. 

The 'aside' tag seems to only be used for marketing purposes inside of articles, hence why adding it.

PR #767, opened 2017-11-06
Add ability to wireless device driver to send non-standard formats …

cbhaley:
…(extensions) to a device


kovidgoyal:
does this work with the code to enable/disable/re-order the  formats to send to the device?

cbhaley:
(Hmmm... I commented by email yesterday but it didn't show up here.)
A wireless device does its own reordering and sends that list back to calibre on every connect. Setting the order in calibre's preferences is disabled.

kovidgoyal:
I think you have to add your email addres to your github profile and also (maybe) setup GPG for email integration.

PR #774, opened 2017-11-22
Overwriting of formats with different languages

juampe:
As requested in https://bugs.launchpad.net/calibre/+bug/864172
I have several books with the title in English but in diferent languages,(for example Dune in english and spanish). Automerge did not work well for this case,( it overwrite the spanish version with the english version, and book data is lost).
I think this is a killer functionality.


juampe:
As I see there are test problems
```
FAIL: test_find_identical_books (calibre.db.tests.reading.ReadingTest)
Test find_identical_books
  File "C:\calibre\src\calibre\db\tests\reading.py", line 694, in test_find_identical_books
    self.assertEqual(books, find_identical_books(mi, data))

```
Certainly I modified some realated methods, but it passes my own test (insert same book, with same title in english and spanish) in my developmen environment.
Someone have some light about this?

kovidgoyal:
I am rather busy at the moment, it will be a little while before I can look at this, but thanks for the PR.

kovidgoyal:
Released.

PR #777, opened 2017-11-25
Update Economist cover for failing current recipe

bobbysteel:
Fixing mirror recipe as well

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -164,17 +164,31 @@ class Economist(BasicNewsRecipe):
                 'recipe needs to be updated.'
             )
         return ans
-
+    
+    def get_cover_url(self):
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I decided to fix it slightly differently.

bobbysteel:
Kovid - can you suggest a couple reference recipes for your preferred
styles? In this case I chose to reference the underlying link as I presume
this will be less fragile than UI layer/CSS references to an object.

Also, the preferred date for download should remain Friday AM - the cover
date is Saturday but issue updates Thurs late evening GMT.

kovidgoyal:
I tend to generally like less code, but I'm not really fussy. In my experience asset URLs are typically no more stable than UI, they are usually closely coupled.

PR #782, opened 2017-12-07
Fix cover resolution and add log line

bobbysteel:
Current method does not pull highest resolution available but instead a very low quality thumbnail

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -171,7 +171,9 @@ class Economist(BasicNewsRecipe):
         if img is not None:
             for part in img['srcset'].split():
                 if part.startswith('//'):
+                    part = part.replace('200-width','1200-width')
                     self.cover_url = 'https:' + part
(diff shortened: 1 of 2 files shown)


kovidgoyal:
The last time I checked, the higher resolution images served by the economist servers were just upscaled versions of the thumbnail cover, so it was pointless downloading them. Has that changed?

bobbysteel:
It's definitely upscaling a bit but better than the 200px ones now. the
200px doesn't even appear on the new Kindle Oasis weirdly presumably since
it's too small.

kovidgoyal:
Hmm, well to me the 640 one looks better than the 1200 one. Less pixelated and large enough to serve as a thumbnail in most places. No point wasting all that bandwidth for higher than that.

bobbysteel:
Fair enough - I'll edit now

bobbysteel:
Updated to 640px width per Kovid's comment

PR #783, opened 2017-12-08
localization: differentiate between C and not set Linux env. variables

vmiklos:
In case of LANG=hu_HU.UTF-8 and LC_MESSAGES=C, the expected behavior is
that the UI is English, not Hungarian.

The problem was that we did not differentiate between a not set
LC_MESSAGES and one that is set to C, as locale.getdefaultlocale()
returns None in both cases.


vmiklos:
Updated patch avoids a code duplication.

PR #795, opened 2017-12-25
Compatibility improvements for FreeBSD systems

cemeyer:
Don't exclude libmtp on FreeBSD systems.

Don't exclude libusb on FreeBSD systems.  Our libusb has a different include
path structure and libname than Linux, but compiles fine once those are
corrected.

Correct path to py-sip on FreeBSD.  We install it to a different directory
than build_environment.py was looking in.

### DIFF
--- a/setup/build.py
+++ b/setup/build.py
@@ -9,7 +9,7 @@ __docformat__ = 'restructuredtext en'
 import textwrap, os, shlex, subprocess, glob, shutil, re, sys, json
 from collections import namedtuple
 
-from setup import Command, islinux, isbsd, isosx, ishaiku, SRC, iswindows, __version__
+from setup import Command, islinux, isbsd, isfreebsd, isosx, ishaiku, SRC, iswindows, __version__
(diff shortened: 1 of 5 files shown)


cemeyer:
Also, merry Christmas 🎄 and happy solstice ☀️!

kovidgoyal:
Thanks, and merry christmas to you too. A question: Are your changes applicable to all BSDs or only FreeBSD (aplogies I dont knwo much about the BSDs). If the latter, perhaps use isfreebsd rather than isbsd.

cemeyer:
py-sip: Definitely for FreeBSD and likely DragonflyBSD (shared ports collection).  Not sure about NetBSD.  (Those 3 seem to be the BSDs Calibre supports?)

libusb: Not sure.

libmtp: I think all the BSDs have the same ported libmtp.

cemeyer:
I've updated the patch to only affect FreeBSD.

cemeyer:
Thanks!

PR #797, opened 2017-12-29
Fix few dbus signatures

muktupavels:


oblongau:
Confirming that Calibre 3.15 in Linux Mint 18.3 64 bit MATE now works correctly. No error messages, and a Calibre icon now appears in the tray. Thanks to all who helped.

PR #798, opened 2017-12-29
Fix typo in a comment.

madpilot78:


kovidgoyal:
iff is short for if and only if

madpilot78:
> iff is short for if and only if

I did not know.

Thanks for explaining!

PR #800, opened 2017-12-30
Just use libusb to scan USB devices on FreeBSD

cemeyer:
The HAL scanner adds a needless dependency on a runtime service and the
libusb scanner is perfectly fine.

This fixes detection of MTP devices on FreeBSD.

The logical patch was written by Guido Falsi; additionally, I removed
the legacy FreeBSDScanner class.


cemeyer:
Tested with: FreeBSD CURRENT, Nook Glowlight+.

PR #802, opened 2018-01-08
Extended MyBlockingBusy with a Progress Report in Form of <Processed …

remggo:
…Books>/<Total Books>

Only if `self.progress_update.emit()` is called the String in the Dialog is changed from the current one to: `"Processed %d/%d books, please wait..."`. I updated the cases I knew how to update to report Progress Updates. In the Fields I was not sure, I did not change the current behaviour.

Feedback is welcome!

### DIFF
--- a/src/calibre/gui2/dialogs/metadata_bulk.py
+++ b/src/calibre/gui2/dialogs/metadata_bulk.py
@@ -57,14 +57,16 @@ class Caser(LineEditECM):
 class MyBlockingBusy(QDialog):  # {{{
 
     all_done = pyqtSignal()
+    progress_update = pyqtSignal()
 
     def __init__(self, args, ids, db, refresh_books, cc_widgets, s_r_func, do_sr, sr_calls, parent=None, window_title=_('Working')):
         QDialog.__init__(self, parent)
 
         self._layout =  l = QVBoxLayout()
         self.setLayout(l)
-
-        self.msg = QLabel(_('Processing %d books, please wait...') % len(ids))
+        self.len_ids = len(ids)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Wouldn't it be rather disconcerting to have the progress keep on resetting back to zero with every type of change? Instead calculate a number of tasks i.e. number of ids * number of things to do and update that smoothly. Also, dont change function calls that process multiple books into individual function calls, as that degrades performance (there are various internal optimizations the db can do when processing the same operation on  multiple books).

remggo:
Reading your comments, I thought it would be best to maybe use a ProgressBar instead. That way, there would be no edits to the UI-strings necessary. The function calls that process multiple books are e.g. `cache.set_field('author_sort', {bid:args.aus for bid in self.ids})` these? There the progress bar could just increase with `self.ids` after finishing. That would not look as smooth especially not when only the one of such fields is set, but I think it would be the best option.

I will start implementing the changes tonight.

kovidgoyal:
Yes, set_field is an example of such a function call. The reason I did not use a progress bar is because the various operations dont take equal amounts of time, so progress is not linear. Thinking about it, probably the best design is to have two bars, one for the current task, and one for overall progress. If there is only one task, hide the overall bar. If the task being executed is a single function call, then set the progress bar to "indeterminate mode" (IIRC by setting min and max values both to zero).

remggo:
Your Idea with the two progress bars sounds very good and it keeps the user informed about the different steps. Will use that, thanks!

kovidgoyal:
I have merged this. I think the bars should be made a little longer

PR #803, opened 2018-01-10
Fix Calibre Companion sync on Linux

jceb:
- listen(0) doesn't work on Linux, select.select() and socket.accept() return no connection

### DIFF
--- a/src/calibre/devices/smart_device_app/driver.py
+++ b/src/calibre/devices/smart_device_app/driver.py
@@ -1927,7 +1927,7 @@ class SMART_DEVICE_APP(DeviceConfig, DevicePlugin):
                     return message
 
             try:
-                self.listen_socket.listen(0)
+                self.listen_socket.listen(1)
             except:
                 message = 'listen on port %d failed' % port
                 self._debug(message)


kovidgoyal:
@cbhaley comments?

cbhaley:
I have years of experience that listen(0) works fine on linux. What evidence is there that this change is needed?

kovidgoyal:
As far as I know, the backlog argument is advisory, the kernel typically ignores it an uses a power of two backlog anyway. As per the listen man page:

A backlog argument of 0 may allow the socket to accept connections, in which case the length of the listen queue may be set to an implementation-defined minimum value.

cbhaley:
I have no objection to the change except possibly triggering the law of unintended consequences.

What I really want to know is whether some Linux kernel *requires* backlog to be greater than 0. TBH that would astonish me, but I have been astonished before. I know for a fact that Linux kernels exist that *do not* require backlog to be greater than 0.

jceb:
I search the rest of the codebase and found only listen(>0) being used.  In addition, the following didn't work for me with listen(0):

```python
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.bind(("", 9091))
s.listen(0)
```
Followed by:

```bash
telnet localhost 9091
```

cbhaley:
Hmm... That snippet with listen(0) works fine for me with kernel 
Linux 4.14.12-x86_64-linode92 #1 SMP Fri Jan 5 15:34:44 UTC 2018 x86_64 GNU/Linux

@Kovid: the above notwithstanding, I don't see how the change will actually break anything except possibly adding delays to multiple simultaneous connection error detection, I have no problem with you merging the change.

PR #805, opened 2018-01-12
Add a tweak to control series number display

Serized:
I'm not sure about the english description, so please check if it's understandable.


Serized:
Actually I wanted to change only what's displayed in the global view. This way the data that is in the library isn't changed (if a book is saved as 1.11 before changing the tweak)
I will look around to change the display in the other places.

Serized:
It already works for custom column and I don't think I can call the tweak from the content server. And as for the metadata writer, I'm not sure it should be changed. This works for me and I'm not good enough at programming to make big changes ;)
So if you don't want to add the change just reject the pull request

kovidgoyal:
Yeah, this is rather too big a change to merge without lots of work to make sure it does not break things in many places.

PR #828, opened 2018-04-01
[WIP] Add ability to open books for reading from the console by id.

mmmm1998:
Well, this is needed feature for me (too large library (~3000), GUI become too slow (library on old laptop))
Now, it's just a addition command(`path`) in `calibredb`. It's get list of library ids and return paths to one of available formats for existing ids. This first version some ugly, but I will be improve it.
So, @kovidgoyal, will you agree with adding this feature(new command to open book by id)?
Where place of `open` command, in `calibredb`? Or in another place?


kovidgoyal:
You can do this already with

calibredb list --search id:whatever --fields formats --for-machine

PR #830, opened 2018-04-28
New option to override KoboTouch driver book resend behaviour and bump the supported firmware version

davidfor:
Adding an option for overriding the Kobo firmware behaviour when a book is replaced. The default behaviour is for the book to be removed from the internal database and then treated as a new book. This means the reading status is lost. But, the KoboTouch driver has always overriden this behaviour and to allow the reading status to be kept. The option is to override the firmware behaviour and defaults to on to keep the existing driver behaviour.

Kobo has released new firmware. The only change is to bump the firmware and database versions.

### DIFF
--- a/src/calibre/devices/kobo/driver.py
+++ b/src/calibre/devices/kobo/driver.py
@@ -69,7 +69,7 @@ class KOBO(USBMS):
 
     dbversion = 0
     fwversion = (0,0,0)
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I made some minor changes to the text for the new option and also some code changes in kobo/driver.py to comply with pep8 formatting. Have a look.

PR #831, opened 2018-04-29
Search parsing: fix using a parenthesized expression following an …

cbhaley:
…implied "and", eg "A (B or C)" now is parsed correctly as "A and (B or C)"


kovidgoyal:
Could you add a test for this in search_query_parser_test.py

cbhaley:
Done. I ran the test without the changes and it failed as expected (with an exception, not an assertion). I also ran it with incorrect info in the set of "books" and it failed as expected.

PR #839, opened 2018-05-22
Fix compilation on qt 5.11

eli-schwartz:
Needed for Arch Linux which now has qt 5.11 in [testing].

I've tested building this against 5.10 and 5.11, seems to work fine with both. :)

It's basically just merging some upstream qfontconfigdatabase.cpp changes in, the important thing is https://code.qt.io/cgit/qt/qtbase.git/commit/src/platformsupport/fontdatabases/fontconfig/qfontconfigdatabase.cpp?id=8bfabb34dec8a437a08b5a6e0ecac4a9dd3ae18c

I've  thrown in some free updates to the INSTALL file, which is something that irked me since a while, which I rewrote then forgot about until just now (since I'm filing a PR anyway).


kovidgoyal:
calibre needs to be buildable all the way back to Qt 5.6 with pre c++11 compilers. So I dont think the replacements of Q_NULLPTR and _DECL_OVERRIDE with C++11 versions will work. I've merged in the changes to specialLanguages. Let me knwo if anythong else needs to be done.

And feel free to send a separate PR with your changes to INSTALL

eli-schwartz:
Oh, good point I guess.

PR #842, opened 2018-05-29
Add support for Kobo Clara HD

davidfor:
Kobo has announced a new device, the Clara HD. It is the same size and resolution as the Glo HD, so just needed to add the device id.


kovidgoyal:
I have merged, but there seems to be a mismatch metween max_supported_fw-version and clara_firmware_version (the latter is larger than the former)

PR #844, opened 2018-06-01
Workaround (maybe) for bug #1766762: it seems that there is something…

cbhaley:
… strange with wordwrapped labels with some fonts. One part of QT thinks it is showing a single line and sizes the line vertically accordingly. Another part thinks there isn't enough space and splits the label. The result is two lines in a single line space. It doesn't happen with every font, nor with every "long" label.

This change works around the problem by setting the maximum display width and telling QT to respect that width. I wasn't able to repeat it after making this change.

While there I implemented a minimum label length so that there is a better chance that the field edit boxes line up..


kovidgoyal:
It would be nice to have a comment explaining the need for it for anybody needing to modify it in the future.

PR #854, opened 2018-07-03
sorted library items by name on opds

DrDoa:
I found that libraries are not sorted by the name, so I fix the feature for display library items by the name when using opds.

### DIFF
--- a/src/calibre/srv/opds.py
+++ b/src/calibre/srv/opds.py
@@ -301,7 +301,8 @@ class TopLevel(Feed):  # {{{
             categories]
         for x in subcatalogs:
             self.root.append(x)
-        for library_id, library_name in request_context.library_map.iteritems():
+        sorted_library_items_by_name = sorted(request_context.library_map.items(), key=lambda item:item[1])
+        for library_id, library_name in sorted_library_items_by_name:
             id_ = 'calibre-library:' + library_id
             self.root.append(E.entry(
                 TITLE(_('Library:') + ' ' + library_name),


DrDoa:
Hi, Kovid Goyal,  I have a difficult problem to solve, but I have not found a valid solution.
I would like to sort the book names on the main view by their names.
As their names are Chinese, some Chinese numbers are included.
the matches are:
一 == 1
二 == 2
三 == 3
...
but the QT default sort is set as (PinYin)
一 its 'PinYin'  is 'yi'
二 its 'PinYin'  is 'er'
三 its 'PinYin'  is 'san'
...
So the numbers are sort using the first letter of the number's 'PinYin' when ascending :
二 (2)
三 (3)
一 (1)
...
Actually, it is not the right natural number order, I would like to use a custom order function to tell the program to know how to order them, but I failed to find an interface.
Do you have any solutions on it?

kovidgoyal:
Preferences->tweaks->recognize numbers inside text when sorting

DrDoa:
I have tested this tweak just now.  Actually,  it is not for my problem, Although it can sort by Arabic numbers. As my problem is to sort text by Chinese numbers, the app does not know how the Chinese numbers work.

kovidgoyal:
It is based on numeric sorting from the ICU library, I'd be very surprised if it does not know about chinese numbers, but if so, you should ask that the library add support for them. As for your problem, why are you embedding numbers inside a text column? Use a separate numeric column for that. Then you can sort it easily.

DrDoa:
Actually, I have tried to convert Chinese numbers to Arabic numbers with a sequence of regular replace, and I have successfully done it. when I am adding new books, their name would be converted. But I think this is a trick, sometimes, looks not perfectly comfortable to Chinese text. If possible, by making the app knowing Chinese numbers, the text would be more beautiful.
I found the sort action trigged QTableView's  sortByColumn() method to sort columns.
At same time,  I know that in C++ Qt could we can override the QTableView's  sortByColumn() method to custom the sorting,  but I have not found how to hook the PyQt's method. Maybe there are some other method for custom sorting in PyQt.

kovidgoyal:
This has nothing to do with Qt. Sorting in calibre does not use Qt functionality. As I said before sorting is done via ICU.

DrDoa:
Thanks, can I hook it?

kovidgoyal:
Since calibre is open source you can do anything you want. The relevant function is sort_key() in icu.py

DrDoa:
Thanks, I am reading the relevant codes now.

DrDoa:
Hi, Kovid Goyal. I need some help. I need to add python package to Calibre, but I cannot find how to make it? Could you give some help?

kovidgoyal:
Simply add whatever path the package is installed into to sys.path before importing it.

DrDoa:
And I am adding djvu meta-data reading plugin on my own machine. I use 'djvused' from 'djvuLibre' for reading meta-data and 'ddjvu' for extract top page of djvu file as cover data. now working on my machine.
Does it possible to add this feature to Calibre master branch? Is it possible to bring 'djvused' and 'ddjvu' with Calibre?

kovidgoyal:
I dont want to add a whole extra dependency just to rad metadata. calibre already has a djvu input plugin, it should be easy to modify that to also read metadata.

DrDoa:
emmm, can it read meta-data from djvu? Sometimes In my case, I modify books by python command in batch and add books in batch. If already exist this feature, can we enable this feature?

kovidgoyal:
Once again: calibre already has a djvu input plugin, it **should be easy to modify** that to also read metadata.

DrDoa:
I am curious about why no one have added djvu metadata reader.

DrDoa:
I have read djvu_input, and it is just calling calibre.*.ui.get_file_type_metadata to get metadata.
`
        # Set metadata from file.
        from calibre.customize.ui import get_file_type_metadata
        from calibre.ebooks.oeb.transforms.metadata import meta_info_to_oeb_metadata
        mi = get_file_type_metadata(stream, file_ext)
        meta_info_to_oeb_metadata(mi, oeb.metadata, log)
`
as there is no djvu metadata reader.

in calibre.ebooks.djvu I find you have import ZPCodec and BZZ from djvuLibre.V3.5, rewritten in C. And package in Python.  
I do not think it is easy to modify and read metadata for a djvu file without import more parts from djvuLibre.

kovidgoyal:
I haven't imported them, the person that contributed the djvu code for it has imported them. And those are simply for decompression, the actual code to parse djvu file structues is all in python, IIRC.

PR #860, opened 2018-07-23
Added support for LazyLibrarian opf files

knobunc:
This patch adds support to parse the OPF files generated by LazyLibrarian so that they can better interoperate.


kovidgoyal:
application_id has a specific meaning in calibre and is meant for internal use. What are you actually trying to achieve?

kovidgoyal:
no follwup

PR #864, opened 2018-08-13
Deleted bloomberg recipe, change in service

dkfurrow:
Hi @kovidgoyal , 
Deleted bloomberg columns recipe that I contributed a month ago.  Pages won't load without a javascript-enabled browser (returns a "violation of service" message).  Bottom line, seems that Selenium works, but mechanize will not.  Unless you know a better way around this, seems like the best thing to do is just delete the recipe.
Thanks,
Dale


kovidgoyal:
Typically, with these kinds of things, the content is loaded in an invisible part of the page (for search engine optimization) so you can extract it from there. Or alternately it is loaded via AJAX calls which you can simulate in mechanize. But if that is too much work, let me know and I will merge your PR.

kovidgoyal:
Use the network inspector tool in the chrome devtools to see what request is being made by the JS to laod the content, then just duplicate that request in mechanize, no js knowledge needed at all.

kovidgoyal:
I'm not familiar with requests -- never felt the need for it. But it likely has some way to dump the actual HTTP request it makes, use that and compare it with the one mechanize makes. You can get mechanize to dump http requests with

br.set_debug_http(True)

0xInfection:
Mechanize is no longer supported and should be deprecated.

kovidgoyal:
says who?

0xInfection:
Oh I just noticed that you're the present maintainer. 😅 Nice.

dkfurrow:
Hi @kovidgoyal; I've gotten busy on some other things, and Bloomberg has changed their website such that the premise of this recipe (the contributors page) no longer exists.  So I'm no longer maintaining this recipe and it's not working.
Thanks,
Dale

PR #870, opened 2018-09-03
WIP: Port calibre to python 3

flaviut:
What works:

- Starting the calibre GUI
- Calibre C extensions
- Adding a book

What doesn't work:

- Recipies
- Everything else
- Python 2 compatibility

Anyway, this pull request is just here to get some early feedback. There's a lot of work that is left to do, like manually testing as many code paths as possible, as well as porting recipies to use MechanicalSoup or something like that instead of mechanize.

My goal is to have this code work the same on python2 and on python3, with the goal of eventually dropping python2 support late 2019 or early 2020.

To bootstrap this version, you'll need to install the following pip packages:

```
apsw==3.9.2.post1
dukpy==0.3
Elements==0.2.0
feedparser==5.2.1
html2text==2018.1.9
html5-parser==0.4.5
netifaces==0.10.7
```

and delete all recipes except `ajc.recipie`


kovidgoyal:
Some quick notes glancing through the commits:

1) The icu module is broken (you cannot just remove the PY >= 3.3 checks) while it will compile with that, the actual logic will not work. You will need to implement proper fixes for it since the existing code is designed to work with either UCS2 or UCS4 python strings.

2) Replacing ur'' with r'' is fine for python3 but it breaks in python2 since the strings are no longer unicode,

3) Rather than eliminating xrange (since that is a performance regression in python 2) create a new xrange builtin for python 3

4) Similarly replacing cStringIO with StringIO is a huge performance regression for python 2 Instead you would need to analyze the code where it is used and see if it can be replaced by BytesIO or StringIO

There are probably lots more issues. So if you are serious about proceeding with this, I suggest the following approach. 

Start by identifying relatively safe changes to the code base that make it compatible with both py2 and py3. Send PRs for those and I will review and merge them one by one. Get to the point where you can start running the test suite. Then start making changes to get the tests passing on py2 and py3.

kovidgoyal:
Oh and another huge pain to port to python 3 will be the calibre server, since it relies on pretty deep integration with the python socket and ssl modules. In particular all the error handlers will need to be re-written since IIRC in python 3 the socket module has very different error semantics.

flaviut:
Thank you for your detailed comments:

> 1. the conversion subsystem (particularly some of the input/output plugins
>    whose code is very old -- for example RTF input and ODT input). And also
>    plugins like MOBI input/output which make extensive use of operations on
>    bytes that py 3 does not support.

The only difference between python2 & python3 that I've found wrt bytes is that
str[x] returns str in python2, and bytes[x] returns int in python3.

I took a quick look at the mobi support, and didn't see anything that would be
a porting problem.

> 2. The recipe system -- this requires either porting mechanize to py 3 or
>    writing an mostly API equivalent replacement, I doubt MechanicalSoup will
>    cut it as calibre allows custom recipes, so you cant just port the builtin
>    ones and call it a day.

The way I'd see this being done would be a year-long deprecation
period--convert all internal recipies to MechaincalSoup, and the plugin authors
can update their recipes as their target websites change code & require
modifications.

> 3. The device subsystems: these contain lots of OS and device specific code
>    that is impossible to test thoroughly, and when porting, for exmaple, the
>    wpd C extension you will need to deal witht he fact that windows APIs
>    expect UTF-16 which python 2 string are antively and python 3 strings are
>    not.
> 4. Interacting with the OS generally -- subprocesses, OS specific APIs sunch
>    as winutil/usbobserver etc.

I understand; thank you for the warning. I'm not looking forward to dealing
with this part.

> My current plan is to simply support python 2 myself, which I already do to
> some extent, see https://github.com/kovidgoyal/cpython that is much less work
> and less potential for regressions. However, that is not to say I am opposed
> to trying to port calibre to py3 -- it's just a lot of work for very low
> return.

I disagree :)

I believe that being able to share ideas and code with a wider community has
important advantages, even if they are hard to quantify.

> I'd estimate it would take a couple of man-years and that is before all the
> regressions.

I think (hope, at least) that this is one of the few instances where the time
taken to complete a software project is overestimated :)

> 1. The icu module is broken (you cannot just remove the PY >= 3.3 checks) while
>    it will compile with that, the actual logic will not work. You will need to
>    implement proper fixes for it since the existing code is designed to work
>    with either UCS2 or UCS4 python strings.
> 2. Replacing ur'' with r'' is fine for python3 but it breaks in python2 since
>    the strings are no longer unicode,

Good points; thank you.

> 3. Rather than eliminating xrange (since that is a performance regression in
>    python 2) create a new xrange builtin for python 3

[six.moves.range is the same as xrange in py2 and range in
py3](https://pythonhosted.org/six/#module-six.moves)

> 4. Similarly replacing cStringIO with StringIO is a huge performance regression
>    for python 2 Instead you would need to analyze the code where it is used and
>    see if it can be replaced by BytesIO or StringIO

[six.moves.StringIO is the same as cStringIO in
py2.](https://pythonhosted.org/six/#module-six.moves)

> There are probably lots more issues. So if you are serious about proceeding
> with this, I suggest the following approach.
> 
> Start by identifying relatively safe changes to the code base that make it
> compatible with both py2 and py3. Send PRs for those and I will review and
> merge them one by one. Get to the point where you can start running the test
> suite. Then start making changes to get the tests passing on py2 and py3.

Got it.

> Oh and another huge pain to port to python 3 will be the calibre server, since
> it relies on pretty deep integration with the python socket and ssl modules. In
> particular all the error handlers will need to be re-written since IIRC in
> python 3 the socket module has very different error semantics, that are also
> platform dependent.

Just curious, why choose to create a custom HTTP framework? There's plenty of
other choices for web frameworks in python.

flaviut:
Ok, one more question:

The logical starting point here seems to be porting `setup.py bootstrap` to work with python 3. However, setup.py depends on various modules in calibre, so it's not possible to do anything without first porting those portions of calibre.

Do you have any suggestions on how to handle this situation, or maybe of a module that is more self-contained?

kovidgoyal:
I suggest you let bootstrap run under python 2 for the time being. Use a checkout that is already bootstrapped with python 3 (most of the things bootstrap does should be python version independent anyway). The first thing to get working would be python3 setup.py build to build at least a few C extensions that are widely used such as ICU, speedups, etc. 

You would need to first make calibre python code importable under python 3. This means porting the simple things like print_function and ur'' etc. Make separate PRs for those and I will review and merge. 
After that I will see if I can get python3 setup.py build to run, once I have that done, you can work on porting individual C extensions, make PRs for each and I will review and merge. After that we can work on getting python3 setup.py test to run. Once that is done porting work for individual failing tests will need to be done step by step.

calibre used to use an HTTP framework for its server (cherrypy). It was removed and replaced with custom code in calibre 3 because it had various bugs and inadequacies and I wanted an async server -- none of the existing libraries for python 2 did that well. 

I would be very surprised if this porting effort took less than a man-year of effort. But, the only way to be sure is to try...

Also, just a note on general principles. I am not a big fan of six. It has burned me in the past with incompatible changes and it is very heavyweight. The calibre codebase does not doa  lot of fancy stuff, so what we need for compatibility is only a small subset of six. I would prefer if we create a dedicated module for it with just what we need, much lighter weight. For an example of such a module, see https://github.com/python-mechanize/mechanize/issues/9 where I wrote one for mechanize

kovidgoyal:
Also, just to note, I am fine with requiring python >= 3.7 for the port, since by the time the port is done, 3.7 should be fairly old.

kovidgoyal:
Link to mechanize's polyglot module: https://github.com/python-mechanize/mechanize/blob/master/mechanize/polyglot.py

ofek:
Has there been progress on this?

flaviut:
Only what you can see here: https://github.com/kovidgoyal/calibre/pulls?q=is%3Apr+author%3Aflaviut

I've been busy with real-life stuff, but I should be able to get back into it this month.

NotAFile:
I'd highly recommend not trying to do this all in one go. Trying to keep up with new changes is a relatively futile exercise. In my experience of porting a few projects, it is much more effective to require new code to be at least syntax-compatible with python3 and to then slowly grow out the number of files that are python2+python3 compatible, using tests to ensure anything that has been marked as working does not regress.

eli-schwartz:
Current status on porting C extensions:

- [x] _patiencediff_c #907
- [ ] bzzdec
- [ ] cPalmdoc
- [x] certgen #887
- [ ] chm_extra
- [ ] chmlib
- [ ] freetype
- [ ] headless
- [x] html #891
- [x] hunspell #872
- [x] icu #908
- [ ] imageops
- [ ] libmtp
- [ ] libusb
- [ ] libusb (freebsd edition)
- [ ] lzma_binding
- [x] lzx #914
- [x] matcher #889
- [x] monotonic #874
- [ ] msdes
- [ ] pictureflow
- [ ] podofo
- [ ] progress_indicator
- [ ] qt_hack
- [x] speedup #879
- [ ] sqlite_custom
- [x] tokenizer #905
- [x] unicode_names #875
- [ ] usbobserver
- [ ] winfonts
- [ ] winutil
- [ ] wpd
- [x] zlib2 #881

eli-schwartz:
@kovidgoyal looks like pychm (which is where chmlib and chmlib_extra come from) is technically maintained since 2014 at https://github.com/dottedmag/pychm/ (PyPI links it as the new homepage), although that does seem fairly inactive too...

There is a port to python3 which has not been merged, at https://github.com/dottedmag/pychm/pull/5

What is the exact provenance of the version calibre includes? Are there any calibre-specific modifications to the original (unmaintained) code? I'm hoping that, like the regex module, this could be reduced to a thirdparty dependency, possibly by working to upstream any changes.

As far as I can tell, post inclusion, e69fc6b8b3c665c95611bf951376c98959526b0e is fixed upstream, and 1f2daebce60aa01438ddbe35b02746552a399114 could just as well be done in calibre and not in pychm.

eli-schwartz:
Well, I've cp'ed things for now and it builds. I've attempted to preserve things, like commit 1f2daebce60aa01438ddbe35b02746552a399114, which seem to be modifications made by you.

kovidgoyal:
> EDIT: Not like this code is exactly high-maintenance though. Except for porting to python3...

Yeah but given that there exists an open PR and its been ignored for years...

eli-schwartz:
Just since March is not quite years :D but yeah.

eli-schwartz:
#932 ports bzzdec, #931 added more compat layers for python code.

eli-schwartz:
#934 ports two more plugins

eli-schwartz:
#935 ports sqlite_custom

eli-schwartz:
#936 ports freetype
#937 ports msdes

eli-schwartz:
#940 should clean up a bunch more preliminaries for misc modules, and port libmtp/podofo.

eli-schwartz:
#941 is a massive cleanup of unicode/unichr types for python3.

Some bits missing for now:
- recipes, stores, metadata plugins need a migration period because they get online updates
- beautifulsoup, src/odf/ support python3 upstream and it would be preferable to migrate to a tested python3 port if possible.

ofek:
Keep up the excellent work, thank you!

eli-schwartz:
#942 should be the last of the C extension bits.

eli-schwartz:
Made some more progress on porting code that does not import on python3: #945 for c?StringIO, #944 for cPickle (but Kovid is probably just going to drop the pickle code in favor of things that, well, aren't pickle. :p)

Also some fixes for xrange on master...

Anyway, given it is now possible to "build" calibre, it is time to start attempting to import it, see how far it runs before crashing, and incrementally fix stuff. Let's see how soon we can start up a graphical interface. :)

kovidgoyal:
I suggest before attempting to run calibre, get all the tests to pass under py3. Sadly calibre has a pretty meager test suite, but given how large it is, even the meagre test suite does cover a lot fo code.

eli-schwartz:
Status update:

- #946, #947, #948 with general cleanups for syntax differences
- bug found in PyQt5: https://www.riverbankcomputing.com/pipermail/pyqt/2019-March/041490.html
- misc dead code removed from calibre
- calibre now uses the upstream html2text module, and has ported over to stock beautifulsoup4 (which supports python3)

eli-schwartz:
I don't understand this error at all:

```
*
* Running resources
*

  File "setup.py", line 121, in <module>
    sys.exit(main())
  File "setup.py", line 106, in main
    command.run_all(opts)
  File "/home/eschwartz/git/calibre/py3/setup/__init__.py", line 242, in run_all
    self.run_cmd(self, opts)
  File "/home/eschwartz/git/calibre/py3/setup/__init__.py", line 238, in run_cmd
    cmd.run(opts)
  File "/home/eschwartz/git/calibre/py3/setup/resources.py", line 307, in run
    from calibre.web.feeds.recipes.collection import \
  File "/home/eschwartz/git/calibre/py3/src/calibre/web/feeds/recipes/__init__.py", line 8, in <module>
    from calibre.web.feeds.news import (BasicNewsRecipe, CustomIndexRecipe,
  File "/home/eschwartz/git/calibre/py3/src/calibre/web/feeds/news.py", line 23, in <module>
    from calibre.web.fetch.simple import option_parser as web2disk_option_parser, RecursiveFetcher, AbortArticle
  File "/home/eschwartz/git/calibre/py3/src/calibre/web/fetch/simple.py", line 28, in <module>
    from calibre.utils.img import image_from_data, image_to_data
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/img.py", line 29, in <module>
    imageops, imageops_err = plugins['imageops']
  File "/home/eschwartz/git/calibre/py3/src/calibre/constants.py", line 226, in __getitem__
    self.load_plugin(name)
  File "/home/eschwartz/git/calibre/py3/src/calibre/constants.py", line 211, in load_plugin
    self._plugins[name] = (p, err)
UnboundLocalError: local variable 'err' referenced before assignment
```

But the variable is set in both the try and except clauses right before this.

eli-schwartz:
Also:

```
>>> from calibre.plugins import icu
>>> from calibre.plugins import imageops
  File "<stdin>", line 1, in <module>
ImportError: dynamic module does not define module export function (PyInit_imageops)
```

But imageops is one of the sip-generated modules, so this is supposed to automatically work, and the symbol is in there, so why is it not working? :/

kovidgoyal:
Should be fine now.

eli-schwartz:
Thanks, seems to work -- the interaction with the version script escaped my notice.

eli-schwartz:
#950 has my first new test! All it is doing is checking that all code files can be successfully imported, though. Still, it's helped find a bunch more low-hanging issues, and will hopefully be better at catching stupid typos too.

eli-schwartz:
The testsuite is beginning to be feasible to run.

```
$ CALIBRE_PY3_PORT=1 python3 setup.py test --test-module=build
[...]
======================================================================
ERROR: test_feedparser (calibre.test_build.BuildTest)
  File "/home/eschwartz/git/calibre/py3/src/calibre/test_build.py", line 306, in test_feedparser
    from calibre.web.feeds.feedparser import parse
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 668, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 640, in _load_backward_compatible
KeyError: 'calibre.web.feeds.feedparser'

======================================================================
ERROR: test_imports (calibre.test_build.BuildTest)
  File "/home/eschwartz/git/calibre/py3/src/calibre/test_build.py", line 95, in test_imports
    importlib.import_module(name)
  File "/usr/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/eschwartz/git/calibre/py3/src/calibre/db/legacy.py", line 766, in <module>
    setattr(LibraryDatabase, prop, MT(getter(prop)))
  File "/home/eschwartz/git/calibre/py3/src/calibre/db/legacy.py", line 753, in <lambda>
    MT = lambda func: types.MethodType(func, None, LibraryDatabase)
TypeError: method expected 2 arguments, got 3

======================================================================
ERROR: test_lzma (calibre.test_build.BuildTest)
  File "/home/eschwartz/git/calibre/py3/src/calibre/test_build.py", line 66, in test_lzma
    test_lzma2()
  File "/home/eschwartz/git/calibre/py3/src/lzma/xz.py", line 429, in test_lzma2
    raw = P('template-functions.json', allow_user_override=False, data=True)
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/resources.py", line 76, in get_path
    with open(fpath, 'rb') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/eschwartz/git/calibre/py3/resources/template-functions.json'

======================================================================
ERROR: test_markdown (calibre.test_build.BuildTest)
  File "/home/eschwartz/git/calibre/py3/src/calibre/test_build.py", line 301, in test_markdown
    create_markdown_object(sorted(MD_EXTENSIONS))
  File "/home/eschwartz/git/calibre/py3/src/calibre/ebooks/txt/processor.py", line 112, in create_markdown_object
    from calibre.ebooks.markdown import Markdown
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 668, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 640, in _load_backward_compatible
KeyError: 'calibre.ebooks.markdown'

======================================================================
ERROR: test_change_case (calibre.utils.icu_test.TestICU)
Test the various ways of changing the case
  File "<string>", line 5, in upper
TypeError: Not a unicode string

During handling of the above exception, another exception occurred:

  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/icu_test.py", line 84, in test_change_case
    self.ae('A', icu.upper(b'a'))
  File "<string>", line 15, in upper
UnicodeDecodeError: 'utf-16-le' codec can't decode byte 0x41 in position 0: truncated data

======================================================================
ERROR: test_collation_order (calibre.utils.icu_test.TestICU)
Testing collation ordering
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/icu_test.py", line 130, in test_collation_order
    self.ae(dict(icu.partition_by_first_letter(['A1', '', 'a1', '\U0001f431', '\U0001f431x'])),
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/icu.py", line 295, in partition_by_first_letter
    c = icu_upper(key(item) or ' ')
  File "<string>", line 5, in upper
UnicodeDecodeError: 'utf-16-le' codec can't decode byte 0x20 in position 0: truncated data

======================================================================
ERROR: test_find (calibre.utils.icu_test.TestICU)
Test searching for substrings
  File "<string>", line 5, in primary_find
AttributeError: 'NoneType' object has no attribute 'find'

During handling of the above exception, another exception occurred:

  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/icu_test.py", line 102, in test_find
    self.ae((0, 4), icu.primary_find('pena', 'peña'))
  File "<string>", line 7, in primary_find
ValueError: U_INVALID_CHAR_FOUND

======================================================================
ERROR: test_sorting (calibre.utils.icu_test.TestICU)
Test the various sorting APIs
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/icu_test.py", line 51, in test_sorting
    self.ae(german_good, sorted(german, key=dsk))
  File "<string>", line 5, in dsk
ValueError: U_INVALID_CHAR_FOUND

======================================================================
ERROR: test_color_parsing (tinycss.tests.color3.TestColor3)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/color3.py", line 158, in test_color_parsing
    result = parse_color_string(css_source)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/color3.py", line 46, in parse_color_string
    tokens = list(tokenize_grouped(css_string.strip()))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_at_media (tinycss.tests.css21.TestCSS21)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/css21.py", line 323, in test_at_media
    stylesheet = CSS21Parser().parse_stylesheet(css_source)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/css21.py", line 407, in parse_stylesheet
    tokens = tokenize_grouped(css_unicode)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_at_page (tinycss.tests.css21.TestCSS21)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/css21.py", line 285, in test_at_page
    stylesheet = CSS21Parser().parse_stylesheet(css)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/css21.py", line 407, in parse_stylesheet
    tokens = tokenize_grouped(css_unicode)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_core_parser (tinycss.tests.css21.TestCSS21)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/css21.py", line 146, in test_core_parser
    stylesheet = CoreParser().parse_stylesheet(css_source)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/css21.py", line 407, in parse_stylesheet
    tokens = tokenize_grouped(css_unicode)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_important (tinycss.tests.css21.TestCSS21)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/css21.py", line 224, in test_important
    declarations, errors = CSS21Parser().parse_style_attr(css_source)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/css21.py", line 425, in parse_style_attr
    return self.parse_declaration_list(tokenize_grouped(css_source))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_parse_style_attr (tinycss.tests.css21.TestCSS21)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/css21.py", line 193, in test_parse_style_attr
    declarations, errors = CSS21Parser().parse_style_attr(css_source)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/css21.py", line 425, in parse_style_attr
    return self.parse_declaration_list(tokenize_grouped(css_source))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_media_queries (tinycss.tests.media3.TestFonts3)
Test parsing of media queries from the CSS 3 media module
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/media3.py", line 61, in test_media_queries
    stylesheet = CSSMedia3Parser().parse_stylesheet(css)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/css21.py", line 407, in parse_stylesheet
    tokens = tokenize_grouped(css_unicode)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_content (tinycss.tests.page3.TestPage3)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/page3.py", line 78, in test_content
    stylesheet = CSSPage3Parser().parse_stylesheet(css)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/css21.py", line 407, in parse_stylesheet
    tokens = tokenize_grouped(css_unicode)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_token_api (tinycss.tests.tokenizing.TestTokenizer)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 253, in test_token_api
    self.run_test(token_api)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 250, in run_test
    func(self, tokenize)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 21, in token_api
    tokens = list(regroup(tokenize(css_source)))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_token_grouping (tinycss.tests.tokenizing.TestTokenizer)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 262, in test_token_grouping
    self.run_test(token_grouping)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 250, in run_test
    func(self, tokenize)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 126, in token_grouping
    tokens = regroup(tokenize(css_source, ignore_comments=False))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_token_serialize_css (tinycss.tests.tokenizing.TestTokenizer)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 256, in test_token_serialize_css
    self.run_test(token_serialize_css)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 250, in run_test
    func(self, tokenize)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 40, in token_serialize_css
    tokens = _regroup(tokenize(css_source, ignore_comments=False))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_tokens (tinycss.tests.tokenizing.TestTokenizer)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 269, in test_tokens
    self.run_test(tokens)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 250, in run_test
    func(self, tokenize)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/tokenizing.py", line 237, in tokens
    tokens = tokenize(css_source, ignore_comments=False)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_parse_font (tinycss.tests.fonts3.TestFonts3)
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/fonts3.py", line 61, in test_parse_font
    t('12pt/14pt sans-serif', size='12pt', height='14pt', family=['sans-serif'])
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tests/fonts3.py", line 57, in t
    self.ae(q, parse_font(raw))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/fonts3.py", line 97, in parse_font
    tokens = list(reversed(tuple(tokenize_grouped(type('')(css_string).strip()))))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 203, in tokenize_grouped
    return regroup(tokenize_flat(css_source, ignore_comments))
  File "/home/eschwartz/git/calibre/py3/src/tinycss/tokenizer.py", line 216, in <lambda>
    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9073: bad argument to internal function

======================================================================
ERROR: test_dictionaries (calibre.spell.dictionary.find_tests.<locals>.TestDictionaries)
  File "/home/eschwartz/git/calibre/py3/src/calibre/spell/dictionary.py", line 442, in test_dictionaries
    self.assertIn('magic. Wand', self.suggestions('magic.wand'))
  File "/home/eschwartz/git/calibre/py3/src/calibre/spell/dictionary.py", line 410, in suggestions
    if capitalize(w2) != w2:
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/icu.py", line 225, in capitalize
    return upper(x[0]) + lower(x[1:])
  File "<string>", line 5, in upper
UnicodeDecodeError: 'utf-16-le' codec can't decode byte 0x57 in position 0: truncated data

======================================================================
FAIL: test_plugins (calibre.test_build.BuildTest)
  File "/home/eschwartz/git/calibre/py3/src/calibre/test_build.py", line 117, in test_plugins
    self.assertFalse(err or not mod, 'Failed to load plugin: ' + name + ' with error:\n' + err)
AssertionError: '/home/eschwartz/git/calibre/py3/src/calibre/plugins/chmlib.so: undefined symbol: PyString_FromStringAndSize' is not false : Failed to load plugin: chmlib with error:
/home/eschwartz/git/calibre/py3/src/calibre/plugins/chmlib.so: undefined symbol: PyString_FromStringAndSize

======================================================================
FAIL: test_roundtrip (calibre.utils.icu_test.TestICU)
Test roundtripping
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/icu_test.py", line 136, in test_roundtrip
    self.ae(r, icu._icu.roundtrip(r))
AssertionError: 'xxx\x00∙🐱xxx' != 'xxx\x00∙'
- xxx∙🐱xxx
?      ----
+ xxx∙


Ran 84 tests in 3.836s

FAILED (failures=2, errors=22, skipped=7)
```

kovidgoyal:
FYI I spent a few days getting the codebase to the point where you can run bootstrap on it and most of the build tests pass. Should make it easier to get to the next milestone -- which is getting all tests to pass. Once that is achieved the real hard work of porting can begin :)

eli-schwartz:
Thanks! Currently the most obvious issue is as I said in the spoiler box just above, that the ICU/tokenize modules cause many related test_build (and various) tests to fail, e.g.

```
  File "<string>", line 5, in upper
UnicodeDecodeError: 'utf-16-le' codec can't decode byte 0x54 in position 0: truncated data

[...]

    c_tokenize_flat = tokenize_flat = lambda s, ignore_comments=False:tok.tokenize_flat(s, ignore_comments)
SystemError: Objects/unicodeobject.c:9079: bad argument to internal function
```

kovidgoyal:
Both of those are fixed in master.

kovidgoyal:
FYI I have committed code to make it possible to run calibre under python2 and python3 form the same source checkout. See README.python3

kovidgoyal:
FYI all tests now pass on python3 in master. The easy part of the porting is done. Now the painstaking work of checking python files and testing for bugs has to be done.

eli-schwartz:
I may be slightly insane...

I've written a modification to my Arch Linux PKGBUILD for calibre/calibre-git, which builds a split calibre-python3 package. It depends on the standard python2 package, but installs the compiled extensions plus python3 bytecode, and shoves the launcher scripts into /opt/calibre-python3/. Do any Arch users have an interest in me packaging it? As current git master development shows, this is still *very* WIP.

It is worth noting, however, that the GUI comes up, plugins install and some of them even run, and you can do basic ebook-viewer/ebook-edit/conversion activities -- even though all sorts of little things will throw error dialogs at random. We are slowly working our way through those...

flying-sheep:
I’d be interested, but I don’t see the advantage of depending on the Py2 package. Wouldn’t it be useful to have the pure Python 3 version packaged? Or is the current architecture an a way that the Python 2 version is still used (like unported parts still calling on Python 2 as subprocess or something?)

eli-schwartz:
No, none of it is still calling python2 subprocesses... it is just that the python3 version is sufficiently unready that I almost certainly expect you to want to have the python2 version available as a fallback.

flying-sheep:
I see. And on Arch, we don’t have “foo-common” split packages, so “calibre” would be “calibre-common + calibre-python2” and you’d add “calibre-python3”. Nice! Thank you for your hard work!

eli-schwartz:
Well, we can have split -common packages, it just very often does not make sense to -- and here, since I think everyone would want the python2 version no matter what, it doesn't seem to make sense at least until much closer to a stable python3 release.

eli-schwartz:
Split packages are not a design feature of Arch, but they're not forbidden either. Like most things, they are a "use your judgment" thing.

Anyway, I've implemented a split python3 package in my calibre-git AUR package and I have prebuilt packages in my custom repo. Check it out. :)

flying-sheep:
I get test failures, quite some of them also locale-dependent.

```pytb
======================================================================
ERROR: test_searching (calibre.db.tests.reading.ReadingTest)
Test searching returns the same data for both backends
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/reading.py", line 321, in test_searching
    '#formats:fmt1', '#formats:fmt2', '#formats:fmt1 and #formats:fmt2',
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/reading.py", line 285, in <dictcomp>
    oldvals = {query:set(old.search_getting_ids(query, '')) for query in (
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/library/caches.py", line 875, in search_getting_ids
    matches = self.parse(q)
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/utils/search_query_parser.py", line 341, in parse
    return self._parse(query, candidates=candidates)
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/utils/search_query_parser.py", line 362, in _parse
    t = self.evaluate(res, candidates)
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/utils/search_query_parser.py", line 370, in evaluate
    return self.method(parse_result[0])(parse_result[1:], candidates)
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/utils/search_query_parser.py", line 416, in evaluate_token
    return self._get_matches(location, query, candidates)
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/utils/search_query_parser.py", line 420, in _get_matches
    return self.get_matches(location, query, candidates=candidates)
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/library/caches.py", line 709, in get_matches
    return self.get_dates_matches(location, query.lower(), candidates)
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/library/caches.py", line 408, in get_dates_matches
    raise ParseException(_('Date conversion error: {0}').format(query))
ParseException: Datums-Konvertierungsfehler: 100daysago

======================================================================
FAIL: test_workers (calibre.srv.tests.loop.LoopTest)
Test worker semantics
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/srv/tests/loop.py", line 101, in test_workers
    self.ae(pool.busy, 1)
AssertionError: 0 != 1

======================================================================
FAIL: test_accept_language (calibre.srv.tests.http.TestHTTP)
Test parsing of Accept-Language
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/srv/tests/http.py", line 103, in test_accept_language
    test('en', 'en')
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/srv/tests/http.py", line 101, in test
    self.ae(r.read(), q.encode('utf-8'))
AssertionError: 'deUnbekannt' != 'enUnbekannt'

======================================================================
FAIL: test_ajax_categories (calibre.srv.tests.ajax.ContentTest)
Test /ajax/categories and /ajax/search
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/srv/tests/ajax.py", line 73, in test_ajax_categories
    self.assertIn(q, names)
AssertionError: u'Newest' not found in {u'Alle B\xfccher': u'/ajax/category/616c6c626f6f6b73/db_test_5Fuzpw', u'My Series': u'/ajax/category/23736572696573/db_test_5Fuzpw', u'Schlagw\xf6rter': u'/ajax/category/74616773/db_test_5Fuzpw', u'Autoren': u'/ajax/category/617574686f7273/db_test_5Fuzpw', u'Neue': u'/ajax/category/6e6577657374/db_test_5Fuzpw', u'My Tags': u'/ajax/category/2374616773/db_test_5Fuzpw', u'Enum': u'/ajax/category/23656e756d/db_test_5Fuzpw', u'My Rating': u'/ajax/category/23726174696e67/db_test_5Fuzpw', u'Composite Tags': u'/ajax/category/23636f6d705f74616773/db_test_5Fuzpw', u'My Authors': u'/ajax/category/23617574686f7273/db_test_5Fuzpw', u'Serie': u'/ajax/category/736572696573/db_test_5Fuzpw', u'Bewertung': u'/ajax/category/726174696e67/db_test_5Fuzpw', u'Suchen': u'/ajax/category/736561726368/db_test_5Fuzpw', u'Good Series': u'/ajax/category/40476f6f6420536572696573/db_test_5Fuzpw', u'Verlag': u'/ajax/category/7075626c6973686572/db_test_5Fuzpw', u'Sprachen': u'/ajax/category/6c616e677561676573/db_test_5Fuzpw', u'Good Authors': u'/ajax/category/40476f6f6420417574686f7273/db_test_5Fuzpw'}

======================================================================
FAIL: test_srv_restrictions (calibre.srv.tests.ajax.ContentTest)
Test that virtual lib. + search restriction works on all end points
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/srv/tests/ajax.py", line 131, in test_srv_restrictions
    ae(data['bad_restriction'], 'Invalid syntax. Expected a lookup name or a word')
AssertionError: u'Ung\xfcltiger Syntax. Erwarte einen Nachschlagenamen oder ein Wort' != u'Invalid syntax. Expected a lookup name or a word'
- Ungültiger Syntax. Erwarte einen Nachschlagenamen oder ein Wort
+ Invalid syntax. Expected a lookup name or a word


======================================================================
FAIL: test_many_many_basic (calibre.db.tests.writing.WritingTest)
Test the different code paths for writing to a many-many field
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/writing.py", line 240, in test_many_many_basic
    af(cache.set_field(name, {3:None if name == 'authors' else 'Unknown'}))
AssertionError: set([3]) is not false

======================================================================
FAIL: test_composites (calibre.db.tests.reading.ReadingTest)
Test sorting and searching in composite columns
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/reading.py", line 663, in test_composites
    self.assertEqual([1, 3, 2], cache.multisort([('#ccdate', True)]))
AssertionError: Lists differ: [1, 3, 2] != [2, 1, 3]

First differing element 0:
1
2

- [1, 3, 2]
+ [2, 1, 3]

======================================================================
FAIL: test_find_identical_books (calibre.db.tests.reading.ReadingTest)
Test find_identical_books
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/reading.py", line 700, in test_find_identical_books
    self.assertEqual(books, cache.find_identical_books(mi))
AssertionError: Items in the first set but not the second:
3

======================================================================
FAIL: test_sorting (calibre.db.tests.reading.ReadingTest)
Test sorting
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/reading.py", line 161, in test_sorting
    'Ascending sort of %s failed'%field)
AssertionError: Lists differ: [3, 2, 1] != [3, 1, 2]

First differing element 1:
2
1

- [3, 2, 1]
+ [3, 1, 2] : Ascending sort of languages failed

======================================================================
FAIL: test_get_property (calibre.db.tests.legacy.LegacyTest)
Test the get_property interface for reading data
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/legacy.py", line 145, in test_get_property
    self.assertEqual(old_vals, new_vals)
AssertionError: {u'rating': (6, 4, None), u'#rating': (4, 2, None), u'pubdate': (datetime.dateti [truncated]... != {u'rating': (6, 4, None), u'#rating': (4, 2, None), u'pubdate': (datetime.dateti [truncated]...
  {u'#authors': (u'My Author Two', u'Custom One|Custom Two', None),
   u'#comments': (u'<div>My Comments Two<p></p></div>',
                  u'<div>My Comments One<p></p></div>',
                  None),
   u'#comp_tags': (u'News, Tag One', u'Tag One, Tag Two', ''),
   u'#date': (datetime.datetime(2011, 9, 1, 6, 0, tzinfo=tzutc()),
              datetime.datetime(2011, 9, 5, 6, 0, tzinfo=tzutc()),
              None),
   u'#enum': (u'Two', u'One', None),
   u'#float': (10.01, 20.02, None),
   u'#formats': (u'FMT1, FMT2', u'FMT1', ''),
   u'#isbn': ('', '', ''),
   u'#rating': (4, 2, None),
   u'#series': (u'My Series Two', u'My Series One', None),
   u'#tags': (u'My Tag Two', u'My Tag One|My Tag Two', None),
-  u'#yesno': (0, 1, None),
?              ^  ^

+  u'#yesno': (False, True, None),
?              ^^^^^  ^^^^

   u'4_index': (3.0, 1.0, None),
   u'au_map': (u'Author Two:::Two, Author::::#:Author One:::One, Author:::',
               u'Author One:::One, Author:::',
               u'Unknown:::Unknown:::'),
   u'author_sort': (u'Two, Author & One, Author', u'One, Author', u'Unknown'),
   u'authors': (u'Author Two,Author One', u'Author One', u'Unknown'),
   u'comments': (u'<p>Comments Two</p>', u'<p>Comments One</p>', None),
-  u'cover': (1, 1, 0),
+  u'cover': (True, True, False),
   u'formats': (set([u'FMT1', u'FMT2']), set([u'FMT1']), None),
   u'id': (1, 2, 3),
   u'identifiers': (u'test:two', u'test:one', None),
   u'languages': (u'deu', u'eng', None),
   u'last_modified': (datetime.datetime(2019, 4, 30, 10, 18, 5, tzinfo=tzutc()),
                      datetime.datetime(2019, 4, 30, 10, 18, 5, tzinfo=tzutc()),
                      datetime.datetime(2011, 9, 7, 19, 54, 41, tzinfo=tzutc())),
   u'marked': (None, None, None),
-  u'ondevice': ('', '', ''),
+  u'ondevice': (u'', u'', u''),
?                +    +    +

   u'path': (u'Author Two/Title Two (1)',
             u'Author One/Title One (2)',
             u'Unknown/Unknown (3)'),
   u'pubdate': (datetime.datetime(2011, 8, 5, 6, 0, tzinfo=tzutc()),
                datetime.datetime(2011, 9, 5, 21, 6, tzinfo=tzutc()),
                datetime.datetime(2011, 9, 7, 19, 54, 41, tzinfo=tzutc())),
   u'publisher': (u'Publisher Two', u'Publisher One', None),
   u'rating': (6, 4, None),
   u'series': (u'A Series One', u'A Series One', None),
   u'series_index': (2.0, 1.0, 1.0),
-  u'series_sort': (u'A Series One', ''),
?                     --

+  u'series_sort': (u'Series One, A', u''),
?                               +++   +

   u'size': (9, 9, None),
   u'sort': (u'Title Two', u'One', u'Unknown'),
   u'tags': (set([u'News', u'Tag One']), set([u'Tag One', u'Tag Two']), None),
   u'timestamp': (datetime.datetime(2011, 9, 6, 6, 0, tzinfo=tzutc()),
                  datetime.datetime(2011, 9, 5, 21, 6, tzinfo=tzutc()),
                  datetime.datetime(2011, 9, 7, 19, 54, 41, tzinfo=tzutc())),
   u'title': (u'Title Two', u'Title One', u'Unknown'),
   u'uuid': (u'c67a09ae-ea43-4d7f-b754-d38847319512',
             u'69323d27-dc28-42b5-813b-ea9c001946ad',
             u'e029fd52-f564-451f-823a-5611ba1bef46')}

======================================================================
FAIL: test_remove_books (calibre.db.tests.add_remove.AddRemoveTest)
Test removal of books
  File "/tmp/yaytmp-1000/calibre-git/src/calibre/src/calibre/db/tests/add_remove.py", line 221, in test_remove_books
    self.assertIn(_('Unknown'), set(itervalues(authors.id_map)))
AssertionError: u'Unbekannt' not found in set([u'Unknown', u'Author One', u'Author Two'])

```

kovidgoyal:
locale related test failures are all fixed.

eli-schwartz:
Preferences => Toolbars and Menus
Try to remove an icon from a toolbar, and the following error message appears (but it is successfully removed):

```
  File "/home/eschwartz/git/calibre/src/calibre/gui2/preferences/toolbar.py", line 317, in remove_action
    self.all_actions.model().add(removed)
  File "/home/eschwartz/git/calibre/src/calibre/gui2/preferences/toolbar.py", line 124, in add
    self._data.sort()
TypeError: '<' not supported between instances of 'AuthorMapAction' and 'FakeAction'
```

These classes (Plugin and FakeAction) might need comparison methods to be defined. I'm not entirely sure what the code does when sorting, though, even on python2. AFAICT it is sorting these all based on their hash(), which is wrong. In fact, simply moving an action back and forth between a toolbar and the available actions, will cause the sort order on the left to be totally messed up and appear basically randomly.

This is a python2 bug that triggered a traceback on python3. \o/

The comparison methods won't help on python3, I think -- instead this needs a sort key. So, here is #984 to fix it.

eli-schwartz:
I think we've mostly got the GUI figured out, so now the content server gets to be a huge issue... currently it starts, but returns nothing other than an empty response body. No errors even, except for a translation error:

```
  File "/home/eschwartz/git/calibre/src/calibre/srv/loop.py", line 571, in tick
    conn.handle_event(event)
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_request.py", line 308, in parse_header_line
    self.finalize_headers(parser.hdict)
  File "/home/eschwartz/git/calibre/src/calibre/srv/web_socket.py", line 287, in finalize_headers
    return HTTPConnection.finalize_headers(self, inheaders)
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_request.py", line 347, in finalize_headers
    self.read_request_body(inheaders, request_content_length, chunked_read)
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_request.py", line 361, in read_request_body
    self.prepare_response(inheaders, BytesIO())
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_response.py", line 448, in prepare_response
    self.translator_cache, self.tdir, self.forwarded_for
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_response.py", line 236, in __init__
    self.set_translator(self.get_preferred_language())
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_response.py", line 279, in get_preferred_language
    return preferred_lang(self.inheaders.get('Accept-Language'), self.get_translator)
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_response.py", line 107, in preferred_lang
    found, lang, translator = get_translator_for_lang(x)
  File "/home/eschwartz/git/calibre/src/calibre/srv/http_response.py", line 276, in get_translator
    return get_translator_for_lang(self.translator_cache, bcp_47_code)
  File "/home/eschwartz/git/calibre/src/calibre/srv/utils.py", line 282, in get_translator_for_lang
    cache[bcp_47_code] = ans = get_translator(bcp_47_code)
  File "/home/eschwartz/git/calibre/src/calibre/utils/localization.py", line 167, in get_translator
    return found, lang, get_single_translator(lang)
  File "/home/eschwartz/git/calibre/src/calibre/utils/localization.py", line 139, in get_single_translator
    buf = io.BytesIO(zf.read(mpath + '/%s.mo' % which))
  File "/usr/lib/python2.7/zipfile.py", line 958, in read
    return self.open(name, "r", pwd).read()
  File "/usr/lib/python2.7/zipfile.py", line 984, in open
    zinfo = self.getinfo(name)
  File "/usr/lib/python2.7/zipfile.py", line 932, in getinfo
    'There is no item named %r in the archive' % name)
KeyError: "There is no item named 'en/messages.mo' in the archive"
```

kovidgoyal:
IIRC the content server mostly works fine in py3. That translation error was recently introduced and should now be fixed.

eli-schwartz:
There are various conversion fixes in #996, after which most (but not all) conversions should begin to work okay. Mostly, these deal with less common formats, though. ;)

ofek:
Keep up the good work! Is the port almost over?

eli-schwartz:
Lots of it does work. Some of it doesn't. For a fair amount of casual work, it does seem to work. ;)

I think there's a good possibility that, once the remaining conversion stuff which I am currently fighting with is resolved, we might be able to progress to public betas. :)

kovidgoyal:
The big remaining things to port are:

1) odf2xhtml -- it needs python3 compat done from scratch, the work upstreamis very low quality, for example, it modifies sys.path to get around problems with absolute imports

2) Fixing everything reported by ./setup.py unicode_check  this is hundreds of files. Most of the fixes are very simple/no impact, but every once in a while it finds actual things that will break under py3

eli-schwartz:
So yeah, some messy conversion holdouts. :)

kovidgoyal:
odf2xhtml is already done.

eli-schwartz:
So it is, so it is. My test ebook was reporting errors for rtf, but only with --debug-pipeline and that turned out to be reproducible on any historic version of calibre too... due to "RTF features that are not supported".

kovidgoyal:
What error is this? Since I am here, I might as well fix it. I tried converting a couple of RTF books with --debug-pipeline with no errors.

eli-schwartz:
```
$ ebook-convert ./Harry\ Potter\ and\ the\ Goblet\ of\ Fire\ -\ J.\ K.\ Rowling.azw3 /tmp/roundtrip.rtf
[...]
$ ebook-convert /tmp/roundtrip.rtf /tmp/roundtrip.azw3 -d /tmp/roundtrip-pipeline
Conversion options changed from defaults:
  debug_pipeline: '/tmp/roundtrip-pipeline'
1% Converting input to HTML...
InputFormatPlugin: RTF Input running
on /tmp/roundtrip.rtf
Converting RTF to XML...
Running RTFParser in debug mode
Unable to parse RTF
  File "/home/eschwartz/git/calibre/src/calibre/ebooks/conversion/plugins/rtf_input.py", line 258, in convert
    xml = self.generate_xml(stream.name)
  File "/home/eschwartz/git/calibre/src/calibre/ebooks/conversion/plugins/rtf_input.py", line 115, in generate_xml
    parser.parse_rtf()
  File "/home/eschwartz/git/calibre/src/calibre/ebooks/rtf2xml/ParseRtf.py", line 319, in parse_rtf
    self.__bracket_match('make_preamble_divisions')
  File "/home/eschwartz/git/calibre/src/calibre/ebooks/rtf2xml/ParseRtf.py", line 556, in __bracket_match
    raise RtfInvalidCodeException(msg)
calibre.ebooks.rtf2xml.ParseRtf.RtfInvalidCodeException: closed bracket doesn't match, line 193 in file make_preamble_divisions


During handling of the above exception, another exception occurred:

  File "/usr/local/calibre-3/bin/ebook-convert", line 20, in <module>
    sys.exit(main())
  File "/home/eschwartz/git/calibre/src/calibre/ebooks/conversion/cli.py", line 400, in main
    plumber.run()
  File "/home/eschwartz/git/calibre/src/calibre/ebooks/conversion/plumber.py", line 1107, in run
    accelerators, tdir)
  File "/home/eschwartz/git/calibre/src/calibre/customize/conversion.py", line 246, in __call__
    log, accelerators)
  File "/home/eschwartz/git/calibre/src/calibre/ebooks/conversion/plugins/rtf_input.py", line 262, in convert
    'support. Convert it to HTML first and then try it.\n%s')%e)
ValueError: This RTF file has a feature calibre does not support. Convert it to HTML first and then try it.
closed bracket doesn't match, line 193 in file make_preamble_divisions
```

It should hopefully be easy to reproduce, since I can reproduce on multiple books if I roundtrip them through ebook-convert.

kovidgoyal:
Should be fine now.

aimylios:
One of the tests fails for me on a Python 3 build of Calibre (revision abd9e6b):
```
test_palmdoc_compression (calibre.ebooks.compression.palmdoc.find_tests.<locals>.Test) ... ERROR [0.0004 s]

======================================================================
ERROR: test_palmdoc_compression (calibre.ebooks.compression.palmdoc.find_tests.<locals>.Test)
  File "/home/aimylios/calibre-3.45.0/src/calibre/ebooks/compression/palmdoc.py", line 92, in test_palmdoc_compression
    self.assertEqual(py_compress_doc(test), x)
  File "/home/aimylios/calibre-3.45.0/src/calibre/ebooks/compression/palmdoc.py", line 51, in py_compress_doc
    och = ord(ch)
TypeError: ord() expected string of length 1, but int found

Ran 316 tests in 44.733s

FAILED (errors=1, skipped=10)
get_categories: item Unknown is not in authors list!
```
I'm pretty sure all of the tests passed when I did the same exercise about a month ago. It might thus be related to the changes merged on 26th of May ([f2e0181](https://github.com/kovidgoyal/calibre/commit/f2e01819265dc173e3bf672174054de8f07381d6#diff-5f539a0cb3efaeade875d3d9848947c4) and [b842fe7](https://github.com/kovidgoyal/calibre/commit/b842fe758a56e1b855ff4d3276e63ddaf2d53897#diff-5f539a0cb3efaeade875d3d9848947c4)).

eli-schwartz:
In the second commit you reference, we *started* collecting a unittest for that function. It was probably always subtly broken -- in python3, a b'' string is a bytearray and indexing it returns a number, not a b'' string.

miasma:
Tried the Arch git build from AUR. The latest version fails with

```
test_prints_nothing_if_no_errors (calibre.db.cli.tests.PrintCheckLibraryResultsTest) ... ok [5e-05 s]

======================================================================
FAIL: test_bonjour (calibre.srv.tests.loop.LoopTest)
Test advertising via BonJour
  File "/tmp/makepkg/calibre-git/src/calibre/src/calibre/srv/tests/loop.py", line 129, in test_bonjour
    self.assertIsNotNone(info)
AssertionError: unexpectedly None

======================================================================
FAIL: test_searching (calibre.db.tests.reading.ReadingTest)
Test searching returns the same data for both backends
  File "/tmp/makepkg/calibre-git/src/calibre/src/calibre/db/tests/reading.py", line 338, in test_searching
    ans, nr, query))
AssertionError: Items in the second set but not the first:
2 : Old result: set([1]) != New result: set([1, 2]) for search: date:9/6/2011

Ran 316 tests in 41.428s

FAILED (failures=2, skipped=9)
```
python-zeroconf is installed at least.

eli-schwartz:
The package has dependencies on python-zeroconf so that should always be available anyway. ;)

On the topic of your errors, I do not get either error. However, note that the first error will unpredictably occur on both python2 and python3, so I usually just try rerunning the testsuite and it often succeeds again.

Not sure why the second error would happen.

eli-schwartz:
FWIW, I did update the binary repository mentioned in the calibre-git AUR package details. So you can get the python3 build as a prebuilt package.

eli-schwartz:
@kovidgoyal I wonder if you've noticed a curious oddity that I am noticing? Seems like when running the python2 and python3 tests in quick succession, regardless of how often test_bonjour/test_websocket_perf flakes out on python2, it seems to be much more reliably succeeding on python3.

I just ran the tests 5 times in a row, and on python2 it succeeded once, failed 4 times. On python3, it succeeded all 5 times.

kovidgoyal:
The bonjour test is utterly reliable for me on both pythons. The http_basic python3 test failure I do see from time to time, have to investigate that as it indicates pielining is out of order which is a pretty bad bug.

kovidgoyal:
Actually given that pipelining is not really used by browsers or even
curl anymore, maybe best to just drop that test.

eli-schwartz:
So... since calibre seems to be more or less running fairly stable so far...

With the calibre 4.4.0 release, I've uploaded a split package to the official archlinux [community] repository. `calibre` now depends on `calibre-common`, which contains all python-agnostic code, and `calibre` and `calibre-python3` can now be installed together. Which one you use when you try to run the command line tools depends on the symlinks that were created by `calibre-alternatives set [2|3]`.

While I anticipate that people will still want to use python2, simply because plugins are a very important part of the ecosystem, people now have the option to install the python3 version and try it out. Hopefully this will help us discover any more lingering bugs.

kovidgoyal:
Look at the py3 branch -- I am working on getting binary builds done
with python3. Once that is done, hopefully the plugin developers can
work on porting their plugins using those builds.

eli-schwartz:
Looking forward to it. :)

kovidgoyal:
FYI binary builds are available as described here: https://www.mobileread.com/forums/showthread.php?t=325721

@escwartz: Feel free to post in that thread if you have tips for porting plugins to python 3.

pevik:
Hm, it'd be great if python3 port was finished, because new distros already dropped python2 but you require it for `setup.py bootstrap` (I know I can download precompiled calibre or compile PyQt5 for python2 myself). So what is missing?

eli-schwartz:
@pevik 

You want this for OpenSUSE, correct? You can import Fedora 31's specfile here: https://src.fedoraproject.org/rpms/calibre/blob/f31/f/calibre.spec

Debian likewise builds with python3, which "mostly works" albeit not all plugins are ported.

Gentoo and Arch Linux provide the option to use either python2 or python3.


I'm not positive what the porting stats are for https://plugins.calibre-ebook.com/, unlike the pyqt4 -> pyqt5 move there isn't an obviously easy way to heuristically prove that a plugin has been ported... the dates seem to indicate many many plugins were not ported yet, though I know a couple of my favorite ones were. (We cannot just "finish" the porting by declaration and make all those work.)

pevik:
@eli-schwartz No, for Debian, which has PyQt5 uninstallable for python2 in Debian testing release (it has python2 itself). Also PyQt5 does not provide builds in pip for python2 any more. But thanks for info.
Understand, to port 136 plugins is a lot work (even some of them are already ported).

eli-schwartz:
Oh, I was just looking at your membership in @SUSE :p

Debian sid/Bullseye builds calibre using python3 (it must, since python2 is being forcibly removed), Debian stable/Buster builds calibre using python2. What is the problem exactly?

/cc @norbusan

norbusan:
Yes, Debian/buster is the last with Python2.
Debian/testing and sid are in the process of removing Python2, lots of packages have already disappeared.
Thus, we need to build Calibre on with Python3 on sid/testing.

pevik:
I use both distros and want calibre package was working in both. As I wrote I can use precompiled package or compile also PyQt5, but that's not ideal.

**Debian**: I reported Debian package [1], where you put many suggestions [2]. Unfortunately I cannot fix it as I haven't forced to learn Debian packaging (I'm not a Debian developer, but that could be solved via [3] or even to send a patch in a bug tracker).

**openSUSE**: There is updated package in [4]. I'll test the bug in it. Fixing openSUSE package would be for me much easier than Debian package.

Out of curiosity for python3 port: esting your `py3` branch (that's the one which  it looks to be working (2 test fails: `test_unrar` and `test_get`., but that's minor issue). So now we're waiting on plugins being ported?

[1] https://bugs.launchpad.net/calibre/+bug/1850835
[2] https://bugs.launchpad.net/calibre/+bug/1850835/comments/7
[3] https://mentors.debian.net/
[4] https://build.opensuse.org/package/show/Documentation:Tools/calibre

eli-schwartz:
Note that most or possibly all of my suggestions from https://bugs.launchpad.net/calibre/+bug/1850835/comments/7 are more or less obsolete. @norbusan has done some great work to improve calibre packaging on Debian, so as far as I know the latest versions should be in pretty good shape!)

Where is test_unrar and test_get failing? The former might be failing due to licensing issues preventing your distribution from providing the libunrar software, and thus, the python3-unrardll module which links to it.

ofek:
Hey quick question: have we considered using the officially Qt supported https://www.qt.io/qt-for-python (https://pypi.org/project/PySide2/)

kovidgoyal:
On Tue, Mar 24, 2020 at 08:17:45PM -0700, Ofek Lev wrote:
> Hey quick question: have we considered using the officially Qt supported https://www.qt.io/qt-for-python (https://pypi.org/project/PySide2/)

No, we haven't. I am very happy with PyQt and have been for over a
decade.

eli-schwartz:
My sole experience with PySide2 has been it crashing because for subtle reasons it didn't support python 3.8 yet after our distro python 3.8 mass rebuild. PyQt just worked...

ofek:
Might be related to note here https://wiki.qt.io/Qt_for_Python

> 5.14 will not work on Windows with Python 3.8.0, please use Python 3.8.1 or greater.

Which doesn't surprise me, nearly every initial `.0` Python release has had a few issues

eli-schwartz:
No, it was actually https://bugs.archlinux.org/task/64514 and it was solved by "wait for qt 5.14" which was months after python 3.8, and the bug was in fact solved by PySide, so this has nothing to do with "nearly every initial .0 Python release has had a few issues".

PR #871, opened 2018-09-04
Convert syntax to meet py3 standards

flaviut:
This has just one easy-to-fix syntax error in python 3:

```console
$ python -m compileall src build recipes/**/*.recipe manual -q
*** Error compiling 'src/duktape/__init__.py'...
  File "src/duktape/__init__.py", line 160
    raise JSError(e), None, sys.exc_info()[2]
                    ^
SyntaxError: invalid syntax
```

I also patched `python3.7/compileall.py:137` to be

```python
        head, tail = os.path.splitext(name)
        if tail in {'.py', '.recipe'}:
```

in order to test the syntax of the `.recipe` files

This patch was tested following this process:

- [x] `python ./setup.py bootstrap --clean-all`
- [x] `SIP_DIR=/usr/share/sip python ./setup.py bootstrap`
- [x] `python ./setup.py test`
    - 1 error in test_bonjour, both before & after changes
- [x] `python3 -m compileall src build recipes/**/*.recipe manual -q`


flaviut:
> Also, just a note on general principles. I am not a big fan of six. It has burned me in the past with incompatible changes and it is very heavyweight.

I haven't been using six for very long, so I haven't run into any issues. However, I did [take a brief look at `six.py`](https://github.com/benjaminp/six/blob/master/six.py), and I don't see anything particularly crazy.

> The calibre codebase does not doa lot of fancy stuff, so what we need for compatibility is only a small subset of six.

This is definitely true.

> I would prefer if we create a dedicated module for it with just what we need, much lighter weight. For an example of such a module, see python-mechanize/mechanize#9 where I wrote one for mechanize

I did take at `polyglot.py`, but there's a few things I like about six:

- once py2 support is dropped, six makes it easy to just find-replace `six.moves.` -> `<nothing>`. `polyglot.py` imports everything into one namespace, so this is harder.
- https://python-modernize.readthedocs.io/en/latest/ uses the six module in its automated conversions to python3, which makes the conversion much easier.

--------------------------------

I bring this up here because six or a similar library is needed to replace `raise JSError(e), None, sys.exc_info()[2]`

kovidgoyal:
six is about a thousand lines of code that takes signficant time (several milliseconds) to import. polyglot is a hundred lines of code by contrast. But, my main issue with six is that in the past it has made backwards incompatible changes that broke working code for me. For something that requires so little code to write, and that is used so widely, I prefer not adding an eternal dependency. That way we are in control of the code.

As for removing it when dropping py2 support, that is esaily done with some help from pyflakes and a little cleanup script. It also has the advantage that it is both faster (avoid extra dict lookup) and more pleasant to use (less to type).

However, I dont want you to have to do extra work for my preferences. So feel free to use six in your PRs that use the modernize library. I will take care of migrating them to polyglot.

kovidgoyal:
Note that there is going to be a calibre release in a couple of days, I will review and merge this PR after that.

flaviut:
Sounds good to me. Ping me once you make the release, and I'll be happy to rebase onto current master.

kovidgoyal:
The release has been made

kovidgoyal:
As per roadmap above, I have ported the setup package to the extent that setup.py build now runs. Now it is over to you to port the C extensions. Please send a separate PR for each that I will review and merge.

kovidgoyal:
Oh and just FYI you can build individual C extension using ./setup.py build --only

see ./setup.py build --help for details

PR #874, opened 2018-09-14
Port Monotonic to py3

flaviut:


kovidgoyal:
Doesn't py3 have a built in monotonic. You could just edit utils.monotinic and use that in py3

flaviut:
Good catch, I didn't realize that. PR updated to used `time.monotonic()` in py3.

PR #875, opened 2018-09-14
update unicode_names for py3

flaviut:


src/unicode_names/unicode_names.c
```diff
@@ -7,18 +7,6 @@
 
 #include "names.h"
 
-static PyObject*
```
flaviut: dead code as far as I can tell--a grep shows that it isn't used anywhere.

src/unicode_names/unicode_names.c
```diff
@@ -84,38 +72,39 @@ nfc(PyObject *self UNUSED, PyObject *args) {
...
 
-static PyMethodDef module_methods[] = {
-    {"all_words", (PyCFunction)all_words, METH_NOARGS, ""},
-    {"codepoints_for_word", (PyCFunction)cfw, METH_VARARGS, ""},
-    {"name_for_codepoint", (PyCFunction)nfc, METH_VARARGS, ""},
+static PyMethodDef unicode_names_methods[] = {
+    {"codepoints_for_word", (PyCFunction)cfw, METH_VARARGS,
+     "Return a set of integer codepoints for where each codepoint's name "
+     "contains ``word``,"},
+    {"name_for_codepoint", (PyCFunction)nfc, METH_VARARGS,
```
flaviut: `unicodedata.name()` exists in both python2  & 3, but it's way out of date in py2. In any case, I think it's best to use this library for both functions so that they don't fall out of sync.

PR #879, opened 2018-09-29
[WIP] Build speedups for py3

flaviut:


src/calibre/utils/speedup.c
```diff
@@ -128,23 +128,25 @@ speedup_detach(PyObject *self, PyObject *args) {
...
     int fd = -1, bufsize = -1;
     char *mode = NULL;
 
-    if (!PyArg_ParseTuple(args, "iOs|i", &fd, &name, &mode, &bufsize)) return NULL;
+    if (!PyArg_ParseTuple(args, "iss|i", &fd, &name, &mode, &bufsize)) return NULL;
+#if PY_MAJOR_VERSION >= 3
+    ans = PyFile_FromFd(fd, name, mode, bufsize, NULL, NULL, NULL, 1);
+#else
+    ans = PyFile_FromFile(fp, name, mode, fclose);
     fp = fdopen(fd, mode);
```
kovidgoyal: should come before pyfile_fromfile not after it

src/calibre/utils/speedup.c
```diff
@@ -353,6 +357,41 @@ clean_xml_chars(PyObject *self, PyObject *text) {
...
+                (0xe000 <= ch && ch <= 0xfffd) ||
+                (0xffff < ch && ch <= 0x10ffff)) {
+            // we can overwrite the same buffer that we're reading from because
+            // all our writes will occur at or before the current character
+            ucs4_text[target_i] = ch;
+            target_i += 1;
+        }
+    }
+    result = PyUnicode_FromKindAndData(
+        PyUnicode_4BYTE_KIND, ucs4_text, target_i);
```
kovidgoyal: Do we really need to allocate the buffer twice? Also I dont like up-converting all strings to UCS 4 internally. It would be bad for performance since most XML data will be either ascii or ucs 2 at worst.

src/calibre/utils/speedup.c
flaviut: I was debugging this right before I went to bed, I have no idea how I missed it :)

src/calibre/utils/speedup.c
flaviut: > Do we really need to allocate the buffer twice?

Yes. PyUnicode_New() explicitly says that the objects created are not
resizeable. Therefore 2 allocations are needed:

1. the scratch space in which the filtered data is written
2. the PyObject that is returned

The python 2 code above digs into cpython implementation details (`ans->length
= j;` could be written as `PyUnicode_GET_SIZE(ans) = j;`).

There are a couple other options here:
 - do two passes over the array, first calculating the required length
 - ignore the docs and overwrite the PyUnicode length

> Also I dont like up-converting all strings to UCS 4 internally. It would be
> bad for performance since most XML data will be either ascii or ucs 2 at
> worst.

You're right, there is significant performance impact:

py3 gcc-4.8:

```console
$ python -m timeit -s  'import speedup; v = "a"*1000000000' -n 2 'speedup.clean_xml_chars(v)'
2 loops, best of 5: 2.81 sec per loop
```

py2:

```console
$ python -m timeit -s  'import speedup; v = u"a"*1000000000' -n 2 'speedup.clean_xml_chars(v)'
2 loops, best of 3: 1.46 sec per loop
```

However, most of this worse performance is inside `clean_xml`, and additional
copying is ~8% of the time. 

I'll see what I can do about this.

kovidgoyal: The way I see it working is the following:

1) Basically write three versions of the function for ascii string, ucs2 and ucs4. The character validity checks in the case of ascii are also much simpler, which should be a small speedup

2) The vast majority of the time this function really does nothing, i.e.e there are no invalid chars in the XML. Therefore I think we should simply allocate a single buffer at the same size as the original string using PyMem_NEW(), fill it up and then create a unicode objet using PyUnicode_FromKindAndData

kovidgoyal:
Nevermind, I looked at the py 3 source code and PyUnicodeFromKindAndData does makes a copy, sigh. Seems like a glaring hole in the python API. I suppose we either have to accept two allocs or mess with the unicode object internals ourselves.

flaviut:
```console
$ python3 -m timeit -s  'import speedup; v = u"💩"*100000000' -n 2 'speedup.clean_xml_chars(v)'
2 loops, best of 5: 367 msec per loop
$ python3 -m timeit -s  'import speedup; v = u"a"*100000000' -n 2 'speedup.clean_xml_chars(v)'
2 loops, best of 5: 199 msec per loop
$ python2 -m timeit -s  'import speedup; v = u"💩"*100000000' -n 2 'speedup.clean_xml_chars(v)'
2 loops, best of 3: 586 msec per loop
$ python2 -m timeit -s  'import speedup; v = u"a"*100000000' -n 2 'speedup.clean_xml_chars(v)'
2 loops, best of 3: 152 msec per loop
```

flaviut:
Unicode strings in python 3.3+ are fixed width: see the `PyUnicode_New(size, maxchar)`. `maxchar` is used to allocate strings with the appropriate character size so that no surrogate pairs are used. For more details, you can take a look at https://www.python.org/dev/peps/pep-0393/

PR #881, opened 2018-10-01
Use builtin zlib in py3

flaviut:
Done by disabling building zlib extension in py3, and falling back to
py's zlib when calibre's zlib extension fails to load.

However, I can't see much point in calibre shipping it's own zlib2 module, in either py2 or py3:

With calibre's zlib2:

```console
$ python2 -m timeit -s  'import test; f = open("./rand", "rb")' -n 2 'list(test.compress_readable_output(f))'
2 loops, best of 3: 1.39 sec per loop
```

With python's zlib2:

```console
$ python3 -m timeit -s  'import test; f = open("./rand", "rb")' -n 2 'list(test.compress_readable_output(f))'
2 loops, best of 5: 1.39 sec per loop
$ python2 -m timeit -s  'import test; f = open("./rand", "rb")' -n 2 'list(test.compress_readable_output(f))'
2 loops, best of 3: 1.4 sec per loop
```

test.py:

```python
import struct
#import zlib2 as zlib
import zlib
from io import DEFAULT_BUFFER_SIZE
def gzip_prefix():
    # See http://www.gzip.org/zlib/rfc-gzip.html
    return b''.join((
        b'\x1f\x8b',       # ID1 and ID2: gzip marker
        b'\x08',           # CM: compression method
        b'\x00',           # FLG: none set
        # MTIME: 4 bytes, set to zero so as not to leak timezone information
        b'\0\0\0\0',
        b'\x02',           # XFL: max compression, slowest algo
        b'\xff',           # OS: unknown
    ))

def compress_readable_output(src_file, compress_level=6):
    crc = zlib.crc32(b"")
    size = 0
    zobj = zlib.compressobj(compress_level,
                            zlib.DEFLATED, -zlib.MAX_WBITS,
                            zlib.DEF_MEM_LEVEL, zlib.Z_DEFAULT_STRATEGY)
    prefix_written = False
    while True:
        data = src_file.read(DEFAULT_BUFFER_SIZE)
        if not data:
            break
        size += len(data)
        crc = zlib.crc32(data, crc)
        data = zobj.compress(data)
        if not prefix_written:
            prefix_written = True
            data = gzip_prefix() + data
        yield data
    yield zobj.flush() + struct.pack(b"<L", crc & 0xfffff
```

`./rand` is 256MiB of random data.

Unless there's something I'm missing, I propose rejecting this PR and instead deleting `zlib2.c`. I'd be happy to make a PR for that too.

### DIFF
--- a/setup/extensions.json
+++ b/setup/extensions.json
@@ -31,7 +31,8 @@
         "inc_dirs": "!zlib_inc_dirs",
         "lib_dirs": "!zlib_lib_dirs",
         "libraries": "z",
-        "windows_libraries": "zlib"
+        "windows_libraries": "zlib",
+        "needs_py2": true
     },
     {
         "name": "certgen",
--- a/src/calibre/srv/http_response.py
+++ b/src/calibre/srv/http_response.py
@@ -12,9 +12,8 @@ from io import BytesIO, DEFAULT_BUFFER_SIZE
 from itertools import chain, repeat, izip_longest
 from operator import itemgetter
 from functools import wraps
-from polyglot.builtins import map
 
-from polyglot.builtins import reraise
+from polyglot.builtins import reraise, map, is_py3
 
 from calibre import guess_type, force_unicode
 from calibre.constants import __version__, plugins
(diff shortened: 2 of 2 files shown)


kovidgoyal:
If you use git blame on the file where it was used, you will see why calibre ships its own: https://github.com/kovidgoyal/calibre/commit/83151cc1f07d9bc070da3a65ac463771a1479ea6

flaviut:
> If you use git blame on the file where it was used, you will see why calibre ships its own: 83151cc

I did take a look, but I guess I didn't fully understand what was going on. Using memoryviews does appear to be faster:

```console
$ python3 -m timeit -s  'import test; f = test.ReadOnlyFileBuffer(open("./rand", "rb").read())' -n 2 'list(test.compress_readable_output(f))'
2 loops, best of 5: 1.37 sec per loop
$ python2 -m timeit -s  'import test; f = test.ReadOnlyFileBuffer(open("./rand", "rb").read())' -n 2 'list(test.compress_readable_output(f))'
  File "/usr/lib/python2.7/timeit.py", line 322, in main
    r = t.repeat(repeat, number)
  File "/usr/lib/python2.7/timeit.py", line 230, in repeat
    t = self.timeit(number)
  File "/usr/lib/python2.7/timeit.py", line 202, in timeit
    timing = self.inner(it, self.timer)
  File "<timeit-src>", line 6, in inner
    list(test.compress_readable_output(f))
  File "./test.py", line 67, in compress_readable_output
    crc = zlib.crc32(data, crc)
TypeError: crc32() argument 1 must be string or read-only buffer, not memoryview
```

> Regarding this PR, I'd prefer it if we used ispy3 explictly and imported thebuiltin zlib module in that case and used the bundled one otherwise.
I've updated the PR to do this

PR #889, opened 2018-10-11
Build matcher in py3

flaviut:
I'm not sure how this ever worked without PyTypeObject.tp_new, but
adding a generic constructor seems to work fine.

### DIFF
--- a/src/calibre/utils/icu_calibre_utils.h
+++ b/src/calibre/utils/icu_calibre_utils.h
@@ -122,6 +122,7 @@ static UChar* python_to_icu(PyObject *obj, int32_t *osz) {
     UChar *ans = NULL;
     Py_ssize_t sz = 0;
     UErrorCode status = U_ZERO_ERROR;
+    Py_UCS2 *data;
     int i;
 
     if (!PyUnicode_CheckExact(obj)) {
@@ -150,10 +151,14 @@ static UChar* python_to_icu(PyObject *obj, int32_t *osz) {
         break;
     case PyUnicode_2BYTE_KIND:
         ans = (UChar*) malloc((sz+1) * sizeof(UChar));
-        // UChar might be more than 2 bytes, so we need to copy manually.
-        // Hopefully this will be optimized as memcpy where possible.
-        for(i = 0; i < sz; i++) {
-            ans[i] = PyUnicode_2BYTE_DATA(obj)[i];
+        data = PyUnicode_2BYTE_DATA(obj);
+        // if UChar is more than 2 bytes, we need to copy manually.
+        if(sizeof(UChar) != sizeof(Py_UCS2)) {
+            for(i = 0; i < sz; i++) {
+                ans[i] = data[i];
+            }
+        } else {
+            memcpy(ans, data, sz);
         }
         // add null terminator
         ans[sz] = 0;
--- a/src/calibre/utils/icu_calibre_utils.h
+++ b/src/calibre/utils/icu_calibre_utils.h
@@ -22,10 +22,11 @@
 #include <unicode/unorm2.h>
 #include <unicode/ubrk.h>
 
-#if PY_VERSION_HEX >= 0x03030000
-#error Not implemented for python >= 3.3
+#if PY_VERSION_HEX < 0x03030000 && PY_VERSION_HEX > 0x03000000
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Building is failing on windows, see https://ci.appveyor.com/project/kovidgoyal/calibre/builds/19419995/job/908t6ab0exr4gj81

src/calibre/utils/icu_calibre_utils.h
```diff
@@ -22,25 +22,26 @@
...
-static UChar* python_to_icu(PyObject *obj, int32_t *osz, uint8_t do_check) {
+static UChar* python_to_icu(PyObject *obj, int32_t *osz) {
     UChar *ans = NULL;
     Py_ssize_t sz = 0;
 #ifdef Py_UNICODE_WIDE
     UErrorCode status = U_ZERO_ERROR;
 #endif
 
-    if (do_check && !PyUnicode_CheckExact(obj)) {
+    if (!PyUnicode_CheckExact(obj)) {
```
kovidgoyal: Wont this change prevent using unicode sub-classes?

flaviut:
Yup, and everywhere else too. I forgot to replace the `matcher_methods` entry with NULL after deleting that. The fixed builds finished just as you posted that.

flaviut:
(and my test script)

```python
from __future__ import print_function
from matcher import Matcher
import os


DEFAULT_LEVEL1 = '/'
DEFAULT_LEVEL2 = '-_ 0123456789'
DEFAULT_LEVEL3 = '.'

class CScorer(object):

    def __init__(
            self,
            items,
            level1=DEFAULT_LEVEL1,
            level2=DEFAULT_LEVEL2,
            level3=DEFAULT_LEVEL3
    ):
        print(items)
        self.m = Matcher(
            items,
            level1, level2, level3
        )

    def __call__(self, query):
        scores, positions = self.m.calculate_scores(query)
        for score, pos in zip(scores, positions):
            yield score, pos


def get_items_from_dir(basedir, acceptq=lambda x: True):
    relsep = os.sep != '/'
    for dirpath, dirnames, filenames in os.walk(basedir):
        for f in filenames:
            x = os.path.join(dirpath, f)
            if acceptq(x):
                x = os.path.relpath(x, basedir)
                if relsep:
                    x = x.replace(os.sep, '/')
                yield x

def main():
    basedir = input('Enter directory to scan [%s]: ' % os.getcwd()).strip() or os.getcwd()
    m = CScorer(list(get_items_from_dir(basedir)))
    while True:
        query = input('Enter query: ')
        data = list(m(query))
        print(data)

main()
```

src/calibre/utils/icu_calibre_utils.h
flaviut: `do_check` was never anything but 1, so I removed it. I wasn't exactly sure why `PyUnicode_CheckExact` was used instead of `PyUnicode_Check`, but I figured that there was some reason, so I left it alone. I can change it to `PyUnicode_Check`

kovidgoyal: ah well, probably best to leave it alone in that case, unless we have some actual reason to worry about unicode sub-classes. I confess I dont recall why I used checkexact there either :)

src/calibre/utils/icu_calibre_utils.h
```diff
@@ -114,3 +115,110 @@ static PyObject* icu_to_python(UChar *src, int32_t sz) {
...
+        }
+        u_strFromUTF8(
+            ans, sz + 1,
+            (int32_t*) osz,
+            (char*) PyUnicode_1BYTE_DATA(obj),
+            (int32_t) sz,
+            &status);
+        break;
+    case PyUnicode_2BYTE_KIND:
+        ans = (UChar*) malloc((sz+1) * sizeof(UChar));
```
kovidgoyal: cant we use sizeof(UChar) to check?

flaviut: Yes, you're right. I even looked at the assembly, and was surprised to see that GCC-4.8 doesn't automatically switch to memcpy (although it does use a unrolled loop)

kovidgoyal:
I have merged making some efficiency improvements to python_to_icu32 (py3 version)

PR #891, opened 2018-10-12
Convert html to py3

flaviut:
I haven't been able to test this in py3 because this module depends pretty closely on other calibre internals. `Tag()` and the attributes do seem to work fine though.


kovidgoyal:
just a heds up, I will likely be working on the two CHM related extensions in the near future, because of https://bugs.launchpad.net/bugs/1796889

so skip porting those two until all the other extensions have been ported.

kovidgoyal:
I have fixed the chm bug (which was caused by the code porting chm.py to python 3), here: https://github.com/kovidgoyal/calibre/commit/d517a8eb4fc8e0ddd74df9d871096eb3be87e341

PR #905, opened 2018-11-25
Build tokenizer on py3

flaviut:
- The actual unicode manipulation wasn't converted to the new 3.3+
style, since it would lead to lots of ifdefs and become quite ugly. This
can be done when py2 support is dropped. The drawbacks will be
temporary slower code (there will be extra copying required).
- The CLEANUP() macro is also used more extensively to avoid redundant
code

I also used my IDE's formatting tools to add some newlines and make things a little easier to read. If you'd like, I'd be happy to revert these changes. You can ignore the whitespace changes in the github diff by appending `?w=1` to the url.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ recipes/debug
 /.metadata/
 .idea
 /*env*/
+cmake-build-*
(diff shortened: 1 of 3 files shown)


kovidgoyal:
seems to be segfaulting somewhere between lines 380 and 395 in setup.resources.py

flaviut:
Yup, I noticed that and then got caught up in some other tasks. I'll ping you once I've fixed it.

flaviut:
@kovidgoyal should be fixed now, just have to wait for CI to run.

kovidgoyal:
I have ported all the code to be py2/py3 compatible. Didn't require that many defines.I haven't tested the py3 version.

PR #908, opened 2018-12-10
Update ICU to py3

flaviut:


flaviut:
What's the rationale in doing

```c
#if PY_MAJOR_VERSION < 3
    temp = PyInt_FromLong((long)input_buf[i]);
#else
    temp = PyLong_FromLong((long)input_buf[i]);
#endif
```

instead of

```c
temp = PyLong_FromLong((long)input_buf[i]);
```

?

kovidgoyal:
ints take less space to store. No reson to cause a performancce
regression (however minor) just to avoid an ifdef.

PR #915, opened 2018-12-25
Port chmlib and chm_extra plugins to build on python2/python3

eli-schwartz:


src/calibre/utils/chm/extra.c
```diff
@@ -42,6 +51,13 @@
 #                       define MODEXPORT(a) __declspec(dllexport) a
 #                       define MODIMPORT(a) extern a
 #               endif
+#define uint64_t unsigned long long
+#define uint32_t unsigned int
+#define uint16_t unsigned short
+#define uint8_t  unsigned char
+#define size_t   int
```
kovidgoyal: this looks wrong. as of VS 2015 (which calibre uses on windows, all these types should be defined anyway, so no need to redefine them here.

eli-schwartz: It is automagically generated by swig, probably under the assumption that people use very old compilers.

kovidgoyal:
Also, I seriously doubt this will actually work on py3, the python bits as oppossed to the c bits are going to require refactoring most likely, because they currently work on bytes and the python 3 bytes type is much less capable, so if anybody does anything like slicing or regexes, thing are likely to break.

src/calibre/utils/chm/extra.c
eli-schwartz: Perhaps https://github.com/dottedmag/pychm/pull/5 is meant to work on very old compilers.

kovidgoyal: yeah but even then size_t is an unsigned type while int is signed. That cant be right.

kovidgoyal:
I decided to modernize the SWIG wrapper for chmlib. It now has no hand written C code an no swig generated python code, so it should be buildable on python 3 automatically.

PR #916, opened 2018-12-28
build: add options to use system mathjax

eli-schwartz:
Some mathjax distributions come with the unpacked versions removed as per the upstream instructions for optimizing an installation. There's no reason to explicitly prefer the unpacked version, regardless of the content server using compression.

Also filter the .woff files to make sure they are actually .woff files, since they may be files like fonts.dir, fonts.scale, .uuid (created by xorg-mkfontdir, xorg-mkfontscale, fontconfig).

My motivation for this change is so I can have my calibre-git distro package use a makedepends on mathjax and pack up the contents of /usr/share/mathjax (which is an optimized version).

### DIFF
--- a/setup/install.py
+++ b/setup/install.py
@@ -77,7 +77,7 @@ class Develop(Command):
         if not opts.prefix:
             opts.prefix = sys.prefix
         for x in ('prefix', 'libdir', 'bindir', 'sharedir', 'staging_root',
-                'staging_libdir', 'staging_bindir', 'staging_sharedir'):
+                'staging_libdir', 'staging_bindir', 'staging_sharedir', 'mathjax_dir'):
             o = getattr(opts, x, None)
             if o:
                 setattr(opts, x, os.path.abspath(o))
@@ -101,6 +101,7 @@ class Develop(Command):
         self.staging_sharedir = getattr(opts, 'staging_sharedir', None)
         if self.staging_sharedir is None:
             self.staging_sharedir = opts.staging_sharedir = self.j(opts.staging_root, 'share')
+        self.mathjax_dir = getattr(opts, 'mathjax_dir', None)
 
         self.staging_libdir = opts.staging_libdir = self.j(self.staging_libdir, 'calibre')
         self.staging_sharedir = opts.staging_sharedir = self.j(self.staging_sharedir, 'calibre')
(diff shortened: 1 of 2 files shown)


eli-schwartz:
Also added installation option `--mathjax-dir` which, if given, will make the viewer resources symlink to a system-provided copy instead of the bundled mathjax.

kovidgoyal:
The reason I prefer unpacked versions is that it makes unhandled tracebacks much easier to decipher. I like users being able to report meaningful error messages to me that I can often use to fix an issue without needing to first reproduce it. So I am afraid that is not going to change.

eli-schwartz:
Two points:
- could you merge the second commit?
- I have pushed a commit that uses the unpacked resources if possible, but falls back on whatever is available. Would this be sufficient?

kovidgoyal:
But isn't the system mathjax always packed? My point is I dont want to use packed versions of mathjax.

eli-schwartz:
It might be, but not by default and not for the prebuilt binary releases. Looks like Debian at least debundles the viewer mathjax already, but their libjs-mathjax package does include the unpacked/ directory.

For the content-server zip, I'd only make use of this when bootstrapping from git master, not when building a stable release (so only with the AUR calibre-git package).

For the viewer, it would have the benefit of automatically upgrading to the latest mathjax version (calibre's copy appears to be unmodified) as well as resulting in somewhat smaller downloads and reduced disk usage when using other packages that also depend on mathjax... I would expect people reporting reader issues to first duplicate the issue using the official binaries anyway.

kovidgoyal:
so while the existing viewer does indeed use an unmodified mathjax, it is going away in calibre 4 anyway. The replacement viewer uses the same mathjax as in-browser viewer which is in resources/content-server/mathjax.zip.xz

So if you really want calibre to use system mathjax you will need to actually replace get_mathjax_manifest() function to use the system mathjax. Which means finding some way of locating the system mathjax and then filtering it to only onclude the same set of files as setup/mathjax.py does.

eli-schwartz:
So I pushed an updated version that is based on the reworked resources/mathjax directory, which takes a mathjax directory that may or may not have the unpacked directory. And another commit adds an option to turn that into a symlink at install time, to some filesystem location.

If you re-open this PR I believe it will automatically detect and show my new commits.

Tested by verifying that the content server and ebook-viewer work on my machine with the symlink to system mathjax. It's assumed the system mathjax directory contains the final directory structure from the manifest.json (having additional files which calibre does not look for, should not cause any issues).
I haven't attempted to do any more complex mappings.

kovidgoyal:
github doesn't let me re-open the PR because the branch was force pushed and re-created, so open a new one.

eli-schwartz:
#921

PR #920, opened 2019-01-14
Checkmodules Edition I

Volkthehunter:
I need a file that controls open or close viewer itself to make it function properly.


kovidgoyal:
Sorry I have no idea what this has to do with calibre.

PR #921, opened 2019-01-18
Add option to build against system mathjax

eli-schwartz:
If unpacked mathjax resources cannot be found, fallback to the packed version.

Allow symlinking a system mathjax directory as `resources/mathjax`.


eli-schwartz:
Switched to using symlinks.

PR #924, opened 2019-01-25
Fixed exception not caught in Startup.py thrown by "pthread_setname_np" on Linux

kruk3t:
I have noticed that Calibre segfaults during startup on my Void Linux distribution running musl libc.

Turns out, that when the "pthread_setname_np" function throws, the exception handler doesn't catch the exception as it expects a type of Exception, instead, it seems to be something else, which I have not determined, as I deem it insignificant.

Catching all exceptions fixes the issue and doesn't have any side effects.


kovidgoyal:
Oh and in case you are not familiar with python, the way to do it would be:

```py
except:
   import traceback
   traceback,print_exc()
```

PR #925, opened 2019-01-26
No-copy books import was implemented

artiomn:
Hello, @kovidgoyal.

## Description

I understand your position about books copying into the Calibre structure:

- Books can be imported from arbitrary source: removable or even non-file.
- It's convenient for administrative purposes to save metadata and data inside common database.

But, please treat following with understanding.

[A lot of people](https://www.mobileread.com/forums/showthread.php?t=272166) want to [add books](https://www.goodreads.com/topic/show/18213711-import-large-amount-of-books-into-calibre-library) without [copying](http://www.mobileread.mobi/forums/showthread.php?t=161275) to the Calibre directory structure.

Me too.

At the first, I have many thousands books stored on the NAS. Directory structure was crafted manually.
Some users and services use it directly.

Calibre is not necessary, but convenient way to grant access for the some users.
I use it with Calibre-Web interface.
But I can't remove current structure and to prefer Calibre database.

Secondly, I don't want to have duplicates. Of course, I can remove them, using filesystem-based deduplication, but it consume memory and CPU time.
Why I need to use it, if I can avoid it?

## Implementation

My patch is look like a crunch, but it works. If parameter `-n` is passed, book file will not be copied, but will be symlinked.
Most of the modern filesystems have symlinks, and I think, it will work on Linux, MacOS and Windows, despite I have tested it only on Linux.

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -8,7 +8,7 @@ __copyright__ = '2011, Kovid Goyal <kovid@kovidgoyal.net>'
 __docformat__ = 'restructuredtext en'
 
 # Imports {{{
-import os, shutil, uuid, json, glob, time, hashlib, errno, sys
+import os, platform, shutil, uuid, json, glob, time, hashlib, errno, sys
 from functools import partial
 
 import apsw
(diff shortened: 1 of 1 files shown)


artiomn:
> As I have explained before I am not willing to have this feature in calibre. There is simply too much prospect of symlinks breaking/books being lost.

I have explained, why this options is useful. This is user's care, how to store his books. Experienced users can make better decision about it, because they know more about them data, than you.

> Even with the current design I regularly have to deal with a few "calibre ate my books" bug reports every week. I absolutely refuse to change the design of calibre to make such bug reports more likely.

I think, experienced users don't send such bug reports, inexperienced will not use console tool.

And this kind of bug report is a sign of incorrect design.

kovidgoyal:
You are welcome to your poinion, but, as I have said, this is not going to happen.

artiomn:
It's a pity. I think this is not very far-sighted of you. This option is useful, and its absence reduces system's scope of applicability.

d-faure:
> 
> 
> It's a pity. I think this is not very far-sighted of you. This option is useful, and its absence reduces system's scope of applicability.

The Calibre structure has to be seen as a black box you shouldn't have to deal with directly, like regular RDBMS do. Storing books in a directory tree is simply an efficient way to take advantage of the underlying OS own file storage features.

Anyway, you could always take the opposite point of view, and develop a plugin which let calibre do it's job when inserting books but once done, remove the original file and replace it with a symlink.

PR #927, opened 2019-02-06
Devendor markdown and feedparser

eli-schwartz:
It would be nice to have markdown update itself with the system, particularly to get access to new extensions. I've stripped out the code copy and updated the current list of extensions in the txt_input plugin to support python-markdown 3.0.

feedparser still has lots of code in it that is broken on python 3, e.g. unicode/unichr types. So I figured we could seize the opportunity to rely on the upstream feedparser, which has native python3 support, indications that there may be a new release soon, and a community of people interested in seeing it improve.

Both modules are inserted into sys.modules in order to guarantee that they can still be imported from where they were historically available -- in fact, I haven't migrated the places which call it, especially feedparser as it is usually called from recipes and that would break old versions.


kovidgoyal:
Simply importing the modules is not nearly enough, to make this work you'd need sys.,meta_path, see startup.py for an example. Not to mention that pre-importing modules is a performance killer.

Also, while I haven't looked at feedparser, markdown 3 seems to have removed various extensions, for example, there is no headerid.py in the source tarball.

eli-schwartz:
> Simply importing the modules is not nearly enough, to make this work you'd need sys.,meta_path, see startup.py for an example. Not to mention that pre-importing modules is a performance killer.

I tested it out and it seemed to work just fine, e.g.
```
$ calibre-debug -c 'import calibre.ebooks.markdown; print(calibre.ebooks.markdown)'
<module 'markdown' from '/usr/lib/python2.7/site-packages/markdown/__init__.py'>
```

However, the pre-importing modules being slow is a good point. As far as that goes, I did not know about PEP 302 import hooks, although I did know about manipulating `sys.modules`. So thank you, both for implementing this PR and for teaching me something new. :D 

> Also, while I haven't looked at feedparser, markdown 3 seems to have removed various extensions, for example, there is no headerid.py in the source tarball.

If you take a close look at my initial implementation, I actually added a check to filter the extensions based on `pkgutil.iter_modules`, so that calibre only tried to use the ones which were available.

This had the dubious advantage of allowing people to use an old version of markdown from before September, but I figured it would be best to allow you to make that decision rather than force it in the PR. Which, well, you did anyway so that is fine.

PR #928, opened 2019-02-11
Time patch

Volkthehunter:
ignore check modules if they appear


kovidgoyal:
what does this have to do with calibre?

Volkthehunter:
This is a code that count times elapsed when reading ebooks.

kovidgoyal:
How does it do that? All I can see is checkmodules.py and time.py neither of which do that

Volkthehunter:
It needed to input times and after it's elapsed it will print something.

Volkthehunter:
checkmodule.py is an accident.

PR #930, opened 2019-02-19
Rework mathjax vendoring and add liberation-fonts as well

eli-schwartz:
Advantages:
- 4MB less binary artifacts in git
- I can build calibre using fonts from /usr/share/fonts/TTF/
- Users get new and presumably improved fonts (Liberation Fonts has seen recent releases, calibre's copies are from 2012 and have seen 4 different versions so far).

In the process, I've switched mathjax to using tar.gz -- I'm not sure what the motivation for using zip was, but Liberation Fonts does not have prebuilt zipped fonts, building them requires fontforge/python-fonttools, and switching to tar results in a mathjax download that is 25MB instead of 35MB, so the whole thing seems like pure win.


kovidgoyal:
The reason I never upgraded the Liberation fonts is because the new ones have been know to cause problems, including kernel leve crashes in windows (blue screens). The ones in calibre I know work well, everywhere, the upstream packages, not so much.

kovidgoyal:
Reference: https://bugs.documentfoundation.org/show_bug.cgi?id=62764

eli-schwartz:
Ouch, that is one issue I completely did not expect.

Should I assume there are other issues with the new fonts, besides for a Windows-specific BSOD? e.g. does it cause unpleasant issues with the Sony PRS? Because in theory I might want to devendor it downstream, but I obviously don't want to break ereaders just to save 4 MB...

kovidgoyal:
I vaguely recall there being other issues, which is why I originally looked up the referenced bug report, but I cant find references to them at the moment. I really dont think it is worth the risk.

eli-schwartz:
Status update:
- I eventually reused my reworked common revendor class in #1074, which was accepted.
- As of commit ebffa0a506f92408a51102f6bb7a669a4872a93d the liberation fonts are now updated to recent versions using `setup.revendor.ReVendor`. Happily, this made it very simple do so *and* simple to make it use the system fonts if you wish. 

Thanks @kovidgoyal!

eli-schwartz:
@norbusan (Debian should use this instead of manually removing it.)

norbusan:
Thanks @eli-schwartz for the heads-up, will put it on our todo list. Much appreciated.

All the best for 2021!

norbusan:
@qykth-git さん、確認をお願いできますか？年明けになってから、僕も確認します。よろしくお願いいたします。よいお年を！

qykth-git:
@norbusan 
Thanks, I will take a look.
ありがとうございます。確認します。

qykth-git:
@eli-schwartz @norbusan
Thanks, I updated Debian package repository as your suggest.

PR #932, opened 2019-02-21
Port bzzdec plugin to build on python2/python3

eli-schwartz:


kovidgoyal:
The format string for PyArg_ParseTuple has to be y# on python3, I'll take care of it when merging

PR #934, opened 2019-02-26
py3: port headless/qt_hack plugins

eli-schwartz:
Plus some general cleanups for plugins I have not entirely touched yet, and another *.py fix during early startup.

### DIFF
--- a/src/calibre/utils/podofo/outline.cpp
+++ b/src/calibre/utils/podofo/outline.cpp
@@ -13,7 +13,7 @@ using namespace pdf;
 static void
 dealloc(PDFOutlineItem* self)
 {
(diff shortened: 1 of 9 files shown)


src/calibre/utils/podofo/outline.cpp
```diff
@@ -13,7 +13,7 @@ using namespace pdf;
 static void
 dealloc(PDFOutlineItem* self)
 {
-    self->ob_type->tp_free((PyObject*)self);
+    py_TYPE(self)->tp_free((PyObject*)self);
```
flaviut: you have a tiny typo here, but otherwise LGTM.

eli-schwartz: Oops.

PR #942, opened 2019-03-13
py3: finish up the rest of the C extension ports

eli-schwartz:
Just some more basic translation to python3 compatible module definitions, but I cannot even test that they compile properly for the Windows/OSX specific ones. Still, they "should" work.

EDIT: now they do


eli-schwartz:
Python 3 has a lzma module, is lzma_binding still needed in this case?

kovidgoyal:
yeah it has a different api to the python module.

PR #944, opened 2019-03-14
py3: add polyglot wrapper for cPickle and use it everywhere

eli-schwartz:
Actually, almost everywhere. src/calibre/utils/config_base.py uses it twice, but both times it shouldn't matter as it is only being used to proxy/load settings files, so they should probably be left alone in order to preserve people's settings.

### DIFF
--- a/src/calibre/db/utils.py
+++ b/src/calibre/db/utils.py
@@ -6,10 +6,11 @@ from __future__ import (unicode_literals, division, absolute_import,
 __license__ = 'GPL v3'
 __copyright__ = '2013, Kovid Goyal <kovid at kovidgoyal.net>'
 
-import os, errno, cPickle, sys, re
+import os, errno, sys, re
 from locale import localeconv
 from collections import OrderedDict, namedtuple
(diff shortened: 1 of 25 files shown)


kovidgoyal:
I dont think it is going to be quite that simple -- unpickling python 2 data on python 3 is not at all straighforward see for instance https://stackoverflow.com/questions/28218466/unpickling-a-python-2-object-with-python-3

kovidgoyal:
pickle is really a bad idea anyway. I'd like to instead excise it from the calibre codebase as much as possible. I've been meaning to do that, but been lazy since there is no overriding need apart from the general bad-ness of pickle.

eli-schwartz:
Ugh.

Haven't thought about data migration at all, really, was just trying to get it to start up. :p Feel free to add a data migration hook or ignore this PR, whichever suits you best.

kovidgoyal:
I have migrated all uses of cPickle except for the ones in the config files, because of the afore-mentioned data portability issues, Will need to investigate how feasible moving the .py and .pickle config files to json is.

eli-schwartz:
Great!

PR #945, opened 2019-03-14
python3: migrate to io.(Bytes|String)IO instead of c?StringIO

eli-schwartz:
This is the only IO representation on python3, and it is backported to python2 to aid in porting but requires determining whether to handle bytes or unicode...  which is sort of the point of porting, so I see no point in using compatibility layers similar to six.moves
(I'm actually quite surprised it seems to be "recommended" to use six.moves to import non-bytes-handling StringIO on python3 and cStringIO on python2, when the whole point of the new module is it no longer handles both str/unicode and bytes.)

In the process of moving over to a modern IO representation, I've updated PyRSS2Gen with the two-line change from upstream, which included the change I was already doing, and bumped the version number to signify this. I've also deleted some code which used cStringIO, but didn't need to exist at all, since there is no user-visible interface to migrating from 2008 era non-filesystem-based ebook databases, and it was a dead code path.


eli-schwartz:
Or alternatively check what it isinstance() of and encode it before passing it into BytesIO()?

As far as safe things go, many of these are used when opening and saving images, when using `open(somefile, 'rb').read()` (actually, why is this copying data into a separate stream inside of a context manager?), or when consuming the output of struct.pack() and etree.tostring().

I've tried to smoketest parts of the code this touched, did not catch any issues yesterday but now I notice src/calibre/ebooks/rtf/rtfml.py is not happy with either io.StringIO or io.BytesIO, because https://github.com/kovidgoyal/calibre/blob/d3ccd4369b7ff9c249559c32daba01c9a40b19c3/src/calibre/ebooks/rtf/rtfml.py#L84-L91 is returning a str(). Given the places where it is used, I believe this should just be calling unicode() every time it is written to an io.StringIO, and I've force-pushed an update to that effect.

kovidgoyal:
Seems like adding isintance() checks or type conversions to ever call site of write() is much more disruptive than simply subclasing BytesIO, like this:

```py
class PolyglotBytesIO(BytesIO):

   def __init__(self, *args, encoding='utf-8'):
       BytesIO.__init__(self, *args)
       self.encoding = encoding
       
   def write(self, x):
        if isinstance(x, unicode_type):
               x = x.encode(self.encoding)
        BytesIO.write(self, x)
```

And similarly for StringIO in the other direction.

eli-schwartz:
Rebased on top of master with only some of the relevant changes for now, essentially for cases where I believe it will be definitely a specific type.

Also for the cases being used in the recipes, I think we want to avoid wrappers anyway. ;)

Will look into the rest at some point. Looking at e.g. the mobi code, it will need to be changed at the calling site regardless... currently there's a lot of cases where it adds strings and bytes together, then writes the combined string to a cStringIO, might as well change it to use b'' strings from the get-go. etc.

PR #946, opened 2019-03-19
python3: misc polyglot fixes

eli-schwartz:
Add more wrappers for urllib, queue.
Remove dead code.

In #941 I overlooked two files that are in metadata/sources/ but they aren't actually metadata plugins and aren't dynamically updated, so it should be safe to use polyglot there.

Question: html2text is currently under active development after it found a new maintainer, is it suitable in its current state to be used directly?


kovidgoyal:
there is a typo in feeds.news (urlplit)

eli-schwartz:
Indeed there is, fixed.

Also turns out src/calibre/devices/manager.py has code that does not seem to be used anywhere? But was trying to `import Queue; Queue(0)`.

kovidgoyal:
I have removed devices/manager.py as it is indeed, unused.

kovidgoyal:
I have also de-vendored html2text.

kovidgoyal:
rebase on master and I will merge.

eli-schwartz:
Done, and thanks.

PR #947, opened 2019-03-21
py3: more fixes, and getting closer to setup.py bootstrap

eli-schwartz:


eli-schwartz:
Ugh. PyQt5 should be using a u'' string but is using default bytes. *grumble*

eli-schwartz:
Thanks, was going to figure out a fix for that but it was too late at night and you beat me to it.

PR #948, opened 2019-03-24
py3: various setup fixes and dict fixes

eli-schwartz:
Includes fixes for recipe/metadata code on the rationale that polyglot.builtins.iter* has existed for a while now, but I can drop those commits if you'd prefer. Where supported, I've flagged the metadata sources as requiring the new version of calibre


kovidgoyal:
I prefer not to rely on polyglot in dynamically loaded code. Either repalce it with the non iter based versions, which might be a slight performance hit on py2 or inline the polyglot definitions you use at the top of the files.

eli-schwartz:
Removed the commit with those changes, for now.

PR #950, opened 2019-03-26
py3: more work towards running the testsuite

eli-schwartz:
It's currently possible to run setup.py test --test-module=build, although further exploration usually results in errors before the test runner can start. I've added a test to check for code that cannot successfully import, incrementally fixed a few issues when running the testsuite, and added some more polyglot wrappers.

Some things to note:
- there is still a lot of cmp use, I only fixed the simple cases where it didn't really make sense to use it on modern python2.
- types.methodType is still being used, and needs a wrapper because on python3 it only takes two arguments. Would this be best in polyglot.builtins or elsewhere?


eli-schwartz:
Sadly the new test does not check whether I committed a new file to git...

eli-schwartz:
Fixed an osx issue, not sure how to debug the failing imports thing on Windows.

eli-schwartz:
Well, that seems to have worked, yay.

PR #953, opened 2019-04-01
py3: misc fixes trying to start calibre.gui_launch

eli-schwartz:


src/calibre/utils/ipc/__init__.py
```diff
@@ -44,7 +44,7 @@ def gui_socket_address():
             if not user:
                 user = os.path.basename(os.path.expanduser('~'))
             if islinux:
-                ADDRESS = (u'\0%s-calibre-gui.socket' % ascii_filename(force_unicode(user))).encode('ascii')
+                ADDRESS = (u'\0%s-calibre-gui.socket' % ascii_filename(force_unicode(user)).decode('ascii'))
```
eli-schwartz: Actually this is wrong as it only works on python3, and the other way only worked on python2 (because the multiprocessing module specifically checks for identity as a `str()` and now instead of being bytes on python3, it is unicode on python2)

eli-schwartz: @kovidgoyal am I correct in thinking that ascii_filename() is intended to return `bytes()`?

kovidgoyal:
Sadly environ items do need to be bytes on windows in python2. This is because python 2 uses the narrow string windows APIs to set environment variables. If the key/value of an environmnet variable is not encodable in whatever locale the windows system is using, then you get an exception. keys are always ascii in calibre so they can be unicode strings. For values, I have created the environ_item() function in polyglot.builtins. Use that to make environ related code polyglot.

eli-schwartz:
> Sadly environ items do need to be bytes on windows in python2.

Ha, fantastic.

Okay -- your implementation of these changes it is, then. Having fewer functions that all seem to do nearly the same thing (sanitize_file_name*) helps too.

> For values, I have created the environ_item() function in polyglot.builtins. Use that to make environ related code polyglot.

Noted!

PR #957, opened 2019-04-10
py3: make install/develop targets work

eli-schwartz:
It's now possible to bootstrap and **install** calibre under python3, even if it doesn't run.

### DIFF
--- a/setup/install.py
+++ b/setup/install.py
@@ -155,7 +155,7 @@ class Develop(Command):
             path = os.path.join(libdir, 'init_calibre.py')
             self.info('Installing calibre environment module: '+path)
             with open(path, 'wb') as f:
-                f.write(HEADER.format(**self.template_args()))
+                f.write(HEADER.format(**self.template_args()).encode('utf-8'))
 
     def install_files(self):
         pass
@@ -189,7 +189,7 @@ class Develop(Command):
         args = self.template_args()
         args['module'] = mod
         args['func'] = func
-        script = template.format(**args)
+        script = template.format(**args).encode('utf-8')
         path = self.j(self.staging_bindir, name)
         if not os.path.exists(self.staging_bindir):
             os.makedirs(self.staging_bindir)
--- a/src/calibre/linux.py
+++ b/src/calibre/linux.py
@@ -800,7 +800,7 @@ class PostInstall:
                     for size in sizes:
                         install_single_icon(iconsrc, basename, size, context, is_last_icon and size is sizes[-1])
 
-                icons = filter(None, [x.strip() for x in '''\
+                icons = list(filter(None, [x.strip() for x in '''\
                     mimetypes/lrf.png application-lrf mimetypes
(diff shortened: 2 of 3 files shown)


eli-schwartz:
@kovidgoyal As the last time I submitted a PR to open files as text instead of adding a decode(), you switched it, I'm assuming you'd prefer the first commit this way.

kovidgoyal:
Yes, I prefer being explicit about encodings in calibre code rather than relying on whatever heuristic python uses to guess encodings. Have been bitten many times in the past when those heuristics have unexpected results.

eli-schwartz:
What is the reason for using env anyway, though? :)

kovidgoyal:
It allows the install scripts to run in more environments. I will make these changes slightly differently, so am closing this PR. In particular in linux.py things should remain unicode strings right upto the point they are written out, rather than starting as bytes.

eli-schwartz:
Okay, feel free to cherry-pick my other two commits then.

kovidgoyal:
I dont understand why the OptionGroup commit is neccessary. There is an add_option_group() method in the OptionParser class in python 3.7 as well and it accepts string arguments:

```
python3 -c "from optparse import OptionParser; p = OptionParser(); g = p.add_option_group('x', 'y')"
```

raises no errors.

eli-schwartz:
I don't really understand what it's doing either, but without it, I get the following error:

```
  File "/home/eschwartz/git/calibre/py3/src/calibre/linux.py", line 554, in write_completion
    o_and_w('calibre-smtp', smtp_op, [])
  File "/home/eschwartz/git/calibre/py3/src/calibre/linux.py", line 538, in o_and_w
    f.write(opts_and_words(*args, **kwargs))
  File "/home/eschwartz/git/calibre/py3/src/calibre/linux.py", line 907, in opts_and_words
    opts  = '|'.join(options(op))
  File "/home/eschwartz/git/calibre/py3/src/calibre/linux.py", line 895, in options
    parser = option_parser()
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/smtp.py", line 192, in option_parser
    _('Options to compose an email. Ignored if text is not specified')).add_option
  File "/home/eschwartz/git/calibre/py3/src/calibre/utils/config.py", line 198, in add_option_group
    return optparse.OptionParser.add_option_group(self, *args, **kwargs)
  File "/usr/lib/python3.7/optparse.py", line 1334, in add_option_group
    raise TypeError("invalid arguments")
TypeError: invalid arguments
```

Elsewhere, calibre creates OptionGroup instances and then adds them, so it would seem to make sense to do the same here.

kovidgoyal:
This should take care of it: https://github.com/kovidgoyal/calibre/commit/d9cd1884af1f8d0e171f66002eca47b46641cd54

kovidgoyal:
I have committed fixes for everything covered in this PR (I think) but I have not tested it on py3.

eli-schwartz:
I've pulled your changes and it works for me. :)

PR #959, opened 2019-04-10
Fix language metadata on device and an option

davidfor:
Kobo changed what format the language is expected to be in the database. Plus I somehow missed saving an option.


kovidgoyal:
Yeah upstream contacted me and adopted (almost) all the patches from
calibre.

PR #968, opened 2019-04-18
py3: make Get Books work

eli-schwartz:


eli-schwartz:
```
hash((self.title, self.author, self.store_name, self.formats))
```

Is this enough granularity? It's possible for stores to have multiple listings of the same book (possibly due to being different ISBN editions, at least I see this commonly enough for paper editions -- my local library also does not do a good job at deduplicating this).

kovidgoyal:
__hash__ has to be consistent with __eq__

PR #970, opened 2019-04-18
py3: make linux file deletion work

eli-schwartz:


eli-schwartz:
BTW it looks like this implementation comes from https://github.com/hsoft/send2trash which has seen some upstream development. It tries to implement fully cross-platform trash with Windows/macOS implementations, unfortunately those use ctypes and the current plat_other.py isn't a standalone file either. Not sure if it makes sense to add an external dependency on that, even for Linux-only.

OTOH the send2trash project now has a giant notice saying it is looking for a new maintainer (and doesn't want to hand it off to "just anyone" as recent NPM security issues show).

kovidgoyal:
Personally i only care about the freedesktop spec implementation -- calibre already has C code is uses for trash operations on other platforms. Indeed on windows it uses a special handler program because microsofts insane move to trash APIs actually spin the event loop.

eli-schwartz:
Well, maybe send2trash would benefit from reliable C code written by someone actively maintaining cross-platform trash APIs. Just a thought. ;)

Meanwhile I will prepare to update the new calibre version. Yay, finally a new release with the refactorings! \o/

kovidgoyal:
Thanks, but I think I maintain quite enough of calibre's dependencies already :)

PR #975, opened 2019-04-24
fix CLI completion for ebook-convert

eli-schwartz:
msgpack.loads will produce mutable, unhashable lists by default when unpacking values that were originally saved as dicts with tuple keys. Which ebook-convert-complete does. This results in errors when
trying to reassemble the dict.

Fix by allowing our msgpack wrapper to forward arguments, and by having ebook-convert-complete unpack lists as tuples.


eli-schwartz:
This is a regression from b9767b2b92a73c87486e3e1378d73d99976b4be8.

eli-schwartz:
:) I initially passed in **kw because I figured it might be handy to be more forgiving in the wrapper. But sure, this works too.

kovidgoyal:
Yeah I prefer to keep things explicit where possible.

PR #976, opened 2019-04-25
py3: make conversion in the GUI work

eli-schwartz:
I suppose my real question, though, is why anything needs to be written to the file at all?


kovidgoyal:
IIRC the conversion result is written to that file.

eli-schwartz:
If this file is supposed to contain the conversion output, then why does it need to be initially created with the contents being the file extension string?

kovidgoyal:
I'm afraid I dont recall why, but doing so is pretty harmless, so...

PR #978, opened 2019-04-29
py3: make completion work

eli-schwartz:
Stop using python2 hack to duplicate python3 print behavior.

### DIFF
--- a/src/calibre/utils/complete.py
+++ b/src/calibre/utils/complete.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python2
 # vim:fileencoding=UTF-8:ts=4:sw=4:sta:et:sts=4:ai
-from __future__ import with_statement
+from __future__ import with_statement, print_function
 
 __license__   = 'GPL v3'
 __copyright__ = '2009, Kovid Goyal <kovid@kovidgoyal.net>'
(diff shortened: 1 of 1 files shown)


eli-schwartz:
While debugging this, I also discovered that ebook-convert-complete.calibre_msgpack has different content when created with python2 or python3 `setup.py resources`. If it was generated with python2, then loading it on python3 results in a dict containing b'' keys (even with raw=False???) and doesn't work. Loading the python3 generated msgpack in a python2 calibre environment results in dictionaries with u'' keys vs. plain old '' keys, which does not matter.

kovidgoyal:
Should be fine now.

On Mon, Apr 29, 2019 at 08:09:19AM -0700, Eli Schwartz wrote:
> While debugging this, I also discovered that ebook-convert-complete.calibre_msgpack has different content when created with python2 or python3 `setup.py resources`. If it was generated with python2, then loading it on python3 results in a dict containing b'' keys (even with raw=False???) and doesn't work. Loading the python3 generated msgpack in a python2 calibre environment results in dictionaries with u'' keys vs. plain old '' keys, which does not matter.
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/978#issuecomment-487618577

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

eli-schwartz:
Almost :) #979

PR #979, opened 2019-04-29
fix typo in previous commit breaking ebook-convert completion on python2

eli-schwartz:
if the content is actually bytes, it needs to be decoded with encoding, not with the string contents


kovidgoyal:
Sorry, you were about 5 seconds too slow :)

PR #982, opened 2019-05-05
Improve conversion performance for edge case ebooks

princesse-framboise:
I've been using Calibre and its ebook conversion feature for a while now and everything worked fine for me until I came upon some epub file (of only 80 pages) that seems to freeze the conversion process. I ended up running the conversion during the night with the command line tool:
```
$ time ebook-convert weird.epub weird.mobi
[... calibre output ...]

real	269m59,303s
user	269m18,286s
sys	0m3,682s
```

After a bit of searching on what Calibre was doing when converting the file, it was spending most of its time creating `Stylizer` objects and, more precisely, generating rules from stylesheets. After taking a look at the html content of the ebook, it seems each word has its own inline style... ending up in generating around a thousand rules each time a stylizer object is created.

I worked on a change to move the rules creation process in a separate class and store the created object in the profile. This object gets reused by the `Stylizer` class if the stylesheets are the same. After the change, the conversion time is way better:
```
$ time ebook-convert weird.epub weird.mobi
[... calibre output ...]

real	9m10,231s
user	9m8,996s
sys	0m0,988s
```

The generated content looks the same (though I'm not sure how I could test if it's the exact same...). I also made some tests with "normal" ebooks and it seems to speed things up a bit:

Before the change:
```
$ time ebook-convert normal.epub normal.mobi
[... calibre output ...]

real	0m29,285s
user	0m29,106s
sys	0m0,176s
```
After the change:
```
$ time ebook-convert normal.epub normal.mobi 
[... calibre output ...]

real	0m24,181s
user	0m24,008s
sys	0m0,160s
```
I also tried running conversions from the GUI and it works fine too.

One possible drawback with this change is the rules are stored in the profiles and those are not destroyed after the conversion process... it may be a problem if the user converts from and to many different profiles in the same session, as the last generated rules for all those profiles will stay in memory. That may be fixed by removing the attributes from the profiles at the end of the conversion process (in `Plumber`).

But before going any further with this change, I wanted to make sure you'd be interested in this kind of changes, as it's kind of an edge case :)


kovidgoyal:
I am definitely interested in patches to improve performance. Sadly I am traveling for the next three weeks so I may not have the time to review this code in the near future. Off the top of my head, why store them in the profile objects? why not store them in the oeb object? This represents the book and so has the same lifetime as the book, which is what we want.

From a quick look at the code, one thing that is wrong is you cannot compare stylesheets as sets, because their order matters. This affects CSS precedence and cascading rules. Only if the order is the same will flattened "rules" be the same.

princesse-framboise:
Thanks for the quick feedback :)

I updated my code following your advices:

- the rules are now stored in the oeb object, it does make much more sense, indeed;
- I added a comparison function to test wether rules should be generated again, it test:
    - the opts object (will always be the same, as I guess it cannot change during a conversion... this is a test on object identity... it might be better to test if properties changed, but again... it's not supposed to change at all),
    - the profile object (this will change, from "input" to "output" during the process, again it's a test on object identity, but I guess profile ojects are immutable anyway)
    - and the stylesheets (and this time, I'm looping over all individual stylesheets).

eli-schwartz:
> The generated content looks the same (though I'm not sure how I could test if it's the exact same...).

From a structural perspective rather than a visual perspective, you could use `calibre-debug --diff`. This will show you how the HTML markup different between the two files (convert to an epub to test).

It would be interesting to see how this affects the markup, especially if it results in fewer rules being generated as that is also less work for e-reader software. If the markup is identical then this change would be very easy to justify as a pure performance speedup.

princesse-framboise:
I did some tests with `calibre-debug` and found a major flaw in my code: stylizer objects were using references to the global "rules" object stored in oeb and each stylizer was using properties from this object... meaning that each stylizer object in the code was using the last generated rules...

I updated it to store pointers to rules in the stylizer object and now, the only difference between ebooks generated before or after my changes are the generated identifiers.

PR #983, opened 2019-05-06
 py3: fix metadata panel showing roman series numbers 

eli-schwartz:
python3 zip() is depleted the first time you use it, but this will need to be used many times.

Noticed while running the python3 version and seeing a series appear as " of XXX", rather than 
"# of XXX".


eli-schwartz:
And should use iterable zip on python2 to avoid waste, but forgot to git add that change. Amended.

PR #985, opened 2019-05-08
Git versions in constants.py

eli-schwartz:
I would like to see a git version when running/installing calibre from git; I've been testing the modification of numeric_version for a long time now without ill effects.

This PR:
- prevents setup.py from breaking when it sees such a version that is manually added
- adds a way to conveniently bump this version without requiring custom hacks

For 99% of people and use cases this has no effect, positive or negative.

### DIFF
--- a/setup/__init__.py
+++ b/setup/__init__.py
@@ -98,8 +98,8 @@ def initialize_constants():
     global __version__, __appname__, modules, functions, basenames, scripts
 
     src = open(os.path.join(SRC, 'calibre/constants.py'), 'rb').read().decode('utf-8')
-    nv = re.search(r'numeric_version\s+=\s+\((\d+), (\d+), (\d+)\)', src)
-    __version__ = '%s.%s.%s'%(nv.group(1), nv.group(2), nv.group(3))
+    nv = eval(re.search(r'numeric_version = (\(.*?\))', src).group(1))
+    __version__   = u'.'.join(map(str if ispy3 else unicode, nv))
     __appname__ = re.search(r'__appname__\s+=\s+(u{0,1})[\'"]([^\'"]+)[\'"]',
             src).group(2)
     epsrc = re.compile(r'entry_points = (\{.*?\})', re.DOTALL).\
--- a/setup/__init__.py
+++ b/setup/__init__.py
@@ -97,13 +97,15 @@ def require_clean_git():
 def initialize_constants():
     global __version__, __appname__, modules, functions, basenames, scripts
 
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I'm fairly sure this will break things in the calibre release pipeline, as the version is assumed to be numeric of the form n.n.n in many places. Furthermore, I dont think it is appropriate to show users git tags/commit offsets in the version number. 

Why do you want to see this in the calibre UI anyway? it would be out of date when developing unless you run the command to update it everytime.

eli-schwartz:
The resulting effect of this command would be to have the version be as follows:
```
numeric_version = (3, 42, 0, 'r48', 'g9bc4d1d375')
```

So it is still numeric for the first three elements, and calibre's internal functions that e.g. check for version updates only compare the first two -- plus, as long as the first three elements compare differently, the fact that there is a level-four string won't weight the version incorrectly, and if they are the same then a string will harmlessly compare as greater than "nothing at all", which to be fair is true.

For the official binary release pipeline, it makes sense to not use git versions at all (because it is not from git). I haven't tested what effect it will have there.

The reason why I want this is for packaging a version of calibre from git: https://aur.archlinux.org/packages/calibre-git/ (It does of course run every time in that case.)
Currently I achieve this by using sed after the install has been run.

As for why I want to see it in the calibre UI, there are many programs that can generate a version string from git. The custom for such projects is that the version string should only use git, when being built from development sources.
I will of course acknowledge that there are, equally, many projects that do *not* generate version strings from git... but I happen to like them, hence I thought I would mention it.

kovidgoyal:
Then I think the better way to do this is to add a new constant named
git_tag or some such and store the git version there. Have the code that
generates human displayable versions use that if not None. Have your git
package run the command to set it. So then people running from the git
package will see the git tag in the version. And there is no risk of
breaking anything. See the get_version() function in constants.py

eli-schwartz:
Hmm, well __version__ is often used in other display areas, but on closer look, it seems like it is not used in the things I was concerned about (--version text, debug tracebacks) which all use get_version() instead. Thanks for the suggestion, I will play with that idea.

kovidgoyal:
get_version() is supposed to be used for all human readable output. If
there are such places where __version__ is used, let me know and I will
fix it.

eli-schwartz:
(Meant to say it is not "directly" used.)

From a quick perusal it looks like the raw __version__ is only used for embedding metadata and sending server version strings and stuff, as you say, get_version should be fine for my purposes.

I'll definitely raise an issue if I see anywhere where that assumption fails.

PR #986, opened 2019-05-09
Bump supported Kobo firmware version

davidfor:
No changes needed except the supported firmware and database versions.


kovidgoyal:
Note that I am travelling fo rthe next two weeks, so it will be a while before I can make a release with these changes.

PR #988, opened 2019-05-12
Staged install: misc fixups

eli-schwartz:
When installing calibre for distro packaging, there are some slightly non-intuitive snags. The primary issues are:
- calibre-uninstall is useless as we don't want to give users the option to circumvent the package manager
- xdg-utils doesn't really have a decent mode for doing staged (`DESTDIR`) installs, thus, some hacks are needed. I've long since implemented these hacks in my packaging, but I suspect others have not... and anyway I would like to drop this code from my packaging, so here is my attempt to merge those actions directly into calibre's build system.

The changes I've implemented here are intended to only run when --prefix is not the same as --staging-root, since we can presume in that case that any user who is not installing directly to their final destination (a live system), is instead installing in a packaging build script. In such cases, we can assume in response to the first concern, that the tool which moves files from the staging root, can also delete them when they are no longer wanted, and in response to the second concern, we do not care what the local system administrator has defined for their custom XDG_DATA_DIRS setup -- we will just install to the standard default for packaging purposes.

### DIFF
--- a/setup/install.py
+++ b/setup/install.py
@@ -214,6 +214,8 @@ class Install(Develop):
             the environment variables:
             XDG_DATA_DIRS=/usr/share equivalent
             XDG_UTILS_INSTALL_MODE=system
+            For staged installs this will be automatically set to:
+            <staging_root>/share
     ''')
     short_description = 'Install calibre from source'
 
--- a/setup/install.py
+++ b/setup/install.py
@@ -135,8 +135,8 @@ class Develop(Command):
         self.opts = opts
         self.regain_privileges()
         self.consolidate_paths()
-        self.write_templates()
         self.install_files()
+        self.write_templates()
         self.run_postinstall()
         self.install_env_module()
         self.success()
--- a/src/calibre/linux.py
+++ b/src/calibre/linux.py
@@ -723,7 +723,8 @@ class PostInstall:
             self.setup_completion()
         if islinux or isbsd:
             self.setup_desktop_integration()
-        self.create_uninstaller()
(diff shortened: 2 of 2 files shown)


eli-schwartz:
Also obligatory mention that XDG utils suck.

eli-schwartz:
One more commit that isn't really in theme with staged installs, but just changes the order of components to be installed.

Motivation: I was looking at what it would take to create co-installable python2/python3 distro packages, with binaries in /usr/lib/calibre/bin-{py2,py3} so that the user could select which version of python they want to use. However, calibre was deleting the newly installed binaries in order to install the code there.

kovidgoyal:
So if calibre-uninstall is not being created at all when using a staging root, why bother with modifying it to also uninstall the mimetypes file from staging?

If we are changing the behavior of the install when prefix == staging_root then it should be documented in the help strings fo those options.

eli-schwartz:
calibre-uninstall should work with the staging sharedir when installed with setup.py install --prefix=/usr, as that will be /usr/share/calibre/calibre-mimetypes.xml

It should work with P('calibre-mimetypes.xml') when installed with `setup.py develop`.

kovidgoyal:
It might be worth just embedding calibre-mimetypes.xml in the uninstaller and writing it to a temp file and call xdg-mime on that. Assuming xdg-mime cares about the file contents and not its path.

eli-schwartz:
Sadly, xdg-mime cares about the path basename, but not the file contents. It also aborts if the path does not exist.

eli-schwartz:
Let's just use a dummy zero-byte file created with the right filename.

PR #990, opened 2019-05-13
Update testsuite for html5-parser 0.4.6 changes

eli-schwartz:
empty `<br />` tags should really have a slash at the end rather than being `<br>`. html5-parser no longer converts the former to the latter as of https://github.com/kovidgoyal/html5-parser/commit/8187842835ca54f78de0a148a02da97df7b6f7e1

In order to pass CI, I believe this will require updating the prebuilt travis sw bundles.


kovidgoyal:
Yeah sadly, I cant update the bundles till I return home, until then, this is going to have to wait.

PR #996, opened 2019-05-20
py3: various conversion fixes

eli-schwartz:
It is now possible to convert to and from most formats. Note that I haven't touched tcr input/output, and rtf2xml is a hot mess (of unmaintained upstream code that has two commits for python3 that don't actually fix any but the most superficial problems).

### DIFF
--- a/resources/templates/html_export_default.tmpl
+++ b/resources/templates/html_export_default.tmpl
@@ -14,16 +14,16 @@ ${head_content}$
   ${for title in meta.titles():}$
     ${if pos1:}$
     <h1>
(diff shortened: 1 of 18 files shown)


eli-schwartz:
Last two commits are a partial attempt to untangle rtf2xml, which should produce the same output on python2, but does not (yet?) work on python3 despite getting past the first couple of error messages. And with that, I'm going to bed. :D

kovidgoyal:
merged with some changes, mostly to fix incorrect additions of r before string literals.

PR #1005, opened 2019-06-02
Greater control over Kobo thumbnails & sleep cover processing

NiLuJe:
* Optional dithering down to the exact eInk color palette.

  Recent FW versions got better at this than they used to,
  but it's still not perfect.
  Note that Mk. 7 devices can and *will* enforce HW dithering,
  at the very least on sleep screens, using an ordered dither pattern,
  like we do.

  Speaking of, said ordered dithering pattern has been ported from ImageMagick.

  This has the added bonus of generating much smaller PNG files.

  Depends on B&W covers to avoid nonsensical results.

* Optional letterboxing of full-screen covers to avoid extra Nickel
  processing.

  Depends on Keep AR to avoid nonsensical results.

* Optional storage as PNG to avoid JPG wrecking the dithering

  Depends on B&W covers to avoid storing stupidly large color PNGs.

* Revamp thumbnail dimension calculations to match what's done with
  store-bought KePubs, fixing a few AR rounding issues in the process.

* Speaking of thumbnails, the target sizes have been unified,
  and the sleep cover dimensions have been updated for a few
  specific models (Aura, H2O, Forma) to match their effective screen resolution,
  instead of something less accurate.

### DIFF
--- a/src/calibre/devices/kobo/driver.py
+++ b/src/calibre/devices/kobo/driver.py
@@ -2637,7 +2637,7 @@ class KOBOTOUCH(KOBO):
                 canvas_size = kobo_size
         return (kobo_size, canvas_size)
 
-    def _create_cover_data(self, cover_data, resize_to, minify_to, kobo_size, upload_grayscale=False, dithered_covers=False, keep_cover_aspect=False, is_full_size=False, letterbox=False, png_covers=False):
+    def _create_cover_data(self, cover_data, resize_to, minify_to, kobo_size, upload_grayscale=False, dithered_covers=False, keep_cover_aspect=False, is_full_size=False, letterbox=False, png_covers=False, quality=90):
         '''
         This will generate the new cover image from the cover in the library. It is a wrapper
         for save_cover_data_to to allow it to be overriden in a subclass. For this reason,
@@ -2655,10 +2655,11 @@ class KOBOTOUCH(KOBO):
                         to smaller thumbnails
         :param letterbox:     True if we were asked to handle the letterboxing
         :param png_covers:    True if we were asked to encode those images in PNG instead of JPG
+        :param quality:       0-100 Output encoding quality (or compression level for PNG, àla IM)
         '''
 
         from calibre.utils.img import save_cover_data_to
-        data = save_cover_data_to(cover_data, resize_to=resize_to, minify_to=minify_to, grayscale=upload_grayscale, eink=dithered_covers, letterbox=letterbox, data_fmt="png" if png_covers else "jpeg")
+        data = save_cover_data_to(cover_data, resize_to=resize_to, compression_quality=quality, minify_to=minify_to, grayscale=upload_grayscale, eink=dithered_covers, letterbox=letterbox, data_fmt="png" if png_covers else "jpeg")
         return data
 
     def _upload_cover(self, path, filename, metadata, filepath, upload_grayscale, dithered_covers=False, keep_cover_aspect=False, letterbox_fs_covers=False, png_covers=False):
@@ -2741,8 +2742,12 @@ class KOBOTOUCH(KOBO):
                              debug_print("KoboTouch:_calculate_kobo_cover_size - expand_to=%s (vs. kobo_size=%s) & resize_to=%s, keep_cover_aspect=%s & letterbox_fs_covers=%s, png_covers=%s" % (
                                  expand_to, kobo_size, resize_to, keep_cover_aspect, letterbox_fs_covers, png_covers))
 
+                        # NOTE: To speed things up, we enforce a lower compression level for png_covers, as the final optipng pass will then select a higher compression level anyway,
+                        #       so the compression level from that first pass is irrelevant, and only takes up precious time ;).
+                        quality = 10 if png_covers else 90
+
                         # Return the data resized and properly grayscaled/dithered/letterboxed if requested
-                        data = self._create_cover_data(cover_data, resize_to, expand_to, kobo_size, upload_grayscale, dithered_covers, keep_cover_aspect, is_full_size, letterbox, png_covers)
+                        data = self._create_cover_data(cover_data, resize_to, expand_to, kobo_size, upload_grayscale, dithered_covers, keep_cover_aspect, is_full_size, letterbox, png_covers, quality)
 
                         # NOTE: If we're writing a PNG file, go through a quick optipng pass to make sure it's encoded properly, as Qt doesn't afford us enough control to do it right...
                         #       Unfortunately, optipng doesn't support reading pipes, so this gets a bit clunky as we have go through a temporary file...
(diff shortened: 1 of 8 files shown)


NiLuJe:
This is the latest iteration of the patch I've been using for, err, quite a while, which was originally posted [over here on MR](https://www.mobileread.com/forums/showthread.php?p=3025725#post3025725).

Pinging @davidfor ;).

NiLuJe:
This probably ought to be squashed, I didn't do it on my end mainly to document the latest changes compared to last year's iteration ;).

kovidgoyal:
If @davidfor signs off on it I will be happy to merge.

src/calibre/devices/kobo/driver.py
```diff
@@ -2587,41 +2610,62 @@ def images_path(self, path, imageId=None):
...
         if keep_cover_aspect:
-            library_aspect = library_size[0] / library_size[1]
-            kobo_aspect = kobo_size[0] / kobo_size[1]
-            if library_aspect > kobo_aspect:
-                kobo_size = (kobo_size[0], int(kobo_size[0] / library_aspect))
+            # NOTE: Py3k wouldn't need explicit casts to return a float
+            # NOTE: Unlike Qt, we round to avoid accumulating errors,
+            #       as ImageOps will then floor via fit_image
+            aspect_ratio = library_size[0] / float(library_size[1])
+            rescaled_width = int(round(kobo_size[1] * aspect_ratio))
```
eli-schwartz: We use `__future__.division` here, so it should not be necessary to convert to a float merely to preserve precision. Also rescaled_width is only used to compare against other numbers for truthiness, so it shouldn't be necessary to use both int() and round().

davidfor:
Firstly, the changes to the configuration are wrong. The order of the options in the KOBO driver is important. Changing them means that the new options will get the existing values for the old options they move. If you are changing options for the KOBO driver, they configuration should be rebuilt like in the KOBOTOUCH driver.

And for the KOBOTOUCH driver, adding the old-style options is not needed. These are only there to allow migration of old stored options to the new style. As these are new, they will not be in the old-style configuration file for someone who last made changes with calibre v1 (and yes, there are). And again, with the way these have been done, it will break the options when they are migrated.

davidfor:
Have the changes to the cover generation been tested on an original Kobo or a WiFI? If not, it probably shouldn't be changed in the KOBO driver. The letterboxing is probably the only option that should be added and that is being done for all sizes, not just the full screen size.

And what about the Mini? Strictly, this is stuck on firmware 3.19.5761. Are any of these changes going to cause problems with it? Or other devices using older firmware?

davidfor:
The only real change I'm not happy with is not matching the image sizes with what Kobo generates for sideloaded books. My general attitude with the driver has been to replicate what Kobo would do for sideloaded and only add things that it cannot do otherwise. I'm fine with generating better images that use less space, but, not using their sizes is a different matter. I need to take some time to test this part and see what happens.

src/calibre/devices/kobo/driver.py
NiLuJe: Good point, I'd missed that, which explains why no-one ever complained that the AR handling was broken, and why I was seeing different results when I tested that logic in a Python shell ;). I'll kill the casts, thanks.

`rescaled_width` *can* be returned as, well, the width in some circumstances, and round() returns a float, so, AFAICT, it does need both.

NiLuJe:
@davidfor:

* Option handling was done in a fairly dumb "search & update" approach 4 years ago, and I never looked into the various quirks of the system. Il'l see if I can figure out how to clear that up, thanks for the heads up ;).

* Not tested on legacy Kobos, no, to the best of my knowledge. Probably ought to rip out the KOBO bits, then (again, that's mostly a result of the original search & replace approach ;)).

* No issues on older FW, no, I started using this over 4 years ago, so, that was on something like 3.8 at the time or something ;). The only question mark is PNG support, and that's reflected in the help message. Although, thinking about it, I doubt that could actually ever be an issue, even on really old Qt4 FW.

* Yeah, for thumbnails, going with the store-bought KePub behavior (vs. what Nickel does to sideloaded content) was a conscious decision. It *feels* like the better approach, but I get your point ;). In this case, none of the approaches actually match the effective size used/available in the final widget where they're used, so I don't have a strong attachment to the final decision.

Replicating the sideloading behavior instead is as simple as always setting expand to False ;).
```diff
diff --git a/src/calibre/devices/kobo/driver.py b/src/calibre/devices/kobo/driver.py
index 1a59f08690..a98470faf7 100644
--- a/src/calibre/devices/kobo/driver.py
+++ b/src/calibre/devices/kobo/driver.py
@@ -2737,7 +2737,7 @@ class KOBOTOUCH(KOBO):
                         #       (i.e., QSize's boundedTo() vs. expandedTo(). See also IM's '^' geometry token, for the same "expand" behavior.)
                         #       Note that Nickel itself will generate bounded thumbnails, while it will download expanded thumbnails for store-bought KePubs...
                         #       We chose to emulate the KePub behavior.
-                        resize_to, expand_to = self._calculate_kobo_cover_size(library_cover_size, kobo_size, not is_full_size, keep_cover_aspect, letterbox)
+                        resize_to, expand_to = self._calculate_kobo_cover_size(library_cover_size, kobo_size, False, keep_cover_aspect, letterbox)
                         if show_debug:
                              debug_print("KoboTouch:_calculate_kobo_cover_size - expand_to=%s (vs. kobo_size=%s) & resize_to=%s, keep_cover_aspect=%s & letterbox_fs_covers=%s, png_covers=%s" % (
                                  expand_to, kobo_size, resize_to, keep_cover_aspect, letterbox_fs_covers, png_covers))
```

* The only case where I object to Nickel's behavior is for the full cover, because not using the effective screen resolution is just plain wrong. That was one of original issue that drove me to work on this, and back when Nickel wasn't doing any dithering, doing it the "right way" did ensure bit-exact results.

NiLuJe:
Okay, should be in a better shape now ;).

davidfor:
I have done some testing over the weekend and cannot find any problems. I haven't built the libraries, so haven't tested anything that needs them . But, overall, I'm happy with the results. 

The only thing I'd want changed is the order of the options. As the letterbox option is associated with the keep aspect ratio option, it should be after it, or on the same line. Same with the options for black and white covers.

Other than that, I'm happy for the changes to be merged.

NiLuJe:
Good point, I forgot about that when I reintroduced the dithering checkbox in December ;).

Before -> After shots:

Cal_Driver

I also updated the [KoboUtilities](https://www.mobileread.com/forums/showthread.php?p=3025725#post3025725) patch to match that:

Cal_KU

kovidgoyal:
Looks generally OK to me. One question, why are you dividing compression_quality by 10 in save_cover_data_to?

NiLuJe:
That's for img_to_data's png compression argument, which is separate, and
expects a different range (0-10), so instead of adding yet another knob, I
just piggyback on the jpeg quality level, vaguely like ImageMagick does.

PR #1006, opened 2019-06-05
Support subtitle in Douban metadata plugin

lifanxi:
Get subtitile from Douban API and merge it to title.
This feature is optional and is disabled by default.

### DIFF
--- a/src/calibre/ebooks/metadata/sources/douban.py
+++ b/src/calibre/ebooks/metadata/sources/douban.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python2
 # vim:fileencoding=UTF-8:ts=4:sw=4:sta:et:sts=4:ai
+
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 __license__   = 'GPL v3'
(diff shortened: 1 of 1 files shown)


src/calibre/ebooks/metadata/sources/douban.py
```diff
@@ -44,112 +45,6 @@ def get_details(browser, url, timeout):  # {{{
 # }}}
 
 
-def to_metadata(browser, log, entry_, timeout):  # {{{
```
lifanxi: I have moved this function into Douban class, because I need to get access to the `options` variable in Douban class.

src/calibre/ebooks/metadata/sources/douban.py
```diff
@@ -170,6 +65,121 @@ class Douban(Source):
...
+                    if ans and ans.strip():
+                        return ans.strip()
+            except:
+                log.exception('Programming error:')
+            return None
+
+        id_url = entry_id(entry_)[0].text.replace('http://', 'https://')
+        douban_id = id_url.split('/')[-1]
+        title_ = ': '.join([x.text for x in title(entry_)]).strip()
+        subtitle = ': '.join([x.text for x in subtitle(entry_)]).strip()
```
lifanxi: Here is the real changes: to include subtitile in title, if the option is enabled.

kovidgoyal:
You can just pass the options object to the to_metadata function, that
way the diff is much smaller.

lifanxi:
Sure I can do that. I just can't see the benefit of keeping the to_metadata function outside the Douban class. So I moved it in.

PR #1007, opened 2019-06-12
Better layout of new cover options in KoboTouch driver.

davidfor:
The order of the existing options should have been maintained. Plus centering one of the didn't look good,


NiLuJe:
Yeah, I was on the fence about the centering myself ;).

(FWIW, I reordered them so that they ended up in the same order as the GUI, because I kept bungling it up by not actually reading the names and assuming that was already the case ^^).

NiLuJe:
(Which means you can probably pull the Qt import I added for the centering constant, too).

PR #1010, opened 2019-06-21
py3: more __future__ work and cleanups along the way

eli-schwartz:


src/calibre/ebooks/metadata/ereader.py
```diff
@@ -48,7 +49,7 @@ def get_metadata(stream, extract_cover=True):
             try:
                 mdata = pheader.section_data(hr.metadata_offset)
 
-                mdata = mdata.split('\x00')
+                mdata = mdata.decode('utf-8').split('\x00')
```
eli-schwartz: I'm actually not sure what this should be handled as. Does this particular format use some terrible legacy encoding? :D

kovidgoyal: I have no idea, one would have to ask @user-none

kovidgoyal: Actually, looking at reader132.py which is where HeaderRecord comes from, I'd guess the appropriate encoding to use is cp1252

user-none: cp1252 is the correct encoding the format uses. While there is an encoding entry in the header the format "spec" says only cp1252 can be used. FYI, https://github.com/kovidgoyal/calibre/blob/master/format_docs/pdb/ereader.txt has the format's text encoding documented.

kovidgoyal: @user-none thanks for the confirmation, I should have remembered about format_docs

PR #1011, opened 2019-06-24
tests: remove last reference to sgmllib

keszybz:
Uses of sgmllib were removed in 692230147c79a1072e37005f91e5664835d53967.

### DIFF
--- a/src/calibre/test_build.py
+++ b/src/calibre/test_build.py
@@ -297,11 +297,6 @@ class BuildTest(unittest.TestCase):
 
     def test_feedparser(self):
         from calibre.web.feeds.feedparser import parse
-        # sgmllib is needed for feedparser parsing malformed feeds
-        # on python3 you can get it by taking it from python2 stdlib and
-        # running 2to3 on it
-        import sgmllib
-        sgmllib, parse
 
     def test_openssl(self):
         import ssl


src/calibre/test_build.py
```diff
@@ -297,11 +297,6 @@ def test_markdown(self):
 
     def test_feedparser(self):
         from calibre.web.feeds.feedparser import parse
-        # sgmllib is needed for feedparser parsing malformed feeds
```
eli-schwartz: I don't see how beautifulsoup 3 to bs4 has anything to do with this line. calibre still very much needs sgmllib.

The test was added in https://github.com/kovidgoyal/calibre/commit/d436e2d7565d1f7ea7c97e0020b16f347e2800e7 and the commit immediately before was when the vendored feedparser (which depends on sgmllib) was removed in-tree.

This test is called test_feedparser, not test_beautifulsoup, and without it we do not know whether feedparser._SGML_AVAILABLE is true so that news recipes will be robust.

keszybz: The test should test something that used by calibre directly, e.g. some functionality of feedparser. Testing for the presence of a third-party module that is not used directly makes assumptions about feedparser dependencies.

eli-schwartz: Okay, so what is your point? Your rationale for removing this is still incorrect, because:
- you're removing something that the author explicitly wanted there, without rationalizing _why_ in a way that answers the initial intent of the test
- your commit message is wrong, it makes an assertion that you are removing the "last references to sgmllib", but the commit you're pointing to never used `sgmllib` in the first place -- it used `calibre.ebooks.sgmllib`

If you feel the test could be better, by testing the existence of functionality rather than assumed indirect dependencies, then maybe you could write a replacement test that tries to parse a malformed feed rather than just removing the import test. :)

kovidgoyal:
This test is there because feedparser still uses sgmllib and since sgmllib is not present in python3, it is there to remind me to make sure I include it in the calibre binaries. As long as feedparser continues to use sgmllib, this test should remain. Changing it to a functionality based test just means that test failures will be harder to diagnose.

The correct fix is to port feedparser to not use it.

PR #1016, opened 2019-07-05
py3: various cleanups and more __future__ work

eli-schwartz:
See last commit -- I'm not sure whether this is the correct solution, but it seems plausible.


kovidgoyal:
yeah it is bytes for binary files, etree objects for html, stylesheet objcets for css and so on.

PR #1017, opened 2019-07-19
py3: fix invalid escapes

keszybz:
python3.8 warns about strings which try to escape characters which do
not need that. To avoid the warning, the backslash should be escaped.
(Behaviour is functionally the same, so e.g. '\$' is still the same as
r'\$', except for the warning.)


eli-schwartz:
Thanks, what do you think about changing this to raw strings instead, though? I've been touching up various places that use backslash-escaped sequences to use raw strings if possible, as they're easier to read.

keszybz:
I started that way, but there were some strings which were already using escaping backslashes and apostrophes this way, so I used `\\` to be consistent. In this case, I don't think r'' is going to be more readable, because there's just too many strings. I can change it though, if you prefer.

eli-schwartz:
Hmm, in the case of a giant list of very small strings you're right that it's not especially readable either way.

It's not very important either way, just something I've been doing as a matter of habit while going through files and switching to unicode_literals. e.g. 67f8b23baa89c6ebb2ccad2f481a418fb09cdcb8 6ca1739d6beb9987e6cebc41bd5070b2c2ef1759

PR #1020, opened 2019-07-23
py3: more future work and assorted cleanups along the way

eli-schwartz:
Of note: unicode_check tried to get me to edit pyuic-generated files, changing things there would be a bit of a recipe for disaster if we actually thought it had issues... get this back into sync with setup/check.py


eli-schwartz:
Regarding `int(foo/2)` vs. `foo//2`, shouldn't floor division always return an int if the dividend is an int, which I believe was the case in all those places? (Yes, it's a bikesheddy style thing and int() certainly isn't *wrong*. I think it's supposed to be very very slightly slower than floor division though.)

kovidgoyal:
It wasn't clear to me at all that the dividend would always be an int,
in such situations I prefer correctness over performance.

eli-schwartz:
Fair enough!

PR #1027, opened 2019-08-09
Support Dublin Core id tags when importing HTML

cszucko:
All of the below formats are supported and would be interpreted as "foo:bar"
```
<meta name="DC.identifier" scheme="foo" content="bar" />
<meta name="dc:identifier.foo" content="bar"/>
<meta name="DCTERMS:identifier" scheme="foo" content="bar" />
<meta name="dcterms.identifier.foo" content="bar" />
```

Refactoring the existing meta and comment tag parsing to use the same `HTMLParser` subclass would make it straightforward to only parse the HTML once and reduce the reliance on regular expressions but I didn't want to go rewriting things before making sure this PR would be accepted.

### DIFF
--- a/src/calibre/ebooks/metadata/html.py
+++ b/src/calibre/ebooks/metadata/html.py
@@ -10,6 +10,8 @@ Try to read metadata from an HTML file.
 
 import re
 
+from HTMLParser import HTMLParser
+
 from calibre.ebooks.metadata import string_to_authors
 from calibre.ebooks.metadata.book.base import Metadata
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Sure I have no problem with adding support for identifiers to HTML metadata

kovidgoyal:
Go ahead and do the re-write I'll review and merge in one go.

cszucko:
Rewrite complete.  If there are any issues just let me know.

PR #1030, opened 2019-08-17
Biblia Tysiaclecia recipe

Calibresque:
Works with EPUB format, reads biblia.deon.pl website


kovidgoyal:
I'm afraid I cant accept recipes that work only with a specific output format into calibre.

PR #1031, opened 2019-08-19
py3: misc fixes and cleanups

eli-schwartz:


src/calibre/gui2/main.py
```diff
@@ -400,11 +396,11 @@ def run_gui(opts, args, listener, app, gui_debug=None):
         if getattr(runner.main, 'debug_on_restart', False) or gui_debug is not None:
             run_in_debug_mode()
         else:
-            import subprocess
+            import shlex, subprocess
             if hasattr(sys, 'frameworks_dir'):
                 app = os.path.dirname(os.path.dirname(os.path.realpath(sys.frameworks_dir)))
                 prints('Restarting with:', app)
-                subprocess.Popen('sleep 3s; open ' + shellquote(app), shell=True)
+                subprocess.Popen('sleep 3s; open ' + shlex.quote(app), shell=True)
```
keszybz: That's probably some old debugging stuff left over. I'd remove the sleep.
Then this should simply be: `subprocess.Popen(app)`.
(But this patch is OK. This can be simplified later.)

src/calibre/gui2/catalog/catalog_epub_mobi.py
```diff
@@ -806,7 +806,7 @@ def preset_save(self):
             self.preset_field.blockSignals(True)
             self.preset_field.clear()
             self.preset_field.addItem('')
-            self.preset_field_values = sorted([q for q in self.presets], key=sort_key)
+            self.preset_field_values = sorted((q for q in self.presets), key=sort_key)
```
keszybz: Is this loop needed at all? `sorted(self.presets, key=sort_key)` should be the same. (Above and below too.)

src/calibre/gui2/main.py
kovidgoyal: shlex.quote is not available in python 2. And the sleep is done via the shell not time.sleep() so that it does not block the main thread. Although, since this is macOS specific, one could use preexec_fn instead.

kovidgoyal: On the other hand, given the warnings about preexec_fn deadlocking, better to just leave it to use the shell

src/calibre/gui2/catalog/catalog_epub_mobi.py
eli-schwartz: I must have been tired when I hunted down these additional instances...

src/calibre/gui2/main.py
eli-schwartz: Commit dropped...

(using run_calibre_debug seems like a fine solution)

PR #1032, opened 2019-08-20
tests: skip unrar test gracefully if module is missing

keszybz:
Fedora (and probably some other distributions) cannot include unrardll
because of licensing reasons. Unfortuantely this situation is unlikely
to change in the near future. Skip the test if the module is not
available.

### DIFF
--- a/src/calibre/test_build.py
+++ b/src/calibre/test_build.py
@@ -17,6 +17,10 @@ from polyglot.builtins import iteritems, map, unicode_type, getenv, native_strin
 
 is_ci = os.environ.get('CI', '').lower() == 'true'
 
+try:
+    import unrardll
+except ModuleNotFoundError:
+    unrardll = None
 
 class BuildTest(unittest.TestCase):
 
@@ -236,6 +240,7 @@ class BuildTest(unittest.TestCase):
         from calibre.gui2.win_file_dialogs import test
         test()
 
+    @unittest.skipUnless(unrardll, 'Module unrardll is missing')
     def test_unrar(self):
         from calibre.utils.unrar import test_basic
         test_basic()


kovidgoyal:
I'm not willing to skip that test unconditionally, because some distributions have problems shipping unrardll. The main purpose of that test is to ensure that unrardll is correctly built. Skipping it like that would mean that is for whatever reason it is not built, then it wont detect that. 

You can instead set either a command line argument or an env var to skip that test.

eli-schwartz:
Don't those distros also patch calibre to remove references to the cbr former for conversion? I'd prefer such a patch (env/cli option to skip unrardll checks) to come with functionality to obsolete that patch at the same time. You can check if unrardll exists before showing the conversion option. Or set a single, easily sed'able global that can be checked both by conversion code and the build tests.

As it is, I'd argue the patch which removes GUI features should also be removing the test which checks that the GUI feature works.

keszybz:
> Don't those distros also patch calibre to remove references to the cbr former for conversion?

No, at least in Fedora. Users might install the missing bits separately, e.g. from rpmfusion or pip or whatever, and then things will work for them. An error that gives a hint to the user that something is missing is better than removing the functionality altogether.

kovidgoyal:
There is a good deal of functionality that uses unrar. Support for extracting books from RAR archives, conversion from both CBR and RAR files. And this functionality is used by command line tools, the server, the GUI. I am really not interested in implementing friendly error reporting in all those cases to cater to a handful of linux distributions, but patches are welcome.

keszybz:
No, the existing error message is good enough, and I'm not asking for it to be changed. I don't remember the exact details, but a window pops up with a traceback that makes it pretty clear that the module cannot be imported.

PR #1037, opened 2019-09-08
py3: misc cleanup work, even more __future__s

eli-schwartz:


kovidgoyal:
I have merged dropping the https commit since that code no longer exists in engine anyway and engine will become master soon enough.

eli-schwartz:
0kay, sounds reasonable :)

PR #1038, opened 2019-09-10
more py3 fixes and cleanups

eli-schwartz:


eli-schwartz:
So uh apparently sqlite3 needs some native_string_types.

PR #1043, opened 2019-09-13
debug: return the fully qualified path to calibre-debug

eli-schwartz:
We need to use sys.executables_location, because there may be multiple calibre-debug programs in the path or we may simply not be in the path at all. I ran into this due to testing multiple installations of calibre for python2/python3, and discovering that run_calibre_debug used the wrong python version.


eli-schwartz:
Thanks.

I'm curious, though... I'm not sure I understand what else was missing here... I thought that calibre should only be available via a platform freeze (frameworks_dir or isfrozen) which is always checked first, and alternatively by the setup.py develop/install launchers -- which set executables_location. (And run-local, but that already did not work). Is there some other method too?

kovidgoyal:
Yeah you can technically build and run calibre from source code on
platforms other than linux, though I dont think anyone does so, other
than the CI jobs on Azure and the calibre release pipeline.

PR #1044, opened 2019-09-17
Implement setMetadata for odt files - 1841506

pieplu:
Fix: https://bugs.launchpad.net/calibre/+bug/1841506

### DIFF
--- a/src/calibre/ebooks/metadata/odt.py
+++ b/src/calibre/ebooks/metadata/odt.py
@@ -20,17 +20,18 @@
 #
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-import zipfile, re, io
+import zipfile, re, io, os
 import xml.sax.saxutils
 
 from odf.namespaces import OFFICENS, DCNS, METANS
 from odf.opendocument import load as odLoad
 from odf.draw import Image as odImage, Frame as odFrame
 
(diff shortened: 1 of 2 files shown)


src/calibre/ebooks/metadata/odt.py
```diff
@@ -220,6 +224,28 @@ def get_metadata(stream, extract_cover=True):
...
+        metaFields[fields.get('language')] = lang_as_iso639_1(l) or l
+    return metaFields
+
+def set_metadata(stream, mi):
+    from calibre.utils.zipfile import safe_replace
+    metaFields = get_meta_doc_props(mi)
+
+    zin, odfs = get_odf_meta_parsed(stream, addfields=metaFields, deletefields=metaFields)
+    stream.seek(os.SEEK_SET)
+    safe_replace(stream, "meta.xml", io.BytesIO(str(odfs.meta())))
```
eli-schwartz: As part of the effort to port to python3, we always use unicode_type from polyglot.builtins -- never str(). In this case it reveals a bug -- you cannot use a python3 str() when writing to a BytesIO buffer.

Instead, since meta() returns a MetaCollector.content() (which is a unicode_literals string value), you should be using .encode('utf-8'). Using str() to cast to bytes only on python2 is wrong.

pieplu: Ok, thanks, I fix-it!

PR #1046, opened 2019-09-17
remove lines that incorrectly raises an error

jn8029:
I have a paid WSJ subscription. Upon using this recipe, I found that the if condition as removed in this PR incorrectly raises an error, while the login attempt was successful. After removing this condition, the recipe works fine.

### DIFF
--- a/recipes/wsj.recipe
+++ b/recipes/wsj.recipe
@@ -148,9 +148,9 @@ class WSJ(BasicNewsRecipe):
             self.log('Performing login callback...')
             res = br.submit()
             self.wsj_itp_page = raw = res.read()
-            if b'>Sign Out<' not in raw:
-                raise ValueError(
-                    'Failed to login (callback URL failed), check username and password')
+            #if b'>Sign Out<' not in raw:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Can you figure out what the correct if condition for you is instead? Saw raw somewhere like this: open('/path/to/some/file.html', 'wb').write(raw)  and inspect that file to check why Sign Out is not present and what is present instead.

eli-schwartz:
This removes the check for successful logins. I assume the text ">Sign Out<" no longer appears in the returned HTML, you should check for a new heuristic to replace it instead.

jn8029:
Just checked. "Sign Out" is not in the html because of i18n. I guess they checked the request ip and does i18n accordingly.  登出 (sign out in Chinese) is present instead.

jn8029:
This is the tag. Maybe use the href url for the if function instead?
```
<a href="https://accounts.wsj.com/logout?target=">登出</a>
```

jn8029:
#1048

PR #1050, opened 2019-09-24
Avoid error on python3 about str and bytes

zazaho:
When I launch the calibre application build for python3 I get two errors about expecting bytes not str and the other way around.
in src/calibre/devices/mtp/driver.py and src/calibre/devices/mtp/unix/driver.py
I found that theses edits remove the error.


eli-schwartz:
Why do you have half of your changes in one PR and half in the other PR?

zazaho:
Hi Eli, sorry about that. I made a mistake when adding the patches. They
should go together.

PR #1055, opened 2019-10-07
Speed-up of the FB2 writer and other fixes

PalmtopTiger:
Hi,

Thank you for adding the [support for PNG images](https://bugs.launchpad.net/calibre/+bug/1844911), it speeds up the conversion.

But even after these changes the conversion of one average book takes roughly one hour. I looked into the code and found that most of the time spent on splitting of the base64 string. Python strings are immutable and appending just one char to a string leads to copying the whole string on each iteration. The better way to use the slice method.
The old code converts one image with size of 820 Kb more than 11 minutes, the new one – faster than one second of time.

Code | Seconds
--- | ---
old | 680.590
new | 0.014

[split_test.zip](https://github.com/kovidgoyal/calibre/files/3696952/split_test.zip)

So now the conversion of one book takes only several seconds.
We can observe this terrible slowdown only on Python 2, the Python 3 concatenate strings much faster, but it’s still not the best way to solve this problem.
I should note, that the [FB2 documentation](http://www.fictionbook.org/index.php/Элемент_binary) says, that the base64 string should be splitted by 72 chars, but Calibre splits it at 72th char (i.e. splits by 71 chars). I fixed this too.
This is the first commit.

Let’s get going further. After adding the PNG support there appeared an issue, when the ID of PNG image contained ".jpg" extension.
`<binary id="_0.jpg" content-type="image/png">`
I did not find a way to get the correct extension, because the link to the cover image adds up before the images was converted, so we cannot know if the cover image was converted to another format or not. So I just removed extension from ID, because it’s just a identifier, and file extension there is not required. I have these kind of books in my library and they work just fine. This is the second commit for.

The next commits are just prettify the document as well as making the file smaller. I strive to adjust document to the same way, as the official FictionBook Editor does.

Last three commits fix the issues with empty tags in the body (sometimes writer generates something like `<p><strong></strong></p>`) and annotation tag, which should contain paragraph tag.

In the end, I have a question: can I reduce the indentation level of the header to position it on the same level as the body?

Best regards,
Andrey.

### DIFF
--- a/src/calibre/ebooks/fb2/fb2ml.py
+++ b/src/calibre/ebooks/fb2/fb2ml.py
@@ -103,6 +103,10 @@ class FB2MLizer(object):
         if self.opts.insert_blank_line:
             text = re.sub(r'(?miu)</p>', '</p><empty-line/>', text)
 
+        # Put line breaks between paragraphs on a separate line.
+        text = re.sub(r'(?miu)</(p|title)>\s*<empty-line/>', r'</\1>\n<empty-line/>', text)
+        text = re.sub(r'(?miu)<empty-line/>\s*<p>', '<empty-line/>\n<p>', text)
+
         return text
 
     def fb2_header(self):
--- a/src/calibre/ebooks/fb2/fb2ml.py
+++ b/src/calibre/ebooks/fb2/fb2ml.py
@@ -19,7 +19,7 @@ from calibre.constants import __appname__, __version__
 from calibre.utils.localization import lang_as_iso639_1
 from calibre.utils.img import save_cover_data_to
 from calibre.ebooks.oeb.base import urlnormalize
-from polyglot.builtins import unicode_type, string_or_bytes
+from polyglot.builtins import unicode_type, string_or_bytes, range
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'll review this when I have some time, kinda busy with calibre 4 things right now. Not sure what you are refering to in your question? You mean reduce indentation of the tags in the generated XML?

PalmtopTiger:
Yes. The book header has indentation of 4 spaces, the body - 0 spaces (because it's hard to keep the indentation level when we write XML as plain text). Because of this, the final document looks unsightly.

kovidgoyal:
On Mon, Oct 07, 2019 at 05:00:10AM -0700, Andrey Efremov wrote:
> Yes. The book header has indentation of 4 spaces, the body - 0 spaces (because it's hard to keep the indentation level when we write XML as plain text). Because of this, the final document looks unsightly.

Sure, feel free, I have no objections.

PalmtopTiger:
Thanks. I also simplified regular expressions a bit.

PR #1056, opened 2019-10-13
Allow a user to specify the version that is installed

michaelwittig:
I was looking for a way to install a specific version of calibre instead of latest.

I'm not sure about the `check_version` function. It doesn't seem to be used and I'm not sure what it does.


kovidgoyal:
As is noted on the download page, you can just manually download an untar the tarball for any release to install it. However, adding this capability to the installer is useful.

PR #1064, opened 2019-10-29
Add render page tweak settings.

JimmXinu:
A couple tweak settings to allow customization of page size and margins for page renders of ebooks.  This arose from [this discussion](https://www.mobileread.com/forums/showthread.php?p=3909276&postcount=3780).

I will admit that I don't know where else this render code might be called from and may have named and described the settings poorly.

### DIFF
--- a/resources/default_tweaks.py
+++ b/resources/default_tweaks.py
@@ -566,3 +566,14 @@ exclude_fields_on_paste = []
 # Useful if for some reason your operating systems network checking
 # facilities are not reliable (for example NetworkManager on Linux).
 skip_network_check = False
+
+#: Render Page Settings
+# Default page size and margins to use when rendering an HTML page.
+# Affects size of cover rendered from ebook file.
+# Examples:
+#   render_page_size = 'A4'
+#   render_page_size = 'A6'
+#   render_page_size = 'Letter'
+#   render_page_margins = (left, top, right, bottom)
+render_page_size = 'A4'
+render_page_margins = (0, 0, 0, 0)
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I'm not sure I understand the use case. Why is having calibre generate
larger covers which is generally a good thing, bad in this case? Is it
because when sending to a kobo the generated cover is used and becomes
illegible? Wouldn't it be better to simply have a nice image cover or
even a plain text one that has less text on it and is thus legible
everywhere? After all at even 800x600 or whatever size, if it is
thumbnailed to a small size it will still not be legible. 

In any case, if you want to fix this, the best way would be to
have some metadata in the HTML that tells calibre what size to render 
the cover at. Something like

<meta name="calibre-html-render-size" content='{"size": "9x11", "margins": [0, 0, 0, 0]}'>

That way the person generating the cover can decide what is the
best way to render that cover.

JimmXinu:
Somebody noticed that their rendered covers from FanFicFare title pages weren't readable anymore.  The larger image using the same size font results in smaller text when scaled to a smaller size in Calibre's cover view, making it less readable.  

I'm just trying to get it back to roughly the relative font size it was before.

Is `<meta name="calibre-html-render-size" content='{"size": "9x11", "margins": [0, 0, 0, 0]}'>` an existing feature?

JimmXinu:
Sorry, I should have seen the commit link--my mistake.

I'm not sure how GitHub reports comments to you, so sorry if you get this twice.  There's a problem with that commit src/calibre/ebooks/render_html.py:78.  Missing quotes around 'margins' as dict key.
`margins = QMarginsF(*data['margins'])`

Thanks, I'll look into this feature (which does work for me when I added 'margins') and vw vh units.

kovidgoyal:
Thanks fixed

PR #1065, opened 2019-10-30
fix loading of bash completion

eli-schwartz:
And other misc install fixups in the process.

### DIFF
--- a/setup/install.py
+++ b/setup/install.py
@@ -147,8 +147,8 @@ class Develop(Command):
         self.consolidate_paths()
         self.install_files()
         self.write_templates()
-        self.run_postinstall()
         self.install_env_module()
+        self.run_postinstall()
         self.success()
 
     def install_env_module(self):
@@ -164,6 +164,7 @@ class Develop(Command):
             self.info('Installing calibre environment module: '+path)
             with open(path, 'wb') as f:
                 f.write(HEADER.format(**self.template_args()).encode('utf-8'))
+            self.manifest.append(path)
 
     def install_files(self):
         pass
--- a/src/calibre/linux.py
+++ b/src/calibre/linux.py
@@ -548,17 +548,16 @@ def write_completion(bash_comp_dest, zsh):
         complete = os.path.join(getattr(sys, 'frozen_path'), complete)
 
     with open(bash_comp_dest or os.devnull, 'wb') as f:
-        w = polyglot_write(f)
 
         def o_and_e(*args, **kwargs):
-            w(opts_and_exts(*args, **kwargs))
+            f.write(opts_and_exts(*args, **kwargs))
             zsh.opts_and_exts(*args, **kwargs)
 
         def o_and_w(*args, **kwargs):
-            w(opts_and_words(*args, **kwargs))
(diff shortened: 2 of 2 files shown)


kovidgoyal:
Doesn't this mean that installing the next calibre version will orphan the old completions file? It should probably be deleted by the install process. Or better just continue to use it, no complications that way. Also I'd rather not convert the desktop file templates into bytestrings. In general, the principle I follow is convert to bytes as close to output as possible and be unicode everywhere else.

eli-schwartz:
The old completions file is /usr/share/bash-completion/completions/calibre, and that's also **one of** the new completion files. So it should be fine.

On the topic of converting bytestrings, it is raw file data and the only thing it ever accomplishes is be raw file data... the first and only time it is used after being defined is in f.write(), so it seemed simpler and saves an encode() to make the desktop templates be handled the same way similar objects are handled in:

```
setup/linux-installer.py:CACERT = b'''\
src/calibre/ebooks/conversion/plumber.py:DEBUG_README=b'''
```
I mean, if you want me to remove it, I can just switch it to use `f.write(VIEWER.encode('utf-8'))` etc.

kovidgoyal:
On Wed, Oct 30, 2019 at 04:51:17AM -0700, Eli Schwartz wrote:
> The old completions file is /usr/share/bash-completion/completions/calibre, and that's also **one of** the new completion files. So it should be fine.

OK.

> 
> On the topic of converting bytestrings, it is raw file data and the only thing it ever accomplishes is be raw file data... the first and only time it is used after being defined is in f.write(), so it seemed simpler and saves an encode() to make the desktop templates be handled the same way similar objects are handled in:

That's true now, but suppose in the future we want to manipulate it
before writing? Will work on py2 and fail on p3. Encoding at th elast
minute works everywhere and is robust against future changes to the
code.

> 
> ```
> setup/linux-installer.py:CACERT = b'''\
> src/calibre/ebooks/conversion/plumber.py:DEBUG_README=b'''
> ```
> I mean, if you want me to remove it, I can just switch it to use `f.write(VIEWER.encode('utf-8'))` etc.

I suggest using as_bytes() instead, if you dont like encode()

eli-schwartz:
> That's true now, but suppose in the future we want to manipulate it before writing? Will work on py2 and fail on p3. Encoding at th elast minute works everywhere and is robust against future changes to the code.

idk, it felt pretty unlikely that a desktop file would ever be modified. They aren't "configurable", really, and the only time in git history that they were ever modified (as far as I can tell) was before 2009 when the "Version" key (a static string defining which version of the Desktop Entry Specification it conforms to) was bugfixed in commit 51c08484ef254c8256116e024f16c50db46e4ca6 to *not* be modified with the version of calibre.

The mimetypes have always been written separately.

But if you would prefer not to take chances, then it can always start life as unicode and be immediately encoded. It would be consistently defensive against refactoring (albeit a micro-deoptimization) to change those other locations too...

> I suggest using as_bytes() instead, if you dont like encode()

I think encode() is fine unless you have a particular reason to think you don't know what the original type is, like using native_string_type. In the long run, I think http://mypy-lang.org/ will do better static type checking than as_bytes() anyway.

PR #1066, opened 2019-10-31
Possible fix for #1850764.

cbhaley:
Assuming I guessed right on the column icon rule, the problem is that Polish doesn't clear the icon caches (or any other cache). I fixed this by adding a call to refresh_ids
after line 532. My problem is that I don't know what to do with the resulting lines 534-536. Do they need to stay? They are updating the book details pane. As far as I can tell, refresh_ids does not do that update, so the lines should stay. Do you agree?

### DIFF
--- a/src/calibre/gui2/actions/polish.py
+++ b/src/calibre/gui2/actions/polish.py
@@ -530,6 +530,7 @@ class PolishAction(InterfaceAction):
             pass
         self.gui.tags_view.recount()
         if self.gui.current_view() is self.gui.library_view:
+            self.gui.library_view.model().refresh_ids((book_id,))
             current = self.gui.library_view.currentIndex()
             if current.isValid():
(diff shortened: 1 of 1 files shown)


cbhaley:
Kovid, is the failure on Arch something for me? The log seems to be mosting git stuff.

eli-schwartz:
```
ERROR: test_workers (calibre.srv.tests.loop.LoopTest)
Test worker semantics
  File "/__w/calibre/calibre/src/calibre/srv/tests/loop.py", line 99, in test_workers
    res.status, res.reason, res.getheaders(), res.read()))
Exception: Got unexpected response: code: 408 Request Timeout headers: [('Content-Length', '0'), ('Content-Type', 'text/plain; charset=UTF-8'), ('Date', 'Thu, 31 Oct 2019 18:29:27 GMT'), ('Connection', 'close')] data: b''
```

Well, it's a timeout, so it's unlikely it matters here.

PR #1071, opened 2019-11-26
Fix Czech local time handling in Denik N recipe

mihalyr:
This is a follow up patch for https://bugs.launchpad.net/calibre/+bug/1853788 
I realized on Tuesday (Úterý) that there is an encoding issue with the title string that is using `calibre.strftime` which tries to decode the `time.strftime` output (python2) with a detected preferred encoding in my case `utf-8`. In the Deník N recipe there is a locale switch to include Czech date in the title string which is not `utf-8` and results in invalid chars (�) in the title string.

So I could just revert back to using `time.strftime().decode('iso-8859-2')` or change `calibre.strftime` to accept a character set for decoding. I chose the latter here, but don't have strong feelings if you prefer to keep the change in the recipe.

Tested on Linux and sent to Kindle, show today's date correctly as "Deník N - Út, 26 lis 2019" instead of the previous "Deník N - �t, 26 lis 2019".

I wanted to add a test case for the `calibre.strftime` function for this and to avoid regression etc., but I am not familiar with the codebase and the testing approach and couldn't find the right place for it. Please advise if you would like to me to add some tests too.

### DIFF
--- a/recipes/denikn.cz.recipe
+++ b/recipes/denikn.cz.recipe
@@ -12,15 +12,31 @@ def cz_title_time():
     """
 
     orig_locale = locale.getlocale(locale.LC_TIME)
+
+    # Try to switch time locale to Czech and return the formatted current date
     try:
         locale.setlocale(locale.LC_TIME, 'cs_CZ')
+        # In Python 2 time.stftime returns czech chars in ISO 8859-2 encoding
+        # e.g. the character Ú is returned as the bytes 'da'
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I dont think is this a good idea. in general, switching locales in the middle of a multithreaded program is not a good idea, it can screw up lots of things as it is a global variable. 

It should be trivial to implement a function to return localized date in czech. Just needs an array of day and month names.

PR #1072, opened 2019-11-26
Fix Czech local time handling in Denik N recipe

mihalyr:
This is a follow up patch for https://bugs.launchpad.net/calibre/+bug/1853788
I realized on Tuesday (Úterý) that there is an encoding issue with the title string that is using calibre.strftime which tries to decode the time.strftime output (python2) with a detected preferred encoding in my case utf-8. In the Deník N recipe there is a locale switch to include Czech date in the title string which is not utf-8 and results in invalid chars (�) in the title string.

This is a fix for the above based on the suggestion in https://github.com/kovidgoyal/calibre/pull/1071

Using strftime requires locale change which is not thread safe.
Keeping it simple and using a hardcoded list of day and month names,
this should work the same way on python 2 and 3 as well.

### DIFF
--- a/recipes/denikn.cz.recipe
+++ b/recipes/denikn.cz.recipe
@@ -2,25 +2,26 @@
 # vim:fileencoding=UTF-8:ts=4:sw=4:sta:et:sts=4:ai
 from __future__ import unicode_literals, division, absolute_import, print_function
 from calibre.web.feeds.news import BasicNewsRecipe
-from calibre import strftime
-import locale
+from datetime import datetime
+
+
+CZ_DAYS = ['Po', 'Út', 'St', 'Čt', 'Pá', 'So', 'Ne']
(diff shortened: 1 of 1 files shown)


mihalyr:
I made a mistake with month indexing, will send the fix in a second

eli-schwartz:
Python doesn't have a uselocale implementation that could be used in a thread, but I suppose you could still use multiprocessing for this.

IIRC there's also a pypi module which provides a thread-safe uselocale implementation but that isn't very helpful in the here and now.

mihalyr:
Yeah, I don't think it would be worth adding more complexity here for the sake of a recipe. Day and month names are not really changing and the code is not reused in any other context, also should work under python 3 too.

mihalyr:
Actually, I could perfectly live without the czech dates too and just keep using whatever the default locale is and use strftime, but thought it is nicer to have czech dates in a czech newspaper title...
But either way, this patch or removing czech dates, one should be applied as the current impl changes locales which is not good...

PR #1074, opened 2019-12-03
setup: add option to use offline hyphenation dicts

eli-schwartz:
Motivation: I'd like to be able to point the hyphenation dicts at a cached tarball for non-network builds. So I figured I could reuse the mathjax code, and I'd previously attempted to refactor this so I had a bit of handy progress.

I've also taken a stab at making the download code not clean and retry it if it already exists: https://github.com/eli-schwartz/calibre/commit/627955f7d2224059bab192837f64a397fa1b1315
But I'm not sure about this since the similar mathjax code checks the version in the mathjax manifest. I think it would be quite nice to do this somehow, though, because currently every execution of setup.py resources will redownload this tarball. (But with this PR, you can make it use --path-to-hyphenation instead.)

P.S. Why do these need to be stored compressed, and then extracted into the cachedir? Why not just store them directly unpacked, and use them from resources/ ?

P.P.S. Have you considered the possibility of using the ones which distributions install to /usr/share/hyphen/ ?

### DIFF
--- a/setup/hyphenation.py
+++ b/setup/hyphenation.py
@@ -10,11 +10,8 @@ import os
 import shutil
 import tarfile
 from io import BytesIO
-from zipfile import ZipFile
 
-from setup import Command, download_securely
-
-URL = 'https://github.com/LibreOffice/dictionaries/archive/master.zip'
+from setup.revendor import ReVendor
 
 
 def locales_from_dicts(dicts):
@@ -83,30 +80,19 @@ def compress_tar(buf, outf):
         compress(buf, outf)
 
 
-class Hyphenation(Command):
+class Hyphenation(ReVendor):
 
     description = 'Download the hyphenation dictionaries'
-
-    def add_options(self, parser):
-        pass
-        # parser.add_option('--path-to-mathjax', help='Path to the MathJax source code')
-
-    @property
-    def hyphenation_dir(self):
-        return self.j(self.RESOURCES, 'hyphenation')
-
-    def clean(self):
-        if os.path.exists(self.hyphenation_dir):
-            shutil.rmtree(self.hyphenation_dir)
+    NAME = 'hyphenation'
+    TAR_NAME = 'hyphenation dictionaries'
+    VERSION = 'master'
+    DOWNLOAD_URL = 'https://github.com/LibreOffice/dictionaries/archive/%s.tar.gz' % VERSION
 
     def run(self, opts):
         self.clean()
-        os.makedirs(self.hyphenation_dir)
-        self.info('Downloading hyphenation dictionaries...')
-        with self.temp_dir() as src, ZipFile(BytesIO(download_securely(URL))) as zf, self.temp_dir() as output_dir:
-            zf.extractall(src)
-            if len(os.listdir(src)) == 1:
-                src = os.path.join(src, os.listdir(src)[0])
+        os.makedirs(self.vendored_dir)
+        with self.temp_dir() as dl_src, self.temp_dir() as output_dir:
+            src = opts.path_to_hyphenation or self.download_vendor_release(dl_src, opts.hyphenation_url)
             process_dictionaries(src, output_dir)
             dics = sorted(x for x in os.listdir(output_dir) if x.endswith('.dic'))
             m = hashlib.sha1()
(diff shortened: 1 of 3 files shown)


kovidgoyal:
On Mon, Dec 02, 2019 at 06:43:15PM -0800, Eli Schwartz wrote:
> P.S. Why do these need to be stored compressed, and then extracted into the cachedir? Why not just store them directly unpacked, and use them from resources/ ?

Because that reduces storage requirements for a feature that is not
commonly used. Not to mention that for dmg and msi distributions,
compression is much worse than that achieved by a .tar.xz

> 
> P.P.S. Have you considered the possibility of using the ones which distributions install to /usr/share/hyphen/ ?

No, I've never understood how that is supposed to work. What if the user
is hyphenating a book that has a language other than their primary one?
Hyphenation is a background process, there is no opportunity to tell the
user to go install whatever distro specific hyphenation package the book
requires. It's kind of absurd to expect users to a-priori install
dictionaries. And given that compressed the hyphenation dicts are under
1MB I am fine with storing/distributing them.

> 
>   https://github.com/kovidgoyal/calibre/pull/1074
> 
> -- Commit Summary --
> 
>   * setup: refactor mathjax class into generic class for vendoring things
>   * setup: port over the hyphenation dictionary code to ReVendor()
> 
> -- File Changes --
> 
>     M setup/hyphenation.py (38)
>     M setup/mathjax.py (71)
>     A setup/revendor.py (65)
> 
> -- Patch Links --
> 
> https://github.com/kovidgoyal/calibre/pull/1074.patch
> https://github.com/kovidgoyal/calibre/pull/1074.diff
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/1074

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

kovidgoyal:
On Mon, Dec 02, 2019 at 06:43:15PM -0800, Eli Schwartz wrote:
> I&#39;ve also taken a stab at making the download code not clean and retry it if it already exists: https://github.com/eli-schwartz/calibre/commit/627955f7d2224059bab192837f64a397fa1b1315
> But I&#39;m not sure about this since the similar mathjax code checks the version in the mathjax manifest. I think it would be quite nice to do this somehow, though, because currently every execution of setup.py resources will redownload this tarball. (But with this PR, you can make it use --path-to-hyphenation instead.)

If you want to do this, do the best way would be to store the ETag
header from the HTTP response. Then use If-Modified-Since when
re-downloading.

kovidgoyal:
On Mon, Dec 02, 2019 at 06:43:15PM -0800, Eli Schwartz wrote:
> I&#39;ve also taken a stab at making the download code not clean and retry it if it already exists: https://github.com/eli-schwartz/calibre/commit/627955f7d2224059bab192837f64a397fa1b1315
> But I&#39;m not sure about this since the similar mathjax code checks the version in the mathjax manifest. I think it would be quite nice to do this somehow, though, because currently every execution of setup.py resources will redownload this tarball. (But with this PR, you can make it use --path-to-hyphenation instead.)

Store the ETag in the ./.build-cache directory.

eli-schwartz:
Well, I tried it but AFAICT github is simply ignoring the If-None-Match header. I've attempted to manually compare etags, see this: https://github.com/eli-schwartz/calibre/commit/885deaede4311793ebcad56f3354086ddddfda6f

Seems to work, although it feels like this shouldn't all be necessary. :(

kovidgoyal:
Well I guess that is what you have to do. Might be easier to instead use
the github api to get the commit hash of HEAD and use that to key the
cache instead.

On Mon, Dec 02, 2019 at 08:50:51PM -0800, Eli Schwartz wrote:
> Well, I tried it but AFAICT github is simply ignoring the If-None-Match header. I've attempted to manually compare etags, see this: https://github.com/eli-schwartz/calibre/commit/885deaede4311793ebcad56f3354086ddddfda6f
> 
> Seems to work, although it feels like this shouldn't all be necessary. :(
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/1074#issuecomment-560997285

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

PR #1082, opened 2019-12-24
Improved smooth scrolling and add auto-scroll feature

mgziminsky:
In the past, I had used a simple AutoHotkey script I wrote to perform smooth auto scrolling, but at some point a change to the scrolling logic or the web engine broke the script and it doesn't seem to be fixable. So instead, I decided to add the ability directly into calibre.

The current smooth scroll impl is jumpy and inconsistent, ie: scrolling up N times then down N times may not result in the same start and end positions. The lack of configuration also makes it difficult and impractical to implement an external solution as I did before.

This PR adds 2 new scroll settings, "smooth scroll speed" and "auto scroll speed", along with a new smooth scrolling implementation that supports auto-scrolling. Auto scroll can be toggled via a keyboard shortcut and its speed increased/decreased via keyboard shortcuts as well. Normal smooth scroll speed is only configurable via the preferences.

I have tested on win 10 in both the native viewer and from a chrome browser connected to the content server. The changes are exclusively in rapydscript, so unless I am misunderstanding something, I would expect consistent behavior on other platforms.

I didn't include it, but I also considered adding a preference to make the mouse wheel perform smooth scrolling as well. It should be simple to add later though if there is interest or if I get bored.


src/pyj/read_book/settings.pyj
```diff
@@ -4,23 +4,24 @@ from __python__ import hash_literals
...
-    opts.bg_image_fade = settings.bg_image_fade or 'transparent'
-    opts.paged_wheel_scrolls_by_screen = v'!!settings.paged_wheel_scrolls_by_screen'
+    opts.hide_tooltips = settings.hide_tooltips
     opts.is_dark_theme = v'!!settings.is_dark_theme'
-    opts.override_book_colors = settings.override_book_colors or 'never'
+    opts.margin_left = max(0, settings.margin_left)
+    opts.margin_right = max(0, settings.margin_right)
+    opts.override_book_colors = settings.override_book_colors
+    opts.paged_wheel_scrolls_by_screen = v'!!settings.paged_wheel_scrolls_by_screen'
+    opts.user_stylesheet = settings.user_stylesheet
```
kovidgoyal: the original code defends against invalid values being set on the passed in object, not just against no settings. I'd rather not lose that.

src/pyj/read_book/flow_mode.pyj
```diff
@@ -10,6 +10,9 @@ from read_book.viewport import scroll_viewport
 from utils import document_height, viewport_to_document
 
 
+def line_height():
+    return parseFloat(line_height.doc_style.lineHeight)
```
kovidgoyal: are u sure computed style line height is always a pixel value on all browsers?

mgziminsky: It should be if they follow spec according to this:
- https://developer.mozilla.org/en-US/docs/Web/CSS/computed_value
- https://drafts.csswg.org/css2/visudet.html#propdef-line-height

> The computation needed to reach a property's computed value typically involves converting relative values (such as those in em units or percentages) to absolute values. For example, if an element has specified values font-size: 16px and padding-top: 2em, then the computed value of padding-top is 32px (double the font size).

After reading the spec though, there is a special case value "normal" that will be shown as that, in which they recommend using 1 to 1.2 * fontSize. So I will add a fallback for that case.

src/pyj/read_book/settings.pyj
mgziminsky: I may be misunderstanding, but the previous code only guarded against values missing from settings with `or`, not invalid values. All the value checks (min/max, etc.) are still present.

I can improve it a bit though by merging the passed in settings with the defaults to ensure there are no missing properties. It would then handle the missing value case as before.

kovidgoyal:
This has changed the behavior where scrolling with the down key pressed stops at file boundaries. If you want that to happen, then it needs to be controllable by an option defaulting to the current behavior.

kovidgoyal:
You should also add a button to the toolbar in viewer/toolbars.py to toggle auto-scrolling mode. Probably leave it off by default.

mgziminsky:
Somehow I didn't realize that was the current behavior. While skimming through the code, I mistook the timing check in `_check_for_scroll_end` as meaning don't turn the page if the last scroll occurred within 2 seconds. Upon a closer look, I see that's not the case.

So in that case, I will add an additional option to control that behavior. For this new option, which of these do you think would be best?
- Single option that only affects continuous scrolling via hotkey, not auto-scroll
- 2 options, one for continuous scrolling, one for auto-scroll
- Single option that affects both continuous scrolling and auto scrolling

I can't imagine wanting auto-scroll to automatically turn the page, what if you haven't finished reading yet..., so my first choice would be the first or second option, but I'm also biased and not actually a heavy user of the application.

kovidgoyal:
And in the case of auto-scrolling you cant really have it stop at file boundaries, but maybe have a longer pause at file boundaries.

kovidgoyal:
I think the option should apply only to key based scrolling. For auto-scrolling have a pause time at file boundaries which can be set to zero to disable.

mgziminsky:
I considered adding a toolbar button, but wasn't sure where your icons come from and i'm no artist so didn't want to attempt making my own from scratch. I figured that could be added later.

For the options then, a configurable pause for the auto-scroll sounds good. So i'll add a checkbox for the key based scrolling and a delay/pause option for the auto-scroll.

kovidgoyal:
Sounds fine.

PR #1087, opened 2020-01-15
fix: chrome not support `atom+xml`

bluelovers:
chrome and moon+ reader not support Content-Type `atom+xml` only support `xml`

### DIFF
--- a/src/calibre/srv/opds.py
+++ b/src/calibre/srv/opds.py
@@ -35,7 +35,7 @@ from polyglot.binary import as_hex_unicode, from_hex_unicode
 
 
 def atom(ctx, rd, endpoint, output):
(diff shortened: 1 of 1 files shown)


eli-schwartz:
But that's the wrong mimetype, why produce an inferior solution for working software? Maybe moon+ reader should be fixed.

bluelovers:
but in chrome it also not display as xml, it display as source code

and ....

sry no way feedback to moon+ reader author xd

kovidgoyal:
chrome has no support for atom, since they dropped it. OPDS feeds are
atom feeds, they must have the mimetype atom+xml

bluelovers:
nvm, i made this 2 for handle this

https://github.com/bluelovers/ws-calibre/tree/master/packages/calibre-server

https://github.com/bluelovers/ws-calibre/tree/master/packages/calibre-proxy

PR #1089, opened 2020-01-16
py3 fixes for bibtex file writer

norbusan:
Hi Kovid,

here a simple fix for bibtex exporting removing the `encoding(...)` after the `format`. The file is already opened with
```
        with codecs.open(path_to_output, 'w', bibfile_enc, bibfile_enctag)\
            as outfile:
```
and `bibfile_enc` can be set via the command line.

I am not sure whether in case of errors
```
            else :
                log.warn("Incorrect --choose-encoding flag, revert to default")
                bibfile_enc = bibfile_enc[0]
```
the encoding should be reset to `bibfile_enc[0]` or to `preferred_encoding`.


eli-schwartz:
Shouldn't this target the master branch, which the py3 branch is periodically rebased onto? The py3 branch mostly has changes which *only* work on py3, e.g. it's simpler for handling Windows support it seems.

eli-schwartz:
The actual code it's changing doesn't care about the open mode of the bibtex file. The problem is that calibre.utils.date.now().strftime("%A, %d. %B %Y %H:%M") formats a datetime.datetime object and returns a str, which naturally cannot be decoded. (So it was always wrong to decode it, ever since its initial implementation in commit 4fb3a73d85abd314c30a6e94adc514d7939d7985.)

norbusan:
Hi @eli-schwartz 
thanks for the comments, I wasn't clear about the distinction between `py3` branch and `master` branch.The original poster (debian bug #949095) reported that with the py3 version it stopped working, and confirmed that with the fix it is ok again.

eli-schwartz:
Sure, if you run the master branch on python3 you get the same issue -- my calibre builds generate both python2 and python3 support from the same source tree, which means I don't want to use the py3 branch as it started removing python2 compatibility, e.g. removing the custom lzma module now that python3 has a builtin one with a different API.

norbusan:
Indeed, indeed. In Debian they try to purge Python2 ASAP so we only have Py3 now. So should be targeting master. Should I redo the PR? I guess not needed, too trivial a change.

PR #1091, opened 2020-01-21
fix sip loading when not provided from Qt

norbusan:


eli-schwartz:
Hmm, good catch, since every other call site falls back on the generic sip. :)

But sip is really confusing me, I must admit. There's the generic version (Arch python-sip), one built for pyqt5 (python-sip-pyqt5), and currently now python-pyqt5 itself build-depends on sip5 and provides the PyQt5.sip module directly, so it's moving everywhere...

Are you using sip-build to make pyqt5 yet? I think perhaps that is when PyQt5.sip is internal to the base PyQt5 build.

kovidgoyal:
As I understand it, one is supposed to use the PyQt5.sip preferentially
for PyQt5. If that is not present, fallback to the generic sip.

On Tue, Jan 21, 2020 at 07:21:37AM -0800, Eli Schwartz wrote:
> Hmm, good catch, since every other call site falls back on the generic sip. :)
> 
> But sip is really confusing me, I must admit. There's the generic version (Arch python-sip), one built for pyqt5 (python-sip-pyqt5), and currently now python-pyqt5 itself build-depends on sip5 and provides the PyQt5.sip module directly, so it's moving everywhere...
> 
> Are you using sip-build to make pyqt5 yet? I think perhaps that is when PyQt5.sip is internal to the base PyQt5 build.
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/1091#issuecomment-576730941

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

norbusan:
Thanks for merging. It was due to a failing test on Debian where there is no Qt sip I found it out. Thanks to @eli-schwartz for reminding me to run the tests

eli-schwartz:
Yay for the tests catching their first (I think?) upstream bug. :)

PR #1095, opened 2020-02-02
Security:replace PyCrypto with PyCryptodome

jnozsc:
Background:

`pycrypto` is unmaintained with a lot of vulnerability issues

details:
https://github.com/dlitz/pycrypto/issues/173
https://github.com/dlitz/pycrypto/issues/176
https://github.com/dlitz/pycrypto/issues/269
https://github.com/dlitz/pycrypto/issues/285
https://github.com/dlitz/pycrypto/issues/301

Solution:

`pycryptodome` https://github.com/Legrandin/pycryptodome is a fork of `pycrypto`, actively maintained, 100% compatible interface,  no code change required

### DIFF
--- a/bypy/sources.json
+++ b/bypy/sources.json
@@ -691,10 +691,10 @@
     },
 
     {
-        "name": "pycrypto",
+        "name": "pycryptodome",
         "unix": {
-            "filename": "pycrypto-2.6.1.tar.gz",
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Why is pycrypto being built at all, since it is not actually used?

kovidgoyal:
the py3 branch already uses pycryptodome so this will happen
automatically when that is merged into master.

@eli: As I have said before calibre ships with pycrypto for use by third
party plugins several of which started relying on it on ye olden days.

PR #1100, opened 2020-02-10
Fix series sort after using clear_series

Sophist-UK:
If you have books with series and series_index > 1, and use bulk metadata to clear_series, then sort by series puts these at the bottom of the list because the series_index has not been reset to 1.


Sophist-UK:
Additionally, after using clear_series, the old series number is used by default when you add a new series name but do not specify an index.

PR #1101, opened 2020-02-15
Fix touch lag when changing chapter in flow mode

Siron777:
Hi,

With this branch I fix the lag when scrolling with a touch screen in flow mode at the beginning of a chapter:

-  After changing of chapter by scrolling, the following start events are not cancelable and trigger the error "**Ignored attempt to cancel a touchstart event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.**". I guess there is a bug in Chromium with the scrollTo fonction called just before the touch. It somehow creates lags. To fix it, the only solution I found is to add the **touch-action:none** to the style of the body. It does not seem to break any touch interaction in the viewer. 
- I also found that the touch end event is not triggered when changing of chapter. This is probably the bug [https://bugs.chromium.org/p/chromium/issues/detail?id=464579](https://bugs.chromium.org/p/chromium/issues/detail?id=464579). This is exacerbating the previous issue (seem to generate more not cancelable event). Another consequence is that the timer loop to detect the hold gesture is then never stop (up to the next touch event). I add a function to reset it from the iframe.js. I am not a big fan of that change but did not found a more elegant fix. And I prefer to address it instead of waiting for a fix in Chromium which is not coming since 5 years.

Simon

### DIFF
--- a/resources/content-server/reset.css
+++ b/resources/content-server/reset.css
@@ -34,6 +34,7 @@ time, mark, audio, video {
 
 body {
     line-height:1.2;
+    touch-action:none;
 }
 
 article,aside,details,figcaption,figure,
--- a/src/pyj/read_book/iframe.pyj
+++ b/src/pyj/read_book/iframe.pyj
@@ -44,7 +44,8 @@ from read_book.shortcuts import (
     create_shortcut_map, keyevent_as_shortcut, shortcut_for_key_event
 )
 from read_book.toc import update_visible_toc_anchors
-from read_book.touch import create_handlers as create_touch_handlers
+from read_book.touch import (create_handlers as create_touch_handlers,
+                             reset_handlers as reset_touch_handlers)
 from read_book.viewport import scroll_viewport
 from utils import (
     apply_cloned_selection, clone_selection, debounce, html_escape, is_ios
@@ -334,6 +335,8 @@ class IframeBoss:
         self.send_message('content_loaded', progress_frac=self.calculate_progress_frac(), file_progress_frac=progress_frac())
         self.last_cfi = None
         self.auto_scroll_action('resume')
+        reset_touch_handlers() # Needed to mitigate issue https://bugs.chromium.org/p/chromium/issues/detail?id=464579
+
         window.setTimeout(self.update_cfi, 0)
         window.setTimeout(self.update_toc_position, 0)
 
--- a/src/pyj/read_book/touch.pyj
+++ b/src/pyj/read_book/touch.pyj
@@ -140,6 +140,12 @@ class TouchHandler:
                 return True
         return False
 
+    def reset_handlers(self):
+        self.stop_hold_timer()
+        self.ongoing_touches = {}
+        self.gesture_id = None
+        self.handled_tap_hold = False
+
     def start_hold_timer(self):
         self.stop_hold_timer()
         self.hold_timer = window.setTimeout(self.check_for_hold, 100)
@@ -199,9 +205,7 @@ class TouchHandler:
         self.prune_expired_touches()
         if not self.has_active_touches:
             self.dispatch_gesture()
-            self.ongoing_touches = {}
-            self.gesture_id = None
-            self.handled_tap_hold = False
+            self.reset_handlers()
 
     def handle_touchcancel(self, ev):
         ev.preventDefault(), ev.stopPropagation()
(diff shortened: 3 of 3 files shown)


kovidgoyal:
Dont add properties to reset.css this is used everywhere. Instead in touch.pyj in create_handlers do
document.body.style.touchAction = 'none' and add a comment explaining why it is necessary.

Siron777:
I replied to the email of github, but I am not sure you go it, and as I can see, this reply is not visible in this thread. 

As I replied to your previous comment in the mail:
_That was my first approach. Unfortunately it does not work, it seems to be necessary that this properties is set on the main document, not on the document of iframe. And I cannot set it on the main document from the touch.pyj, because of security violation. I need an access to the main document, but then it is away from the touch.pyj and the normal css._

So we need to move the configuration of the properties from the iframe document to the main page ? Where ? In the view.pyj ? Or maybe I pass the main document from the view.pyj as a parameter of the create_handlers ?

kovidgoyal:
Setting it on the main document means you cant use touch gestures such as pinch to zoom  etc anywhere in the UI. That is pretty unfortunate. For the standalone viewer it probably does not matter, but it might for embedded viewer. I have added it for the nonce, we will see what effect it has.

 And no the iframe is sandboxed so it will never be able to modify anything in the main document.

kovidgoyal:
And as I thought, it breaks things in the embedded viewer: https://bugs.launchpad.net/calibre/+bug/1864263

kovidgoyal:
And I got hold of a surface go to test touch interaction in the standalone viewer. Setting touch-actions: none breaks finger based scrolling in the viewer preferences, for example. So you are going to have to find a different solution.

kovidgoyal:
Also, at least on my android phone I get no lag when scrolling via touch at the start of a chapter in flow mode, so what exactly are you talking about??

Siron777:
Clearly breaking scrolling somewhere else to avoid lag at the beginning of a chapter is not a trade-off to do. If I can see the impact for the library view, I do not see how it can explain the lag issue in the reading mode of the 1864263. If I understand well the current scrolling design for the book reader, the viewer ignore all the default behavior of chromium, is it correct? This issue looks more like the one 1858274 in page mode I was about to investigate (by the way, how can I have the list of closed bug in the webpage ?, now I have to find it back via my mail).

For the lag I was referring in this issue:

1. Go to the end of a chapter. In flow mode.
2. Scroll to the next chapter with a touch scroll.
3. In the new chapter, scroll down again with a touch event. Not a swift touch, a gentle one, like to scroll 5 lines (otherwise you do not see the lag). There you can have two behaviors:
   * It will not scroll while moving the finger, and when stopping, it will scroll abruptly.
   * It may also just ignore the scroll at all. You can see however that it is processing hold logic as it may select word on the trajectory of the finger. 

You do not see that behavior while scrolling in the middle of a chapter. You see the error "Ignored attempt to cancel a touchstart event with cancelable=false, for example because scrolling is in progress and cannot be interrupted" in the console when the lag occurs. I can also imagine that that bug is not present on Android but on Windows. For me the root cause of the bug is clearly in Chromium (touch event handling are registered as active, and the scrolling done when changing the chapter, with the fraction mechanism, is finished). I am trying to mitigate it.

kovidgoyal:
On Sat, Feb 22, 2020 at 03:26:50AM -0800, Simon wrote:
> Clearly breaking scrolling somewhere else to avoid lag at the beginning of a chapter is not a trade-off to do. If I can see the impact for the library view, I do not see how it can explain the lag issue in the reading mode of the 1864263. If I understand well the current scrolling design for the book reader, the viewer ignore all the default behavior of chromium, is it correct? This issue looks more like the one 1858274 in page mode I was about to investigate (by the way, how can I have the list of closed bug in the webpage ?, now I have to find it back via my mail).

You can search for closed bugs via launchapd web interface. Yes, the
viewer implements its wn touch gestures. However only in the iframe in
which content is displayed and its side margins. The rest of the
interface uses browser native touch gestures.

> 
> For the lag I was referring in this issue:
> 
> 1. Go to the end of a chapter. In flow mode.
> 2. Scroll to the next chapter with a touch scroll.
> 3. In the new chapter, scroll down again with a touch event. Not a swift touch, a gentle one, like to scroll 5 lines (otherwise you do not see the lag). There you can have two behaviors:
>    * It will not scroll while moving the finger, and when stopping, it will scroll abruptly.
>    * It may also just ignore the scroll at all. You can see however that it is processing hold logic as it may select word on the trajectory of the finger. 
> 

I'll try to reproduce on my surface go when I have a moment.

PR #1108, opened 2020-02-25
Added Spectator Australia, a new recipe; and maintenance

jamescridland:
Added Spectator Australia.
Removed Ads of the World, which has an unparsable result and is currently returning blank pages.
Amended two publications to correctly indicate they're in US English.


kovidgoyal:
Note that language = 'en' is assumed to be en_US so there is not need to
change those

jamescridland:
> Note that language = 'en' is assumed to be en_US so there is not need to change those <

In which case, removing those with en_US would be helpful. Should we be doing that?

kovidgoyal:
It doesn't matter since both are mapped to the same thing.

On Tue, Feb 25, 2020 at 10:43:02PM -0800, James Cridland wrote:
> > Note that language = 'en' is assumed to be en_US so there is not need to change those <
> 
> In which case, removing those with en_US would be helpful. Should we be doing that?
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/1108#issuecomment-591266588

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

jamescridland:
They're not:

Screen Shot 2020-02-26 at 4 51 40 pm

At the top: "English"
At the bottom: "English (United States)"

For consistency with all other languages, it would make sense for "en" to be switched to "en_US" on all US recipes. OR, if you'd like US English to be the default (I'm British, and that makes me shudder a bit) then we should remove all "en_US" to switch to "en".

Or, you could map them programmatically to the same thing. ;)

kovidgoyal:
There shouldn't be any en_US in that case. The should just be en

And yeah it kinda sucks that en is en_US, but it's a historical artifact
from when this feature was first created and there were only US
publications.

PR #1133, opened 2020-04-16
Fix for bug #1873106: when bools are not tristate, undefined values …

cbhaley:
…don't show emblems/icons as if the value is false. This fix requires the user to edit and save the rule.

### DIFF
--- a/src/calibre/gui2/custom_column_widgets.py
+++ b/src/calibre/gui2/custom_column_widgets.py
@@ -164,12 +164,13 @@ class Bool(Base):
         l.addWidget(c)
         c.clicked.connect(self.set_to_no)
 
-        t = _('Clear')
-        c = QPushButton(t, parent)
-        width = c.fontMetrics().boundingRect(t).width() + 7
-        c.setMaximumWidth(width)
-        l.addWidget(c)
-        c.clicked.connect(self.set_to_cleared)
+        if self.db.prefs.get('bools_are_tristate'):
+            t = _('Clear')
+            c = QPushButton(t, parent)
+            width = c.fontMetrics().boundingRect(t).width() + 7
+            c.setMaximumWidth(width)
+            l.addWidget(c)
+            c.clicked.connect(self.set_to_cleared)
 
         c = QLabel('', parent)
         c.setMaximumWidth(1)
--- a/src/calibre/gui2/preferences/coloring.py
+++ b/src/calibre/gui2/preferences/coloring.py
@@ -38,6 +38,10 @@ icon_rule_kinds = [(_('icon with text'), 'icon'),
 class ConditionEditor(QWidget):  # {{{
 
     ACTION_MAP = {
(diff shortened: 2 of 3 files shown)


kovidgoyal:
I'm not comfortable having code outside of the gui2 module depend on it. As I understand it the rule class is responsible for generating the template. This seems sub-optimal. If the user changes the tristate setting the rule will no longer work. May be better to have a special placeholder string for the template and dynamically replace it at run time? Or something similar.

cbhaley:
We probably should move library.coloring into the GUI. That is the only place it is used. 

The major problem with the placeholder idea is that the formatter is entirely self contained, so unless the tristate flag is passed in somehow the formatter can't know the value. Unfortunately the formatter is used in child processes so we can't pass in the value.

One possibility would be to change cache.get_metadata to honor the flag, changing None to False if bools are bistate. Of course lazy.ProxyMetadata.__getattribute__ would also require changes. This would work with child processes and external clients because the "correct" value would be in the OPF. The problem here is performance. Get_metadata would need always to check the type.

However, none of the above will deal with template "errors" when changing from bistate to tristate. There is no way to know what the user intended for both None and False. This can be fixed by making the user tell us, which requires adding some missing tests. The idea: change the condition options in preferences/coloring ACTION_MAP to "is true", "is not true", "is false", "is not false", "is undefined", "is not undefined". Alternative wording: "is true", "is true or false", "is true or undefined", "is false", "is false or undefined", "is undefind".

My suggestion: 
1) take out my changes in library.coloring.
2) move library.coloring into gui2. That is where it should be, really
3) implement the "alternative wording" extension of the bool col condition. This affects preferences.coloring and (the moved) library.coloring.

cbhaley:
Thought: if we leave the changes in library.coloring (check the gui) then when bools are bistate I can make the choices "is true", "is not true". In that case the bahavior will be the same if the user switches back to tristate.

kovidgoyal:
grepping library.coloring shows it is used db.backend for rule migration, so we cant move it into gui2. It should be perfectly possible to use a global for bools are tristate, that defaults to the default value and is set by library.models before calling the formatter. In fact, we could avoid it being a global and have it as a parameter passed to the formatter constructor with a default value.

cbhaley:
The formatter is used by the save-to-disk subprocess which can't see the DB, so it would get the default. That could result in different values than seen in the GUI, which I think isn't good.

I think the best solution is to generate a template that gives the right answer while not needing to know the value of tristate/bistate. I can do this without importing the gui library/coloring. I also have some changes that will improve performance. I will submit changes for this weekend (today is full with family stuff).

kovidgoyal:
Sounds good.

ywgyf:
Is the coronavirus situation in India still not serious enough to kill the developer? It's now 12739 and will  grow exponentially. May your life end up soon. Dirty people from dirty country all clear out of the earth.

PR #1135, opened 2020-04-24
Improve FAQ to emphasize issues with Android Devices

TheMageKing:
This is to prevent others from having the issues I ran into in bug #1874370


TheMageKing:
The download appears to have just crashed: no way did any of my changes cause any sort of failing checks.

kovidgoyal:
FYI calibre works fine with python3. See https://github.com/kovidgoyal/calibre/pull/870

PR #1138, opened 2020-04-28
Fix for #1875758 - double space issue

edent:
Shows up as a double space on the pop-up.
(I hope this is the right branch for Pull Requests)


eli-schwartz:
Pull requests go to master.

The py3 branch contains experimental beta changes for porting to python3 ONLY, and is periodically rebased onto master.

edent:
D'oh! Thanks :-)

eli-schwartz:
Note that it's probably not really necessary to include "I hope this is the right branch" in the commit message itself... you should be able to leave that out of the commit message and just edit the Pull Request description before submitting, which has a tendency to be less immortalizing. ;)

PR #1141, opened 2020-05-04
Allow compiling with VS2019

japm48:
This patch uses `distutils.ccompiler` (which calls vswhere.exe under the hood). This makes it compatible with newer MSVC versions.

See: https://github.com/python/cpython/blob/master/Lib/distutils/_msvccompiler.py

I managed to build and run calibre with this. Everything works, except building the installer, but it seems bypy still needs some python3 porting work anyways.

### DIFF
--- a/setup/build_environment.py
+++ b/setup/build_environment.py
@@ -14,14 +14,26 @@ is64bit
 
 NMAKE = RC = msvc = MT = win_inc = win_lib = None
 if iswindows:
-    from distutils import msvc9compiler
-    msvc = msvc9compiler.MSVCCompiler()
-    msvc.initialize()
-    NMAKE = msvc.find_exe('nmake.exe')
-    RC = msvc.find_exe('rc.exe')
-    MT = msvc.find_exe('mt.exe')
-    win_inc = [x for x in os.environ['include'].split(';') if x]
-    win_lib = [x for x in os.environ['lib'].split(';') if x]
+    try:
+        from distutils.ccompiler import new_compiler
+        msvc = new_compiler('nt', 'msvc')
+        msvc.initialize()
+        RC = msvc.rc
+        MT = msvc.mt
+        # hackish solution (see https://bugs.python.org/issue40478)
+        NMAKE = os.path.dirname(msvc.cc) + os.sep + 'nmake.exe'
+        win_inc = msvc.include_dirs
+        win_lib = msvc.library_dirs
+    except:
+        # Fallback to older version.
+        from distutils import msvc9compiler
+        msvc = msvc9compiler.MSVCCompiler()
+        msvc.initialize()
+        NMAKE = msvc.find_exe('nmake.exe')
+        RC = msvc.find_exe('rc.exe')
+        MT = msvc.find_exe('mt.exe')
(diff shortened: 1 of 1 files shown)


japm48:
Well, it doesn't pass CI with VS2017. I'll close this while I try to find a solution.

kovidgoyal:
And FYI bypy works fine with python3, it is used with the py3 branch of calibre.

japm48:
@kovidgoyal, the problem occurs when compiling with VS2019, which doesn't work with `distutils.msvc9compiler` (unless I'm missing something).

I'm using the py3 branch because I believe VS2019 is not compatible with python 2 (the official python.org binaries are compiled with VS2008).

I used the CI binaries for the dependencies (i.e. https://download.calibre-ebook.com/ci/calibre3/windows-64.tar.xz), and everything worked properly.

For the record, the error I get without the patch is:

```
...> %SW%\private\python\python.exe .\setup.py bootstrap
Already up to date.
*
* Running build
*

  File ".\setup.py", line 122, in <module>
    sys.exit(main())
  File ".\setup.py", line 107, in main
    command.run_all(opts)
  File "C:\tmp\calibre_build\calibre\setup\__init__.py", line 251, in run_all
    self.run_cmd(self, opts)
  File "C:\tmp\calibre_build\calibre\setup\__init__.py", line 243, in run_cmd
    self.run_cmd(scmd, opts)
  File "C:\tmp\calibre_build\calibre\setup\__init__.py", line 247, in run_cmd
    cmd.run(opts)
  File "C:\tmp\calibre_build\calibre\setup\build.py", line 267, in run
    self.env = init_env()
  File "C:\tmp\calibre_build\calibre\setup\build.py", line 157, in init_env
    from setup.build_environment import msvc, is64bit, win_inc, win_lib, NMAKE
  File "C:\tmp\calibre_build\calibre\setup\build_environment.py", line 19, in <module>
    msvc.initialize()
  File "C:\tmp\calibre_build\calibre_inst\private\python\lib\distutils\msvc9compiler.py", line 372, in initialize
    vc_env = query_vcvarsall(VERSION, plat_spec)
  File "C:\tmp\calibre_build\calibre_inst\private\python\lib\distutils\msvc9compiler.py", line 261, in query_vcvarsall
    raise DistutilsPlatformError("Unable to find vcvarsall.bat")
distutils.errors.DistutilsPlatformError: Unable to find vcvarsall.bat
```

kovidgoyal:
If you want to build with python2 use my python2 fork which works with VS 2017. Why are you trying to compile with VS2019 anyway??

japm48:
To be sincere, the real reason is that I don't like having installed multiple versions of VS (SSD space :smile:). Especially when I just wanted to try the latest version.

But more seriously, if two compiler versions are supposed to be ABI compatible, one should not force to use a particular one, IMHO. (You are, of course, free to disagree).

kovidgoyal:
There's a lot more to compiler fungibility than ABI compatibility. And
given that windows is not my primary environment, there is limited
effort I am willing to spend supporting multiple compilers on it. But,
patches are welcome, as long as they dont break VS2017 builds.

PR #1146, opened 2020-05-13
Enhancement: allow simultaneous editing in the category editor. …

cbhaley:
…Select multiple items (tags, whatever), then edit. When you finish editing, all the items will change to the edited value.

Kovid, this change was part of an enhancement request, but I am not convinced that it is a good change to make. What do you think? If you think it is benign, go ahead and accept it.

### DIFF
--- a/src/calibre/gui2/dialogs/tag_list_editor.py
+++ b/src/calibre/gui2/dialogs/tag_list_editor.py
@@ -4,10 +4,13 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from PyQt5.Qt import (Qt, QDialog, QTableWidgetItem, QIcon, QByteArray, QSize,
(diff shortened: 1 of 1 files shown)


kovidgoyal:
For this to be workable, there needs to be some indication that editing the current cell will change all selected cells, maybe change the display of selected but not editing cells to "editing" or similar.

cbhaley:
Pushed a change to mark the other rows with *** EDITING ***. This is a bit fiddly because AFAIK qt doesn't tell me if editing has been abandoned with no changes. I had to trap key events.

kovidgoyal:
As I recall editors can be closed by clicking else where in the table as well. Is subclassing the item delegate and overriding createEditor() and destroyEditor() not sufficient?

cbhaley:
I was looking for a signal and didn't see the editorDestroyed method. It seems to work correctly now.

PR #1151, opened 2020-05-16
Two commits: ...

cbhaley:
1) Tag browser 'find' improvements
2) Enhancement #1878940: allow restricting manage authors to the current VL (and other things)

More detail in the individual commits


kovidgoyal:
In the find shows all matches mode, I dont seem to be able to reset a search. Once I do a search, clicking the rred cross to clear it does not reset to showing all items again.

cbhaley:
Sorry about that. I didn't look for the signals for the X inside the line edit box.

cbhaley:
I pushed a better fix for the problem you found, one that preserves interface compatibility. I also pushed as small fix to deal the "empty box" problem that Nyberg reported.

kovidgoyal:
I have refactored the manage authors dialog a bit to use standard Qt sort indicators, please check that I didn't break anything.

PR #1157, opened 2020-05-23
Update Swedish translation of the Quick Start Guide

jony0008:


jony0008:
@kovidgoyal Do you know why the test failed?

kovidgoyal:
That's a network test failing, those sometimes fail when the server the test is running on is too slow.

PR #1164, opened 2020-05-27
Fix for slowdown if a file URL refers to a missing drive. Might be a …

cbhaley:
…windows-only thing.


cbhaley:
Kovid, the second commit puts the image bytes in the code as a hex string.

PR #1165, opened 2020-05-27
Enhancement #1880264: Additional functionality for tag hierarchies

cbhaley:
This is a significant change to the tag browser.

Two additions:
1) right-click "rename" on an intermediate node renames that node and all the children.
2) Drag & Drop inside a hierarchical category moves the source to the target

Important behavior changes:
1) deleting an item deletes that item and now in addition all its children.
2) renaming an item renames it and now in addition all its children.
3) you can now operate on virtual intermediate nodes that aren't on any book

For #3, consider the tag "foo.bar.mumble" which is on a book. The tag browser creates
2 virtual nodes: foo and foo.bar. Before you couldn't rename and delete them. Now you can.

Before, renaming a non-virtual node with children left a new real node for the renamed node and a new virtual node for the children. There was no way to rename a hierarchy.

### DIFF
--- a/src/calibre/gui2/tag_browser/model.py
+++ b/src/calibre/gui2/tag_browser/model.py
@@ -59,6 +59,8 @@ class TagTreeItem(object):  # {{{
         self.blank = QIcon()
         self.is_gst = False
         self.boxed = False
+        self.temporary = False
+        self.can_be_edited = False
         self.icon_state_map = list(icon_map)
         if self.parent is not None:
             self.parent.append(self)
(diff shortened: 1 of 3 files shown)


cbhaley:
Correction to the above. You can now rename a virtual node but not delete it. Only real nodes and their children can be deleted. Deleting all the children of a virtual node deletes that node.

kovidgoyal:
I have merged, however, one issue I see:

If you drag a tag and drop it and then try to drag again, it doesnt
work. You have to mousedown, mouseup and then dragging works. Probably
you need to reset possible_drag_start on a drop.

PR #1173, opened 2020-06-19
recipes: update gazeta prawna recipe

tjozwiak:
This patch removes obsolete 'gazeta-prawna-calibre-v1.recipe'
and adds new one gazeta-prawna-calibre-new.recipe instead

Signed-off-by: Tomasz Jozwiak <tjozwiakgm@gmail.com>

### DIFF
--- a/recipes/gazeta-prawna-calibre-v1.recipe
+++ b/recipes/gazeta-prawna-calibre-v1.recipe
@@ -1,62 +1,110 @@
 #!/usr/bin/env python2
 
 __license__ = 'GPL v3'
-__copyright__ = u'2011, Vroo <vroobelek@iq.pl>'
-__author__ = u'Vroo'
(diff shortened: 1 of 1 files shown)


kovidgoyal:
You should not rename the recipe file, just replace its contents.

tjozwiak:
Hi, Thx for your comment. I've changed the file name cause this file was absolutely obsolete and suggests first version of something. Also the author of this file is uknow now. To keep this file we also should keep the author. So may be better is create new name. What do you think?

tjozwiak:
BTW I've changed probably everything in this file :)

kovidgoyal:
Creating a new name means that any user that has scheduled the recipe
for download will no longer be able to schedule it and will have to redo
the schedule. Feel free to change the author in the existing recipe
file, if you have completely re-written it.

tjozwiak:
ok

PR #1174, opened 2020-06-20
Enhancement #1884343: Add action to show a tag in the Tag browser …

cbhaley:
…when right clicking on a tag in the Book details panel.

Works for is_multiple editable fields (not composites). Also made it possible to remove languages from the book details context menu.


cbhaley:
Oops -- it works for editable category fields, not just is_multiples.

PR #1179, opened 2020-06-25
build: detect mathjax directory with or without es5/ subdir

eli-schwartz:
The release tarballs contain an es5/ directory with the actual mathjax code. However, when building against a system mathjax it is likely the es5/ directory is skipped and the directory name is just /usr/share/mathjax/

Detect this by moving the es5/ component to the definition of the mathjax source root, but only appending it if the --path-to-mathjax actually contains such a directory.

### DIFF
--- a/setup/mathjax.py
+++ b/setup/mathjax.py
@@ -43,11 +43,13 @@ class MathJax(ReVendor):
         os.mkdir(self.vendored_dir)
         with self.temp_dir(suffix='-calibre-mathjax-build') as tdir:
             src = opts.path_to_mathjax or self.download_vendor_release(tdir, opts.mathjax_url)
+            if os.path.isdir(os.path.join(src, 'es5')):
+                src = os.path.join(src, 'es5')
             self.info('Adding MathJax...')
             for x in 'core loader startup input/tex-full input/asciimath input/mml input/mml/entities output/chtml'.split():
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Nice work porting to mathjax3. Arch Linux *does* package it (we are the only ones so far, other than CRUX), but I need this commit in order to successfully run `python3 setup.py mathjax --path-to-mathjax /usr/share/mathjax --system-mathjax`.

eli-schwartz:
Thanks!

@norbusan @nirik @keszybz Debian and Fedora respectively will need to package mathjax 3.x in order to continue using --system-mathjax. Both distros which do so right now (according to repology), do so without the es5/ directory, so there is that.

My test build of git master is successful now that this change has landed.

norbusan:
Thanks Eli @eli-schwartz 
Thanks for the headsup! I will see what we can do. I am not responsible for mathjax, and there seem to be quite some changes when switching to mathjax3. I think in the worst case we can use the calibre version of mathjax until mathjax3 is shipped in Debian.

nirik:
Yeah, thanks. Fedora isn't moved to mathjax3 yet, but I can just use the bundled one until then... thanks for the headsup.

eli-schwartz:
Arch has both:
- "mathjax" 3.x == /usr/share/mathjax
- "mathjax2" 2.x == /usr/share/mathjax2

PR #1184, opened 2020-07-02
plistlib.Data is going to be dropped in python 3.9

nirik:
Here is a rough cut at removing any plistlib.Data imports and switching
to just using bytes for the objects. No idea how close to correct it is,
but without this calibre using python 3.9 doesn't finish it's startup,
and with it it seems to run fine.

Signed-off-by: Kevin Fenzi <kevin@scrye.com>

### DIFF
--- a/src/calibre/utils/config.py
+++ b/src/calibre/utils/config.py
@@ -363,29 +363,26 @@ class XMLConfig(dict):
         self.update(d)
 
     def __getitem__(self, key):
-        from polyglot.plistlib import Data
         try:
             ans = dict.__getitem__(self, key)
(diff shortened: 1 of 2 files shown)


eli-schwartz:
https://docs.python.org/3/library/plistlib.html#plistlib.Data

> Deprecated since version 3.4: Use a bytes object instead.

Seems fair to me. :D

eli-schwartz:
Actually, looking at the changes -- I'd simply remove these calls entirely.

```
>>> Data(b'foo').data
b'foo'
```

But this doesn't work for bytes(). Just store them natively as bytes and extract them without changes.

PR #1186, opened 2020-07-04
Enhancement #1886219: In category editor, set focus to selected item

cbhaley:


kovidgoyal:
Shouldnt this be done only if there is a pre-selected item? Otherwise focusing search is more appropriate

cbhaley:
There is always a preselected item. It is cell 0,0 if no item was passed in. I thought it better to be consistent. Happy to change it if you prefer.

kovidgoyal:
I meant when the dialog is opened by right clicking on the, for example,
"tags" category rather than an individual tag. In such a case, IMO it
makes more sense to focus search rather than the table.

cbhaley:
Done

PR #1187, opened 2020-07-04
Enhancement #1886079: default value for custom columns

cbhaley:
Kovid, I haven't marked this as "committed" in launchpad because I am not sure you will accept the change.

### DIFF
--- a/src/calibre/db/cache.py
+++ b/src/calibre/db/cache.py
@@ -92,6 +92,29 @@ def _add_newbook_tag(mi):
                     mi.tags.append(tag)
 
 
+def _add_default_custom_column_values(mi, fm):
+    cols = fm.custom_field_metadata(include_composites=False)
+    for cc,col in iteritems(cols):
+        dv = col['display'].get('default_value', '')
+        try:
+            if dv:
+                if not mi.get_user_metadata(cc, make_copy=False):
+                    mi.set_user_metadata(cc, col)
+                dt = col['datatype']
+                if dt == 'bool':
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Seems fairly harmless to me, but isnt the rating supposed to be int(val * 2) not int(val) * 2?

kovidgoyal:
Also for yes no clumns maybe allow yes/no in english as default values otherwise the default value will not work on interface language change.

cbhaley:
I pushed the changes to add English "yes", "no", and "now" as permitted values.

Re  int(val) * 2: in this case val is known to be a string. it is my understanding that Python 3 removed implicit conversion for multiply so ("2" * 3) == "222", not 6. If I am right then we must explicitly convert the string to int before the multiply. Once we have done that the multiply will produce an int result. calibre.db.write accepts an integer type for rating so there is no need to coerce it back to a string.

kovidgoyal:
Hmm, sorry I somehow skipped over the fact that the default value is stored as a string. In that case some further comments:

1) For rating we want int(float(val) * 2) otherwise half star ratings will not work
2) For yes/no columns I think it is better to normalize the stored default value rather than normalize it at the time of adding. otherwise say a german user sends his db to a french user, the default value will not work.

Indeed, I think for all value types normalization should occur at define time not at use time. The data stored in the db should be locale neutral and of appropriate range/type. Since this is json, for ratings and numeric columns might as well store actual numbers.

cbhaley:
I changed the default value stuff to use native types (and english) in the display. I also added lots of error checking.

PR #1198, opened 2020-07-26
Py3

mwgabby-li:
Hi Mr. Goyal -

I realized it would be more effective to split out RTL support and vertical RTL support into two changelists. Basic horizontal RTL support is properly done as a prerequisite. I had my sister test this and she says that Hebrew books are now displaying properly.
I tried to test thoroughly. Everything should be defaulting to what it did before in books without RTL direction and _page-progression-direction: rtl_, so it should only be an improvement.

Three notes:

1. I am new to Git, so if I made some mistake when merging the history or something, let me know and I can clean it up and re-do it.
2. It's been a while since I wrote Python code, so keep an eye out for novice mistakes, and my knowledge of CSS and HTML is not great.
3. I wasn't 100% sure which file the function _rtl_page_progression()_ should go, so feel free to suggest a different location.

Here is a more thorough description.

Support for Books in RTL Languages, Such as Hebrew

Three parts: RTL layout support, made inputs behave naturally in RTL modes, and updated tutorial for RTL books.

To add RTL layout support:
 ‣ Don't force LTR layout on DOM when loading.
 ‣ Disable "overflow: hidden" in RTL mode to compensate for an issue in Chrome, and disable scrollbars on all browsers to compensate for the side effect of this turning on scrollbars.
 ‣ Have viewport hide the RTL mode by negating scrolls and viewport position requests in RTL mode.
 ‣ Initialize viewport in paged mode layout so that it properly saves RTL mode.
 ‣ Remove all raw calls to pageXOffset and pageYOffset 
 ‣ Put viewport_to_document in ScrollViewport so that it can abstract away RTL mode in this case, too.
   ‣ Get rid of calls to reset_transforms and make it private, since it is only called to prepare for viewport_to_document. Put the call there, instead.
 ‣ Use bounding box right instead of left for position of various elements when the viewport is in RTL mode, since the right side is the beginning of the element in this case.

To made input behave naturally in RTL:
 ‣ Get page progression direction from ePub.
 ‣ Add rtl_page_progression() function to return whether we're in this mode.
 ‣ Added support to tell page advance functions whether they should flip their direction if RTL is on.
 ‣ Changed touch behavior to support different position for going back and forward in RTL books.

Tutorial changes:
 ‣ Flipped tutorial so that next and previous page are reversed in position and size in RTL mode.
 ‣ Added a new counter so that RTL tutorial is shown even if user has already seen LTR tutorial.


src/calibre/srv/render_book.py
```diff
@@ -639,6 +639,12 @@ def work_priority(name):
     spineq = frozenset(spine)
     landmarks = [l for l in get_landmarks(container) if l['dest'] in spineq]
 
+    page_progression_direction = None
+    try:
+        page_progression_direction = container.opf_xpath('//opf:spine/@page-progression-direction')[0]
+    except IndexError:
+        pass
+
```
mwgabby-li, line 647: This is based on your suggestion in your email.

src/pyj/read_book/cfi.pyj
```diff
@@ -57,21 +57,6 @@ def window_scroll_pos(w):  # {{{
...
-        rect = frame.getBoundingClientRect()
-        x += rect.left
-        y += rect.top
-        doc = frame.ownerDocument
-    wx, wy = window_scroll_pos(doc.defaultView)
-    x += wx
-    y += wy
-    return x, y
-# }}}
-
```
mwgabby-li, line 74: ScrollViewport now has contextual information about the viewport, namely whether things are in RTL mode, so it needs to wrap this to make sure the returned coordinates are in the correct space.

src/pyj/read_book/cfi.pyj
```diff
@@ -634,7 +619,6 @@ def scroll_to(cfi, callback, doc):  # {{{
         span.setAttribute('style', 'border-width: 0; padding: 0; margin: 0')
         r.surroundContents(span)
         scroll_viewport.scroll_into_view(span)
-        scroll_viewport.reset_transforms()  # needed for viewport_to_document()
```
mwgabby-li, line 637: This is now called in ScrollViewport's viewport_to_document function.

src/pyj/read_book/cfi.pyj
```diff
@@ -677,17 +661,18 @@ def scroll_to(cfi, callback, doc):  # {{{
...
         node = point_.node
         scroll_viewport.scroll_into_view(node)
-        scroll_viewport.reset_transforms()  # needed for viewport_to_document()
 
         fn = def():
             r = node.getBoundingClientRect()
-            x, y = viewport_to_document(r.left, r.top, node.ownerDocument)
+            # Start of element is right side in RTL, so be sure to get that side in RTL mode
+            x, y = scroll_viewport.viewport_to_document(
+                r.left if scroll_viewport.ltr() else r.right, r.top, node.ownerDocument)
```
mwgabby-li, line 675: Various places assume the left side of a bounding rectangle is the start of an element. I changed this to be dependent on ScrollViewport's RTL setting.

PR #1199, opened 2020-07-26
Support for Books in RTL Languages, Such as Hebrew

mwgabby-li:
Three parts: RTL layout support, made inputs behave naturally in RTL modes, and updated tutorial for RTL books.

To add RTL layout support:
 ‣ Don't force LTR layout on DOM when loading.
 ‣ Disable "overflow: hidden" in RTL mode to compensate for an issue in Chrome, and disable scrollbars on all browsers to compensate for the side effect of this turning on scrollbars.
 ‣ Have viewport hide the RTL mode by negating scrolls and viewport position requests in RTL mode.
 ‣ Initialize viewport in paged mode layout so that it properly saves RTL mode.
 ‣ Remove all raw calls to pageXOffset and pageYOffset
 ‣ Put viewport_to_document in ScrollViewport so that it can abstract away RTL mode in this case, too.
   ‣ Get rid of calls to reset_transforms and make it private, since it is only called to prepare for viewport_to_document. Put the call there, instead.
 ‣ Use bounding box right instead of left for position of various elements when the viewport is in RTL mode, since the right side is the beginning of the element in this case.

To made input behave naturally in RTL:
 ‣ Get page progression direction from ePub.
 ‣ Add rtl_page_progression() function to return whether we're in this mode.
 ‣ Added support to tell page advance functions whether they should flip their direction if RTL is on.
 ‣ Changed touch behavior to support different position for going back and forward in RTL books.

Tutorial changes:
 ‣ Flipped tutorial so that next and previous page are reversed in position and size in RTL mode.
 ‣ Added a new counter so that RTL tutorial is shown even if user has already seen LTR tutorial.


src/pyj/read_book/cfi.pyj
```diff
@@ -57,21 +57,6 @@ def window_scroll_pos(w):  # {{{
...
-        rect = frame.getBoundingClientRect()
-        x += rect.left
-        y += rect.top
-        doc = frame.ownerDocument
-    wx, wy = window_scroll_pos(doc.defaultView)
-    x += wx
-    y += wy
-    return x, y
-# }}}
-
```
mwgabby-li, line 74: The ScrollViewport fools the rest of the code into thinking that things progress left-to-right, in a positive direction, but in RTL mode, things go right-to-left and become negative.

The coordinates this will return in RTL mode will be negative - but we want to use the ScrollViewport to pretend everything is positive, so move this function to the viewport and have it negate the value of the X coordinate returned if in RTL mode.

src/pyj/read_book/cfi.pyj
```diff
@@ -634,7 +619,6 @@ def scroll_to(cfi, callback, doc):  # {{{
         span.setAttribute('style', 'border-width: 0; padding: 0; margin: 0')
         r.surroundContents(span)
         scroll_viewport.scroll_into_view(span)
-        scroll_viewport.reset_transforms()  # needed for viewport_to_document()
```
mwgabby-li, line 637: This is only called in preparation for viewport_to_document(), but that function is now part of ScrollViewport. So, just call it in ScrollViewport::viewport_to_document, and don't call it outside.

src/pyj/read_book/cfi.pyj
```diff
@@ -677,17 +661,18 @@ def scroll_to(cfi, callback, doc):  # {{{
 
             x = (point_.a*rect.left + (1-point_.a)*rect.right)
             y = (rect.top + rect.bottom)/2
-            x, y = viewport_to_document(x, y, ndoc)
+            x, y = scroll_viewport.viewport_to_document(x, y, ndoc)
```
mwgabby-li, line 664: Everywhere this function is called, use ScrollViewport, for the reason described in the comment on cfi.pyj.

src/pyj/read_book/cfi.pyj
```diff
@@ -677,17 +661,18 @@ def scroll_to(cfi, callback, doc):  # {{{
...
         node = point_.node
         scroll_viewport.scroll_into_view(node)
-        scroll_viewport.reset_transforms()  # needed for viewport_to_document()
 
         fn = def():
             r = node.getBoundingClientRect()
-            x, y = viewport_to_document(r.left, r.top, node.ownerDocument)
+            # Start of element is right side in RTL, so be sure to get that side in RTL mode
+            x, y = scroll_viewport.viewport_to_document(
+                r.left if scroll_viewport.ltr() else r.right, r.top, node.ownerDocument)
```
mwgabby-li, line 675: In cases where the top left of an element is used to get its start position, and we're in RTL mode, use the top right since that's where things begin in horizontal RTL mode.

src/pyj/read_book/flow_mode.pyj
```diff
@@ -3,10 +3,10 @@
 from __python__ import bound_methods, hash_literals
 
 from dom import set_css
-from read_book.globals import current_spine_item, get_boss
+from read_book.globals import current_spine_item, get_boss, rtl_page_progression, ltr_page_progression
```
mwgabby-li, line 6: Two functions. LTR is usually used in cases where we want the default behavior to come first in a clause, RTL is used usually in cases where we override the value afterward.

src/pyj/read_book/flow_mode.pyj
```diff
@@ -57,7 +57,7 @@ last_change_spine_item_request = {}
 
 def _check_for_scroll_end(func, obj, args, report):
     before = window.pageYOffset
-    func.apply(obj, args)
+    should_flip_progression_direction = func.apply(obj, args)
```
mwgabby-li, line 60: Different movement functions may or may not want to reverse the page progression direction. For example, the scrollwheel down is naturally "next page" even in RTL books, but clicking on the left margin should go forward in RTL mode and backward in LTR mode.

src/pyj/read_book/flow_mode.pyj
```diff
@@ -114,15 +119,19 @@ def flow_onwheel(evt):
 
 @check_for_scroll_end
 def goto_boundary(dir):
-    scroll_viewport.scroll_to(window.pageXOffset, 0 if dir is DIRECTION.Up else document_height())
+    scroll_viewport.scroll_to(scroll_viewport.x(), 0 if dir is DIRECTION.Up else document_height())
```
mwgabby-li, line 122: It's important to use the value from ScrollViewport, again, so that the value goes positive from zero in both RTL and LTR books from the perspective of this code.

src/pyj/read_book/flow_mode.pyj
```diff
@@ -114,15 +119,19 @@ def flow_onwheel(evt):
...
     get_boss().report_human_scroll()
 
 
 @check_for_scroll_end_and_report
-def scroll_by_page(direction):
+def scroll_by_page(direction, flip_if_rtl_page_progression):
```
mwgabby-li, line 127: The functions that call this one know whether or not the action should be reversed in RTL, so I added a flag here.

src/pyj/read_book/flow_mode.pyj
```diff
@@ -168,10 +177,10 @@ def handle_shortcut(sc_name, evt):
         goto_boundary(DIRECTION.Down)
         return True
     if sc_name is 'left':
-        window.scrollBy(-15, 0)
+        window.scrollBy(-15 if ltr_page_progression() else 15, 0)
```
mwgabby-li, line 180: This is a case where I wanted to put the default, LTR, first, for clarity.

src/pyj/read_book/globals.pyj
```diff
@@ -21,6 +21,14 @@ def current_book():
...
 current_book.book = None
 
+def rtl_page_progression():
+    # Other options are "ltr" and "default." For Calibre, "default" is LTR.
+    return current_book().manifest.page_progression_direction == 'rtl'
+
+def ltr_page_progression():
+    # Only RTL and LTR are supported, so it must be LTR if not RTL.
+    return not rtl_page_progression()
+
```
mwgabby-li, line 31: globals.py seemed like a good place, but feel free to move it or ask me to move it if somewhere else is better.

src/pyj/read_book/iframe.pyj
```diff
@@ -323,9 +323,9 @@ class IframeBoss:
     def on_next_screen(self, data):
         backwards = data.backwards
         if current_layout_mode() is 'flow':
-            flow_scroll_by_page(-1 if backwards else 1)
+            flow_scroll_by_page(-1 if backwards else 1, data.flip_if_rtl_page_progression)
```
mwgabby-li, line 326: Added this new value to the data since we don't know here whether or not we should flip.

src/pyj/read_book/paged_mode.pyj
```diff
@@ -40,14 +39,9 @@ def has_start_text(elem):
...
 
 def handle_rtl_body(body_style):
-    # Make the body and root nodes have direction ltr so that column layout
-    # works as expected
     if body_style.direction is "rtl":
-        for node in document.body.childNodes:
-            if node.nodeType is Node.ELEMENT_NODE and window.getComputedStyle(node).direction is "rtl":
-                node.style.setProperty("direction", "rtl")
-        document.body.style.direction = "ltr"
-        document.documentElement.style.direction = 'ltr'
```
mwgabby-li, line 50: I needed to disable this, since it makes columns in RTL books display in the wrong order, left to right instead of right to left.

src/pyj/read_book/paged_mode.pyj
```diff
@@ -40,14 +39,9 @@ def has_start_text(elem):
...
+        # If this in not set, Chrome scrolling breaks for some RTL and vertical content.
+        document.documentElement.style.overflow = 'visible'
```
mwgabby-li, line 44: Chrome really doesn't like content that overflows on the left side.

src/pyj/read_book/paged_mode.pyj
```diff
@@ -170,6 +164,7 @@ def current_page_width():
 def layout(is_single_page, on_resize):
     nonlocal _in_paged_mode, col_width, col_and_gap, screen_height, gap, screen_width, is_full_screen_layout, cols_per_screen, number_of_cols
     line_height(True)
+    scroll_viewport.initialize_on_layout()
```
mwgabby-li, line 167: Note that this is not called in flow mode (yet) because flow mode actually doesn't care about the body's text direction, just the page progression direction from the spine.

src/pyj/read_book/paged_mode.pyj
```diff
@@ -435,25 +429,29 @@ def scroll_to_elem(elem):
...
+            # Columns start on the right side in RTL mode, so get that instead here...
+            pos = elem.scrollLeft if scroll_viewport.ltr() else elem.scrollRight
         else:
-            left = br.left
-        scroll_to_xpos(viewport_to_document(
-            left+2, elem.scrollTop, elem.ownerDocument)[0])
+            # and here.
+            pos = br.left if scroll_viewport.ltr() else br.right
+        scroll_to_xpos(scroll_viewport.viewport_to_document(
+            pos+2, elem.scrollTop, elem.ownerDocument)[0])
```
mwgabby-li, line 438: Because of the reversal of the position internally in ScrollViewport, this will actually be +2 to the left in RTL mode and +2 to the right in LTR mode.

src/pyj/read_book/paged_mode.pyj
```diff
@@ -615,10 +616,10 @@ def scroll_to_extend_annotation(backward):
 
 def handle_shortcut(sc_name, evt):
     if sc_name is 'up':
-        scroll_by_page(True, True)
+        scroll_by_page(backward=True, by_screen=True, flip_if_rtl_page_progression=False)
```
mwgabby-li, line 619: I added named parameters for all of these because it was getting really hard to read the logic with a series of three booleans.

src/pyj/read_book/paged_mode.pyj
```diff
@@ -661,19 +662,24 @@ def handle_gesture(gesture):
                 get_boss().send_message('next_section', forward=gesture.direction is 'up')
         else:
             if not gesture.active or gesture.is_held:
-                scroll_by_page(gesture.direction is 'right', True)
+                scroll_by_page(gesture.direction is 'right', True, flip_if_rtl_page_progression=True)
+    # Gesture progression direction is determined in the gesture code;
+    # don't set flip_if_rtl_page_progression=True here.
     elif gesture.type is 'prev-page':
```
mwgabby-li, line 668: I pondered changing the gesture code to send "go left" and "go right" actions, but it seemed pointless because the gesture code has to know what the page progression direction is in order to make the left or right side a bigger touch region.

src/pyj/read_book/settings.pyj
```diff
@@ -88,6 +88,13 @@ def apply_colors():
     selbg = make_selection_background_opaque(selbg)
     text += f'\n::selection {{ background-color: {selbg}; color: {selfg} }}'
     text += f'\n::selection:window-inactive {{ background-color: {selbg}; color: {selfg} }}'
+    # In Chrome when content overflows in RTL and vertical layouts on the left side,
+    # it is not displayed properly when scrolling unless overflow:visible is set,
+    # but this causes scrollbars to appear.
+    # Force disable scrollbars in Chrome, Safari, and Firefox to address this side effect.
+    text += '\nhtml::-webkit-scrollbar, body::-webkit-scrollbar { display: none }'
+    text += '\nhtml { scrollbar-width: none; }'
+    text += '\nbody { scrollbar-width: none; }'
```
mwgabby-li, line 97: I hope I hacked this correctly. Safari seems like it will respect the webkit flags, and I tested Firefox in the web server and there were no scrollbars.

src/pyj/read_book/view.pyj
```diff
@@ -107,6 +107,19 @@ def show_controls_help():
...
+    right_msg = msg(_('Tap to turn page'))
+    right_width = '75vw'
+    if rtl_page_progression():
+        left_msg, right_msg = right_msg, left_msg
+        left_width, right_width = right_width, left_width
+
+    # Clear it out if this is not the first time it's created.
+    # Needed to correctly show it again in a different page progression direction.
+    if container.firstChild:
+        container.removeChild(container.firstChild)
```
mwgabby-li, line 121: This was actually producing a doubled tutorial overlay before, since it's called twice, so the first time I tried this, the last book's tutorial overlay kept showing up, even if the new book was a different direction.

src/pyj/read_book/view.pyj
```diff
@@ -322,7 +335,10 @@ class View:
         if event.button is 0:
             event.preventDefault(), event.stopPropagation()
             sd = get_session_data()
-            self.iframe_wrapper.send_message('next_screen', backwards=True, all_pages_on_screen=sd.get('paged_margin_clicks_scroll_by_screen'))
+            self.iframe_wrapper.send_message(
+                'next_screen', backwards=True,
+                flip_if_rtl_page_progression=True,
+                all_pages_on_screen=sd.get('paged_margin_clicks_scroll_by_screen'))
```
mwgabby-li, line 341: Hope you don't mind the reformatting here. The line was getting pretty long.
This is a place where we need to send a flip signal since it's the left margin.

src/pyj/read_book/view.pyj
```diff
@@ -475,10 +494,14 @@ class View:
             self.overlay.open_book()
         elif data.name is 'next':
             self.iframe_wrapper.send_message(
-                'next_screen', backwards=False, all_pages_on_screen=get_session_data().get('paged_margin_clicks_scroll_by_screen'))
+                'next_screen', backwards=False,
+                flip_if_rtl_page_progression=False,
+                all_pages_on_screen=get_session_data().get('paged_margin_clicks_scroll_by_screen'))
         elif data.name is 'previous':
```
mwgabby-li, line 500: I actually cannot find where 'next' is sent. The only place that looks suspicious is toolbars.py, which seems like the reader UI that shows up on right-click, but I don't see a next page butter there.

src/pyj/read_book/view.pyj
```diff
@@ -827,10 +850,10 @@ class View:
         else:
             self.show_name(name, initial_position=pos)
         sd = get_session_data()
-        c = sd.get('controls_help_shown_count', 0)
+        c = sd.get('controls_help_shown_count' + ('_rtl_page_progression' if rtl_page_progression() else ''), 0)
```
mwgabby-li, line 853: Since the tutorial overlay changes based on reading mode, I thought it made sense to show it again if a book is in a different reading mode, and just doing it twice for each mode seemed like the simplest solution.

src/pyj/read_book/viewport.pyj
```diff
@@ -2,7 +2,7 @@
 # License: GPL v3 Copyright: 2017, Kovid Goyal <kovid at kovidgoyal.net>
 from __python__ import bound_methods, hash_literals
 
-FUNCTIONS = 'x y scroll_to scroll_into_view reset_globals reset_transforms'.split(' ')
+FUNCTIONS = 'x y scroll_to scroll_into_view reset_globals __reset_transforms'.split(' ')
```
mwgabby-li, line 5: Made __reset_transforms private since it doesn't need to be called outside of this class.

src/pyj/read_book/viewport.pyj
```diff
@@ -12,13 +12,28 @@ class ScrollViewport:
...
     def set_mode(self, mode):
         prefix = ('flow' if mode is 'flow' else 'paged') + '_'
         for attr in FUNCTIONS:
             self[attr] = self[prefix + attr]
 
+    def initialize_on_layout(self):
+        self.rtl = False
+        body_style = window.getComputedStyle(document.body)
+        if body_style.direction is "rtl":
+            self.rtl = True
```
mwgabby-li, line 29: I hope this is relatively robust? document.body is what the handle_rtl_layout was testing for, so it seemed like a good idea to borrow that.

src/pyj/read_book/viewport.pyj
```diff
@@ -12,13 +12,28 @@ class ScrollViewport:
...
+
+    def ltr(self):
+        return not self.rtl
```
mwgabby-li, line 32: I toyed with making rtl private and adding an accessor, but ltr() has been the only one I've needed so far.

src/pyj/read_book/viewport.pyj
```diff
@@ -52,6 +70,29 @@ class ScrollViewport:
...
+            rect = frame.getBoundingClientRect()
+            x += rect.left
+            y += rect.top
+            doc = frame.ownerDocument
+        win = doc.defaultView
+        wx, wy = win.pageXOffset, win.pageYOffset
+        x += wx
+        y += wy
+        if self.rtl:
+            return -x, y
```
mwgabby-li, line 94: This would be the place the transform happens. I removed two copies of viewport_to_document from the codebase, so I consider the reduction of duplicate code a win. I hope you agree!

src/pyj/read_book/viewport.pyj
```diff
@@ -82,7 +123,8 @@ class IOSScrollViewport(ScrollViewport):
         ans = parseInt(raw)
         if isNaN(ans):
             return 0
-        ans *= -1
+        if not self.rtl:
+            ans *= -1
```
mwgabby-li, line 127: Just don't do this invert in RTL mode. Seemed reasonable.

src/calibre/srv/render_book.py
```diff
@@ -639,6 +639,12 @@ def work_priority(name):
     spineq = frozenset(spine)
     landmarks = [l for l in get_landmarks(container) if l['dest'] in spineq]
 
+    page_progression_direction = None
+    try:
+        page_progression_direction = container.opf_xpath('//opf:spine/@page-progression-direction')[0]
+    except IndexError:
+        pass
```
mwgabby-li, line 646: Hope I did it right. This was based on the guidance from your email. Thanks again!

kovidgoyal:
Looks good I will review ASAP

kovidgoyal:
Merged with some minor changes, see commits after the merge commit.

PR #1202, opened 2020-08-01
Fix bug #1889973 custom column of bool type: cannot delete default …

cbhaley:
…value once it's set


cbhaley:
Kovid, if you were really fast you might not have seen the second commit. Sorry about that.

PR #1203, opened 2020-08-05
When updating an OPF, remove custom columns from the target that …

cbhaley:
…don't exist in the source


kovidgoyal:
Is it safe to be not depp copying the other objects user metadata into
this object in remove_stale_user_metadata?

cbhaley:
I thought about that for a bit and decided that because "new" is replacing the object's user metadata dict it is safe and perhaps preferred. Any existing references to "self" dict items will either become dangling references (same as if a copy were made) or references to the same data now in "new".

cbhaley:
set_all_user_metadata() does a shallow copy, so all the self dict references become dangling.

PR #1207, opened 2020-08-12
Vertical RTL Book Reading Support

mwgabby-li:
Added a variety of functions to viewport to allow working based on block and inline direction rather than X and Y.

Changed paged mode and flow mode to be based on block and inline directions, making them agnostic to writing direction.

Added jump_to_cfi function to flow mode code, and use this in iframe. This fixes some issues with CFI jumping in flow mode.

Use self.jump_to_cfi in iframe so that it is based on the current mode.

Removed a redundant self.onscroll call that was causing CFIs to be queried twice on load, doubling the time they took.

Fixed some bugs related to scrolling in flow mode related to a decorator not working with keyword arguments, by removing the keyword arguments.

Fixed some bugs with flow mode visibility anchor function in vertical mode.

CFI changes:
 Renamed some functions and variables to clarify meaning.

 Remove use of exceptions, since they were causing CFI calculation to run about six times more slowly in benchmarks. (Worst-case went from ~6 to ~1 second.)

 Remove forward flag and allow text offset to be equal to node length, and correct for this by adding a flag to the decoded CFI to indicate this.

 Added comments clarifying the results of decoding a CFI.

 Fix bugs in the point (now decode_with_range) function and simplified it.

 Added new functions that get CFI position from decoded CFIs, to replace places where this was being done in slightly different ways.

 Fixed some issues with cfi.pyj's scroll_to that were exposed by vertical writing support, mainly caching and using span's parent to create the start and end ranges, which causes re-creating the range to succeed. Also, don't store the span's bounding rect, since it's not needed anymore.

 Rewrote at_current to do a simplified scan of the viewport in the inline and then block direction.

### DIFF
--- a/src/calibre/ebooks/epub/cfi/parse.py
+++ b/src/calibre/ebooks/epub/cfi/parse.py
@@ -57,6 +57,9 @@ class Parser(object):
     def parse_epubcfi(self, raw):
         ' Parse a full epubcfi of the form epubcfi(path [ , path , path ]) '
         null = {}, {}, {}, raw
+        if not raw:
+            return null
+
         if not raw.startswith('epubcfi('):
             return null
         raw = raw[len('epubcfi('):]
--- a/src/calibre/srv/render_book.py
+++ b/src/calibre/srv/render_book.py
@@ -136,11 +136,15 @@ absolute_font_sizes = {
     'medium': '1rem',
     'large': '1.125rem', 'x-large': '1.5rem', 'xx-large': '2rem', 'xxx-large': '2.55rem'
 }
+nonstandard_writing_mode_property_names = ('-webkit-writing-mode', '-epub-writing-mode')
 
 
 def transform_declaration(decl):
     decl = StyleDeclaration(decl)
     changed = False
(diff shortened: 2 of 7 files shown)


mwgabby-li:
As always, let me know if you need anything from me. I'll be waiting on you for now.

Known issues with vertical books remaining after this commit:
 - Japanese books with furigana have them split across columns in vertical books.
 - Selection toolbar in vertical books is always in the upper-left of the screen rather than next to the selection.
 - Text selection handles in vertical books are not displayed in desktop viewer and too large and unusable on Firefox Mobile.

Once this pull request is merged, in addition to the above, I have a bunch of other small polish items for RTL and vertical text on my to-do list, as well as a bunch of smaller general polish items and some potential optimizations.

kovidgoyal:
I am working on reviewing this, have done about half so far, you can see my changes in the vert branch.

kovidgoyal:
Feel free to comment if some change is unclear/incorrect.

mwgabby-li:
Thanks! I'll take a look and let you know.

kovidgoyal:
merged into py3

PR #1213, opened 2020-08-28
Support Selection Handles on Vertical/RTL Books

mwgabby-li:
Hi Dr. Goyal - This change makes selection handles work better in RTL/vertical mode. Let me know what you think, I really wanted to get this working because it seemed like the biggest missing part for first-class support for different writing modes.
If the changes are too major for you to be comfortable with, I understand. Because of the bug with collapsed text ranges in vertical books, I had to significantly modify how the selection extents are obtained.

Known Issues:
 - Some selection edge cases in RTL books produce handles at the corner of the screen.
 - Selections that don't match the dominant text direction in a book will have issues. (Such as trying to select a section of English text in a Hebrew book.)
 - Dragging to go to the next or previous page doesn't work correctly in RTL books.
 - Selection dragging across images is very slow in vertical books in mobile Chrome.

Changes:

 - Pass vertical/rtl mode into selection code.
 - Added new image for vertical selection handle.

selection_bar.pyj:
 - Made code agnostic to text direction.
 - Changed names to start and end rather than left and right to reflect new behavior.
 - Track vertical/rtl state from selection message.
 - Handle selection position modified to support all possible text orientations.
 - Switch to vertical selection handle in vertical mode.
 - Cap selection size at 60px to avoid comically large (and unusable) handles when selecting large element, such as an image.

select.pyj:
 - Improved selection behavior when selected range extents are on elements by search the node's DOM tree for something with a reasonable bounding box.
 - To work around bugs with collapsed range rects and vertical text, use character bounding boxes in most cases.
 - Add width to selection range extents.


kovidgoyal:
This is now merged. And FYI development has shifted back to the master branch.

PR #1222, opened 2020-09-07
Correct imports for OZON.ru plugin to be able run it using Python 3.x

sergeyklay:
For more see: https://bugs.launchpad.net/calibre/+bug/1894751

### DIFF
--- a/src/calibre/ebooks/metadata/sources/ozon.py
+++ b/src/calibre/ebooks/metadata/sources/ozon.py
@@ -71,7 +71,10 @@ class Ozon(Source):
     # }}}
 
     def create_query(self, log, title=None, authors=None, identifiers={}):  # {{{
-        from urllib import quote_plus
+        try:
+            from urllib import quote_plus
+        except ImportError:
+            from urllib.parse import quote_plus
 
         # div_book -> search only books, ebooks and audio books
         search_url = self.ozon_url + '/?context=search&group=div_book&text='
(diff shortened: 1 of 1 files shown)


src/calibre/ebooks/metadata/sources/ozon.py
```diff
@@ -71,7 +71,10 @@ def get_book_url(self, identifiers):  # {{{
     # }}}
 
     def create_query(self, log, title=None, authors=None, identifiers={}):  # {{{
-        from urllib import quote_plus
+        try:
+            from urllib import quote_plus
+        except ImportError:
+            from urllib.parse import quote_plus
```
eli-schwartz, line 77: Can you please first try the python3 version, then fall back on the python2 version?

src/calibre/ebooks/metadata/sources/ozon.py
```diff
@@ -116,10 +119,20 @@ def create_query(self, log, title=None, authors=None, identifiers={}):  # {{{
 
     # }}}
 
+    def unescape(self, s): # {{{
+        try:
+            from HTMLParser import HTMLParser
+            h = HTMLParser()
+            return h.unescape(s)
+        except ImportError:
+            from html import unescape
```
eli-schwartz, line 128: Likewise here.

eli-schwartz (changes requested):

eli-schwartz:
Also it should be possible to test the plugin using
```
calibre-debug src/calibre/ebooks/metadata/sources/ozon.py
```

but it looks like the current data doesn't find anything, so it would be nice to get the self-test working again.

PR #1223, opened 2020-09-08
py3: add missing try/except in metadata sources (ozon.ru)

eli-schwartz:
urllib.quote_plus was missed in this file during commit e0205790b943227ad86a00ca975f078b624b164a and HTMLParser never got checked since this is the only use.

Fixes #1894751 [In Python 3.0, the HTMLParser module has been renamed to html.parser](https://bugs.launchpad.net/bugs/1894751)

...

I could not properly test this, since the ozon.ru test_identify_plugin runner fails early due to no results found and I don't have suitable replacement data to search for with it. But it gets far enough to notice that at least, rather than failing with import errors.


eli-schwartz:
Right... #1222 exists already, my bad

PR #1233, opened 2020-09-24
Update opds cover and thumb rel to match OPDS 1.0 and higher

mknjc:
OPDS 0.9 used /cover and /thumbnail as link rel for covers and thumbnails but this was changed with ODPS 1.0
See: https://specs.opds.io/opds-1.0.html#Artwork_Relations


kovidgoyal:
That would break clients that expect the existing links. I suggest adding extra elements instead.

PR #1237, opened 2020-09-28
Callable saved GPM templates

cbhaley:


kovidgoyal:
I'd rather not add another top level entry to Preferences. Instead make it a tab under Template functions.

cbhaley:
Changed to a tabbed dialog. Also embedded the template dialog.

kovidgoyal:
Looks generally OK but I will review it sometime after the next release.

Couple of quick comments:

1) The stored templates tab should probably be second not first
2) Move the scroll bar inside each individual tab, otherwise one large
tab means all the other tabs also take up that height which makes them
hard to use on smaller screen. Basically wrap individual tabs in
QScrollWidgets instead of wraprring th entire tab widget in a scroll.

cbhaley:
Re:
> The stored templates tab should probably be second not first
I disagree. I think that stored templates will be used far more often than template functions. I want the shortcut to go directly to that tab, and the only way to do that is to have it be the first tab.

> scroll bars
OK.

kovidgoyal:
OK, your choice on (1)

cbhaley:
Scrolling tabs pushed.

cbhaley:
Kovid, overnight I thought of 2 improvements I will make:
1) remove the 'call' keyword, making stored templates usable in single function mode
2) cache the parsed stored template, improving performance in single function and template program modes
I will comment here when I finish, which might be Saturday.

kovidgoyal:
Sure, no rush.

cbhaley:
Well, that was a lot easier then I expected. Done.

PR #1242, opened 2020-10-07
Fix Douban Metadata plugin

yingziwu:
Make Douban Metadata plugin work on python3.


kovidgoyal:
I'm afraid I am not including douban in the core calibre source anymore.
As it breaks too often and I cant maintain it myself. You are most
welcome to make it into a third party plugin however. If you wish to do
so, see https://www.mobileread.com/forums/forumdisplay.php?f=237

yingziwu:
Thanks for your advices.

PR #1245, opened 2020-10-11
Enhancement #1899341 (partial): Add "Search for" to category editor …

cbhaley:
…context menu


cbhaley:
I moved the get_gui import to avoid the "partially initialized" problem

PR #1246, opened 2020-10-11
Still save keyboard shortcuts to actions that aren't visible in current library.

JimmXinu:
When testing different plugins in different libraries I noticed that my keyboard shortcuts in completely different plugins would sometimes disappear.  I've seen it for years.

What happens is that some plugins (Reading List, View Manager & FanFicFare at least) change what menu actions are available depending on library.  Changing _any_ keyboard shortcut discards shortcuts to menu actions that aren't visible/available in the current library.

This change preserves those shortcuts by only removing shortcuts explicitly back to 'default'.

This could conceivably cause shortcuts for removed plugins to hang around indefinitely.  But then, I thought that was a 'feature'; if you reinstall a plugin later, it still has your previous shortcuts?

### DIFF
--- a/src/calibre/gui2/actions/__init__.py
+++ b/src/calibre/gui2/actions/__init__.py
@@ -154,7 +154,7 @@ class InterfaceAction(QObject):
             bn = self.interface_action_base_plugin.name
         return 'Interface Action: %s (%s)'%(bn, self.name)
 
-    def create_action(self, spec=None, attr='qaction', shortcut_name=None):
+    def create_action(self, spec=None, attr='qaction', shortcut_name=None, persist_shortcut=False):
         if spec is None:
             spec = self.action_spec
         text, icon, tooltip, shortcut = spec
@@ -191,7 +191,8 @@ class InterfaceAction(QObject):
                     self.gui.keyboard.register_shortcut(self.unique_name + ' - ' + attr,
                         shortcut_name, default_keys=keys,
                         action=shortcut_action, description=desc,
-                        group=self.action_spec[0])
+                        group=self.action_spec[0],
+                        persist_shortcut=persist_shortcut)
                 except NameConflict as e:
                     try:
                         prints(unicode_type(e))
@@ -216,7 +217,7 @@ class InterfaceAction(QObject):
         return action
 
     def create_menu_action(self, menu, unique_name, text, icon=None, shortcut=None,
-            description=None, triggered=None, shortcut_name=None):
+            description=None, triggered=None, shortcut_name=None, persist_shortcut=False):
         '''
         Convenience method to easily add actions to a QMenu.
         Returns the created QAction. This action has one extra attribute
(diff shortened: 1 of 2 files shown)


kovidgoyal:
The problem with this is that if some shortcut is removed/renamed then the original lingers forever in the users settings, unless she explicitly removes it. 

A better fix I thikn would be to add some data to the shortcut definition, that indicates it should persist, even in the absence of its associated action. The plugins can then set that to True when registering the shortcut and the commit function can preserve such shortcuts.

JimmXinu:
While that's true, the same can be said for plugin settings both in the DB and in files.  We consider it a feature that your settings are still there if you reinstall a plugin again later.

Plus, marking some shortcuts persistent doesn't eliminate the issue, it only reduces the scope of it.  Even if a user assigned every possible key, it's still limited to hundreds of entries at most.

I will investigate adding a 'persist' flag of some kind to actions, but IMO, the benefits: reducing the number of possible dangling shortcuts; are out weighted by the drawbacks: more complexity and plugins/actions that don't implement it will still suffer from mysterious disappearing shortcuts.

Testing this some more, I find that the existing shortcuts code doesn't have a way to forceably take a shortcut that's already been assigned(that would be nice).  But it only triggers the _ERROR: Already Assigned_ dialog if the shortcut action is still available (and mapped to that key).  So previously assigned shortcuts to 'missing' actions don't prevent the user using that key for something else.  

One key can even be assigned to different actions in different libraries.  It does, however, raise the possibility of a key being assigned to more than one action for some libraries when it is assigned to a common action in a different library.  'First' one wins, apparently; not sure what the ordering is.  A 'persist' flag would not help with that.  And if a user goes to assign it again, they will be told what it's already assigned to, so that's less mysterious than current practice.

JimmXinu:
I've implemented a `persist_shortcut` flag as requested.

Because the existing json maps unique action name directly to a list of keys, I don't see a good back-compat way to add a flag to the existing `map` `dict()`:

existing `shortcuts/main.json`:
```
{
  "map": {
    "Interface Action: View Manager (View Manager) : menu action : 2": [
      "Ctrl+2"
    ]
  }
}
```
I've added a parallel `options_map` hash which has a `persist_shortcut` flag:
```
{
  "map": {
    "Interface Action: View Manager (View Manager) : menu action : 2": [
      "Ctrl+2"
    ]
  },
  "options_map": {
    "Interface Action: View Manager (View Manager) : menu action : 2": {
      "persist_shortcut": true
    }
  }
}
```
This is more than a simple name->bool mapping for forward-compat in case other options are needed.  I have some vague ideas about putting library uuid in there too, but nothing concrete yet.

Shortcuts are only added to `options_map` when there is a hotkey for them.  They are only removed when a shortcut is explicitly flagged `set_to_default`.  They are *not* removed from `options_map` when a shortcut is unregistered because the same List or View name may be used in a different library.

Please let me know what you think.  Personally, I thought it was a cleaner solution the first time(aba21a2b1b4d4965124f206690a8989f8511bfc4).

kovidgoyal:
Thanks will look at it soon, been rather busy with calibre 5 related stuff the last few days.

PR #1251, opened 2020-10-14
Add option to tag browser to combine 'first letters' together if there …

cbhaley:
…are fewer items under the letter than X.

### DIFF
--- a/src/calibre/gui2/__init__.py
+++ b/src/calibre/gui2/__init__.py
@@ -134,6 +134,7 @@ def create_defs():
     defs['font'] = None
     defs['tags_browser_partition_method'] = 'first letter'
     defs['tags_browser_collapse_at'] = 100
+    defs['tags_browser_collapse_fl_at'] = 0
     defs['tag_browser_dont_collapse'] = []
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I have merged, nice idea. Do you think the default should be made
non-zero? Perhaps, 5? Or maybe in the release after the next to give it
time for any problems to shake out.

cbhaley:
I thought about setting the default to 5 (or so) but chickened out. That would change calibre's behavior for a lot of people. Is there a way to change it for new installs vs upgrades?

kovidgoyal:
No, I dont think so, at least no existing mechanism.

On Wed, Oct 14, 2020 at 08:53:04AM -0700, Charles Haley wrote:
> I thought about setting the default to 5 (or so) but chickened out. That would change calibre's behavior for a lot of people. Is there a way to change it for new installs vs upgrades?
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/1251#issuecomment-708494581

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

PR #1254, opened 2020-10-15
Updates and cleanups of the template documentation.

cbhaley:
I don't know if I got the formatting right. In particular I moved the function reference to the end, which I hope doesn't break anything.


cbhaley:
I added a commit to remove a doubled "Collapse X" in the tag browser when right-clicking on the top-level item

PR #1258, opened 2020-10-17
Fix enhancement #190009: Width of custom date columns.

cbhaley:
There are two changes:
1) the 'today' and 'clear' buttons for date columns are now right adjusted.

2) Labels are now fixed width. Word wrap is disabled. There are two new tweaks to
set the fixed widths for edit metadata single and bulk.


kovidgoyal:
Maybe use QFontMetrics::averageCharWidth() for performance rather than calculating the width of a string of n's. Or if you do want to use a string of n's them maybe use m instead as that corresponds to the tradition 'em' unint.

cbhaley:
Changed to use averageCharWidth(). This is smaller than an 'n' so I bumped the default width in the tweak up by 2.

PR #1259, opened 2020-10-17
Make the tag browser expand and collapse menu items consistent.

cbhaley:
1) add separate lines to expand item and expand item and children
2) add collapse lines for all parent nodes, not just the immediate parent and the category root


kovidgoyal:
This is making the menu a little too long, maybe have a submenu with the
various expand/collapse options, either one for both or two submenus one
for expand and one for collapse

cbhaley:
Submenu added, along with some icons.

PR #1261, opened 2020-10-18
Two independent commits:

cbhaley:
1) ensure that at startup the booklist gets the keyboard focus instead of quickview
2) Allow the tag browser to get keyboard focus


cbhaley:
BTW: I sorted the list of items in look_feel.ui, which is why there are so many changes.

PR #1263, opened 2020-10-19
Win64

shockergit:
Can You add an option to add background images to  epub viewing. eg i have a wooden image that i can use as a background for my book while reading


kovidgoyal:
There already is such an option look in the viewer preferences. And pull requests are not an appropriate way to ask for features. https://calibre-ebook.com/bugs

PR #1265, opened 2020-10-22
Bug 1900921: various problems with enumerated types.

cbhaley:
Plus an enhancement: add an editor to change the list of permitted values.


cbhaley:
I recognize that I am doing something odd by modifying the display when I save it to the DB. This has worked for a long time (years) with composite templates. However, if you don't want that then I will change it to require a restart.

Alternatively I could change the edit/create custom column dialog to use this new editor. This feels more clunky to me.

kovidgoyal:
On Thu, Oct 22, 2020 at 09:00:20AM -0700, Charles Haley wrote:
> I recognize that I am doing something odd by modifying the display when I save it to the DB. This has worked for a long time (years) with composite templates. However, if you don't want that then I will change it to require a restart.

It's fine, I cant think of any real issues. If any come up, we can
revisit.

PR #1267, opened 2020-10-28
Improve Linux/XDG AppStream file generation

ntninja:
Changes to the generated AppStream XML files:

  * Added OARS (Open Age Rating Service) metadata
  * Added release history to the main Calibre GUI AppStream file based on the contents of `Changelog.yaml` with a layout similar to that on the website

These changes combined make the generated main AppStream file pass the `appstreamcli validate` check, reducing the need to patch it downstream.

(By product of the work at https://github.com/flathub/com.calibre_ebook.calibre/pull/53.)

### DIFF
--- a/src/calibre/linux.py
+++ b/src/calibre/linux.py
@@ -1200,8 +1200,16 @@ def write_appdata(key, entry, base, translators):
         E.metadata_license('CC0-1.0'),
         E.project_license('GPL-3.0'),
         E.summary(entry['summary']),
+        E.content_rating(
+            # Information Sharing: Using any online API, e.g. a user-counter
+            # Details at https://calibre-ebook.com/dynamic/calibre-usage .
+            E.content_attribute('mild', id='social-info'),
+            # In-App Purchases: Users are encouraged to donate real money, e.g. using Patreon
+            E.content_attribute('mild', id='money-purchasing'),
+            type='oars-1.1'
+        ),
         description,
-        E.url('https://calibre-ebook.com', type='homepage'),
+        E.url('https://calibre-ebook.com/', type='homepage'),
         screenshots,
         type='desktop'
     )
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Looks mostly OK, however, I dont want to add a YAML parser to calibre's
dependencies. To that end I will change the changelog format to
something that can be parsed using just the stdlib.

kovidgoyal:
I have moved the changelog to use a plain text format. A parser for it that returns the eat same structure as the yaml parser is in, setup/changelog.py. You can execute it using runpy.run_path. Also note that setup_desktop_integration is called from the linux binary builds as well, where the changelog source is not available.

kovidgoyal:
And I ended up making the changelog available as a resource, so all you need to do is

```
json.loads(P('changelog.json', data=True))
```


Note that this structure will not have dates as date objects but as strings.

ntninja:
Finally was able to test this, release tarball != git checkout caused quite some issues there…

Installation logic now loads the JSON using the command you suggested and generates valid AppStream XML from that.

eli-schwartz:
https://github.com/kovidgoyal/calibre/blob/master/INSTALL.rst#build

I documented this quite some time ago. :D 

(Many projects require more for a git checkout than for a dist tarball, e.g. `autoreconf` as a prerequisite before running .`/configure`.)

kovidgoyal:
I havent paid much attention to XDG/Appstream, so some questions:

1) What's desktop-id for?
2) Why is the name of the file being changed to metainfo.xml from appdata.xml?

eli-schwartz:
According to https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location the rename is for unrelated reasons, that reason being "they decided to rename it at random, and now appdata.xml still works fine but is considered to be legacy".

ntninja:
> What's desktop-id for?

It associates the AppStream file with the installed .desktop file so that the installed application can be immediately launched from the AppStore (GNOME Software, Plasma Discover) after installation.

> Why is the name of the file being changed to metainfo.xml from appdata.xml?

See @eli-schwartz's answer.

Note that for the Flatpak, I rename all these files to include their [AppID](https://github.com/flathub/com.calibre_ebook.calibre/blob/ebb51dcc8ae147d4ff771d560c9e39fe4f0b7860/fixup-exports.py#L135-L138) after installation. While it makes the installed file paths uglier to look at, Flatpak sadly requires this to prevent collisions of installed data files between applications.

Should I propose these renames as well, or do prefer to keep the old names?

kovidgoyal:
On Fri, Oct 30, 2020 at 02:52:56AM -0700, Alexander Schlarb wrote:
 
> Should I propose these renames as well, or do prefer to keep the old names?

I prefer to avoid as much pointless churn as possible.

PR #1268, opened 2020-10-29
Replace unused legacy KDENotifier with XDPNotifier for sandboxed Linux application contexts

ntninja:
This is part of my ongoing work at reducing the number of permissions required for the Calibre Flatpak package (likely useful for Snap too).

The new notification code is programmed to conform to the interface described at https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.Notification.

As can be seen in the code, that interface does not expose the ability to set a timeout. To remedy this I set a timeout on the Calibre side and manually cancel the notification after the given number of milliseconds. **However**, practical experimentation in KDE shows that notifications submitted via this API have a default timeout of 10 seconds and display a visual cue how long it will take until they expire. Additionally, interacting with the popup (mouse-over) will reset the timer and with Calibre cancelling the notification after the default 5 seconds the popup will appear to  disappear for no apparent reason and contrary to the other cues given by the UI.
Based on this, I'd be inclined to remove the timeout handling altogether and just let the DE decide for how long it wishes to display the notification. Let me know whatever you prefer.

None of this affects unconfined builds of Calibre (website download) as these continue to prefer the older FreeDesktop Notification API.

OT: Is there any plan to eventually switch away from *dbus-python* for DBus interaction to something maintained? Would you accept patches for such a thing? If yes, to which API should the switch be (QDBus, PyDBus, Jeepney)?

### DIFF
--- a/src/calibre/gui2/notify.py
+++ b/src/calibre/gui2/notify.py
@@ -49,24 +49,6 @@ class DBUSNotifier(Notifier):
             prints(server, 'found' if self.ok else 'not found', 'in', '%.1f' % (time.time() - start), 'seconds')
 
 
-class KDENotifier(DBUSNotifier):
-
-    SERVICE = 'org.kde.VisualNotifications', '/VisualNotifications', 'org.kde.VisualNotifications'
-
-    def __call__(self, body, summary=None, replaces_id=None, timeout=0):
-        if replaces_id is None:
-            replaces_id = self.dbus.UInt32()
-        event_id = ''
-        timeout, body, summary = self.get_msg_parms(timeout, body, summary)
-        try:
-            self._notify.Notify('calibre', replaces_id, event_id, self.ICON, summary, body,
-                self.dbus.Array(signature='s'), self.dbus.Dictionary(signature='sv'),
-                timeout)
-        except:
-            import traceback
-            traceback.print_exc()
-
-
 class FDONotifier(DBUSNotifier):
 
     SERVICE = 'org.freedesktop.Notifications', '/org/freedesktop/Notifications', 'org.freedesktop.Notifications'
@@ -88,7 +70,7 @@ def get_dbus_notifier():
     import dbus
     session_bus = dbus.SessionBus()
     names = frozenset(session_bus.list_names())
-    for srv in KDENotifier, FDONotifier:
+    for srv in (FDONotifier,):
         if srv.SERVICE[0] in names:
             ans = srv(session_bus)
             if ans.ok:
--- a/src/calibre/gui2/notify.py
+++ b/src/calibre/gui2/notify.py
@@ -30,7 +30,13 @@ class Notifier(object):
 
 class DBUSNotifier(Notifier):
 
+    #XXX: Really, this should instead just send the themable icon name
+    #
+    # Doing that however, requires Calibre to first be converted to use
+    # its AppID everywhere and then we still need a fallback for portable
+    # installations.
     ICON = I('lt.png')
+    ICON_DATA = I('lt.png', data=True)
 
     def __init__(self, session_bus):
         self.ok, self.err = True, None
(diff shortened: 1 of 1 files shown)


src/calibre/gui2/notify.py
```diff
@@ -28,9 +29,20 @@ def __call__(self, body, summary=None, replaces_id=None, timeout=0):
...
+
 class DBUSNotifier(Notifier):
 
+    #XXX: Really, this should instead just send the themable icon name
+    #
+    # Doing that however, requires Calibre to first be converted to use
+    # its AppID everywhere and then we still need a fallback for portable
+    # installations.
     ICON = I('lt.png')
+    ICON_DATA = read_file_contents(ICON)
```
eli-schwartz, line 45:
```
I('lt.png', data=True)
```

And there is no need to create a module level function just to read one file at one call site.

eli-schwartz (changes requested):

kovidgoyal:
On Wed, Oct 28, 2020 at 05:09:39PM -0700, Alexander Schlarb wrote:
> This is part of my ongoing work at reducing the number of permissions required for the Calibre Flatpak package (likely useful for Snap too).
> 
> The new notification code is programmed to conform to the interface described at https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.Notification.
> 
> As can be seen in the code, that interface does not expose the ability to set a timeout. To remedy this I set a timeout on the Calibre side and manually cancel the notification after the given number of milliseconds. **However**, practical experimentation in KDE shows that notifications submitted via this API have a default timeout of 10 seconds and display a visual cue how long it will take until they expire. Additionally, interacting with the popup (mouse-over) will reset the timer and with Calibre cancelling the notification after the default 5 seconds the popup will appear to  disappear for no apparent reason and contrary to the other cues given by the UI.
> Based on this, I&#39;d be inclined to remove the timeout handling altogether and just let the DE decide for how long it wishes to display the notification. Let me know whatever you prefer.

If the backend doesn't support timeouts, then it's fine to ignore
timeouts. I'm not a fan of hacking around things like that. 

> 
> None of this affects unconfined builds of Calibre (website download) as these continue to prefer the older FreeDesktop Notification API.
> 
> OT: Is there any plan to eventually switch away from *dbus-python* for DBus interaction to something maintained? Would you accept patches for such a thing? If yes, to which API should the switch be (QDBus, PyDBus, Jeepney)?

No plans as such, it's used fairly extensively, for tray icons and app
menus and as I use neither of those things myself, tends to be
relatively less maintained/tested. That said, I am in principle OK with
replacing it with QtDBus, provided I dont have to do the work/testing.
-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

ntninja:
> If the backend doesn't support timeouts, then it's fine to ignore timeouts. I'm not a fan of hacking around things like that.

I've dropped the timeout handling code and replaced it with a comment on why the timeout parameter is ignored.

>> OT: Is there any plan to eventually switch away from *dbus-python* for DBus interaction to something maintained? Would you accept patches for such a thing? If yes, to which API should the switch be (QDBus, PyDBus, Jeepney)?
>
> No plans as such, it's used fairly extensively, for tray icons and app menus and as I use neither of those things myself, tends to be relatively less maintained/tested. That said, I am in principle OK with replacing it with QtDBus, provided I dont have to do the work/testing.

Noted. I'll keep that in mind.

eli-schwartz:
Can you retain the ordering of the *Notifier classes? Right now if you diff the before and after for the pull request itself, it looks a bit noisy even though if you look at each commit, FDONotifier isn't being changed.

ntninja:
@eli-schwartz: The ordering is on purpose as `FDONotifier` is superior in functionality to `XDPNotifier` and the former is guaranteed to not exists in sandbox context (unless a special exception is added), while the later may exist outside of sendbox context, but AFAIK may not work correctly there.

eli-schwartz:
The iteration order of use (`for srv in (FDONotifier, XDPNotifier):`) matters. The order of class definition (`class FDONotifier(DBUSNotifier):`) doesn't matter except for prettiness, but it was a minor issue and too late now as it's been merged.

ntninja:
@eli-schwartz: Oh sorry, I thought you meant the iterator. :slightly_smiling_face: 

Yeah, I really should have kept the order of classes, now that you mention it.

PR #1270, opened 2020-10-31
Several fixes from various bug reports and MR discussions

cbhaley:
Much of this work was done before I 'disappeared'. :)

They are individual commits that, I hope, makes it easier for you to see why changes were made.


kovidgoyal:
Couple of questions:

1) Why is F2 disabled for renaming things?

2) there is already a focus indications, why do we need a special background color? See attched screenshot.
Screenshot_20201101_084336

kovidgoyal:
Regarding not being able to ask if virtual libraries are to be respected, you can implement that with F2 by using a custom editor in the delegate

cbhaley:
I think the problem comes from using alternating row colors. The default focus indication for the TB is very hard to see, especially in dark mode. In light mode I can search for it and eventually see it. In dark mode I must move up and down to find the focus. The attachments show what I mean.

I have no problem at all adding an option to keep the current behavior. I also have no problem changing the focus color to something closer to the booklist cell focus color.
Dark_focus_booklist
Dark_focus_tb
Light_focus_booklist
Light_focus_tb

cbhaley:
In my last comment it is "Languages" that has the focus in the TB.

kovidgoyal:
A better fix for that would be to improve the focus highlighting color
in the style. I will look into that later, could you remove the
stylesheet based workaround for now and I will merge the rest.

cbhaley:
> Regarding not being able to ask if virtual libraries are to be respected, you can implement that with F2 by using a custom editor in the delegate

I don't see how this is better than simply disabling the edit keys.

If the edit triggers are disabled then to edit an item the user must focus the item, press the "Menu" key, then either hit return or down-arrow then return depending on whether VLs are to be respected.

If the edit triggers are enabled then the user must focus the item, then press F2. At this point, as far as I can see, calibre must ask the user whether or not to respect vls using a question dialog (or some other way I haven't thought of), then open the editor widget as it does today. This is more work for the user than using the existing context menu.

cbhaley:
> A better fix for that would be to improve the focus highlighting color
> in the style. I will look into that later, could you remove the
> stylesheet based workaround for now and I will merge the rest.

Will do.

kovidgoyal:
On Sun, Nov 01, 2020 at 05:28:54AM -0800, Charles Haley wrote:
> > Regarding not being able to ask if virtual libraries are to be respected, you can implement that with F2 by using a custom editor in the delegate
> 
> I don't see how this is better than simply disabling the edit keys.

It's an expectation/known pattern. And many computers dont have menu
keys, mine for instance. IIRC shift+f10 or something is supposed ot do
the same but i never remember that.

> 
> If the edit triggers are disabled then to edit an item the user must focus the item, press the "Menu" key, then either hit return or down-arrow then return depending on whether VLs are to be respected.
> 
> If the edit triggers are enabled then the user must focus the item, then press F2. At this point, as far as I can see, calibre must ask the user whether or not to respect vls using a question dialog (or some other way I haven't thought of), then open the editor widget as it does today. This is more work for the user than using the existing context menu.

So I dont actually recall what the issue with respecting the VL is, but
I would imagine that it could be asked after editing is complete, if the
item would conflict with the VL, if any.

cbhaley:
I have pushed changes to remove the stylesheet changes.

Re F2: the issue is that if a VL is in effect the user must tell us whether the rename is library-wide or only for books in the VL. The question isn't needed if there isn't a VL.

I will look for a way to ask that question when F2 is pressed. At first blush a question box where ENTER says "yes" and ESC says "no" is easiest for the user to use, adding only one keypress. I will submit my proposition as a new PR, assuming you merge and close this one.

PR #1271, opened 2020-11-01
Re-enable the edit keys in the tag browser.

cbhaley:
If required (in a VL) ask the user whether to apply the rename in the entire or virtual library.

### DIFF
--- a/src/calibre/gui2/tag_browser/view.py
+++ b/src/calibre/gui2/tag_browser/view.py
@@ -23,7 +23,7 @@ from calibre.gui2.tag_browser.model import (TagTreeItem, TAG_SEARCH_STATES,
         TagsModel, DRAG_IMAGE_ROLE, COUNT_ROLE)
 from calibre.gui2.widgets import EnLineEdit
 from calibre.gui2 import (config, gprefs, choose_files, pixmap_to_data,
-                          rating_font, empty_index)
+                          rating_font, empty_index, question_dialog)
 from calibre.utils.icu import sort_key
 from calibre.utils.serialize import json_loads
 from polyglot.builtins import unicode_type, range, zip
@@ -31,12 +31,13 @@ from polyglot.builtins import unicode_type, range, zip
 
 class TagDelegate(QStyledItemDelegate):  # {{{
 
-    def __init__(self, *args, **kwargs):
-        QStyledItemDelegate.__init__(self, *args, **kwargs)
+    def __init__(self, tags_view):
+        QStyledItemDelegate.__init__(self)
         self.old_look = False
         self.rating_pat = re.compile(r'[%s]' % rating_to_stars(3, True))
         self.rating_font = QFont(rating_font())
         self.completion_data = None
+        self.tags_view = tags_view
 
     def draw_average_rating(self, item, style, painter, option, widget):
         rating = item.average_rating
@@ -128,6 +129,15 @@ class TagDelegate(QStyledItemDelegate):  # {{{
         self.completion_data = data
 
     def createEditor(self, parent, option, index):
+        item = self.tags_view.model().get_node(index)
+        item.use_vl = False
+        if self.tags_view.model().get_in_vl():
+            if question_dialog(self.tags_view, _('Rename in Virtual library'), '<p>' +
+                               _('Do you want this rename to apply only to books '
+                                 'in the current Virtual library?') + '</p>',
+                               yes_text=_('Yes, apply only in VL'),
+                               no_text=_('No, apply in entire library')):
+                item.use_vl = True
         if self.completion_data:
             editor = EditWithComplete(parent)
             editor.set_separator(None)
@@ -170,7 +180,7 @@ class TagsView(QTreeView):  # {{{
         self.setTabKeyNavigation(True)
         self.setAnimated(True)
         self.setHeaderHidden(True)
-        self.setItemDelegate(TagDelegate(self))
+        self.setItemDelegate(TagDelegate(tags_view=self))
         self.made_connections = False
         self.setAcceptDrops(True)
         self.setDragEnabled(True)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I have merged. Note that on my linux machine clicking on a tag causes
the search to be executed and focus is lost from the tag browser. Not
sure if this is intended or not.

cbhaley:
> Note that on my linux machine clicking on a tag causes
> the search to be executed and focus is lost from the tag browser. Not
> sure if this is intended or not.

This was the lesser of three evils.

Evil 1: restore focus after a click. Clicking on an item sets the focus to that item. Next the toggle is done. The problem: the toggle sees that the focus is on the TB so "restores" it. The user never actually asked to move the focus but is now required to move the focus back to the book list. Not good. Event order makes it difficult, if not impossible, to determine whether the toggling-click moved the focus or not.

Evil 2: use setFocusPolicy(Qt.TabFocus) to prevent setting the focus when the node is clicked. Then we can regrab the focus in toggle() knowing that the user caused the focus. This works, but it has the side effect that one *cannot* set the focus by clicking. If implemented the user would be able to give focus to the TB only with the shortcut.

Evil 3: lose the focus when you click instead of using the keyboard.  This is where I ended up.

I thought Evil 1 was the worst choice. Evil 3 seemed slightly preferable than Evil 2, but I could be wrong. What do you think?

kovidgoyal:
On Sun, Nov 01, 2020 at 08:01:40AM -0800, Charles Haley wrote:
> > Note that on my linux machine clicking on a tag causes
> > the search to be executed and focus is lost from the tag browser. Not
> > sure if this is intended or not.
> 
> This was the lesser of three evils.
> 
> Evil 1: restore focus after a click. Clicking on an item sets the focus to that item. Next the toggle is done. The problem: the toggle sees that the focus is on the TB so "restores" it. The user never actually asked to move the focus but is now required to move the focus back to the book list. Not good. Event order makes it difficult, if not impossible, to determine whether the toggling-click moved the focus or not.
> 

Why does the toggle have to restore focus? Seems to me we could set the
focus policy to nofocus instead when the setting for no focus is set.

cbhaley:
> Why does the toggle have to restore focus? Seems to me we could set the focus policy to nofocus instead when the setting for no focus is set.

I don't understand. For this question, setting the focus policy to noFocus is no different from setting it to tabFocus. In both cases clicks don't transfer the focus to the TB. (FWIW: I thought noFocus blocked *all* focus transfers. Turns out it blocks only implicit transfers. Explicit transfers work even if noFocus is set.)

The overarching goal is to restore the focus to the TB after a RETURN causes a toggle. By definition the TB haa the focus before the RETURN so we know we want to do the restore. The current codebase does this.

You pointed out that clicking on a focused node caused that node to toggle and lose the focus. Fixing this (restoring the focus to the node after the click) requires either a) knowing that the node had the focus before the click, or b) knowing that clicks never transfer the focus. I can't find any way to know (a). Preventing click-to-focus (b) solves the problem by providing the required certainty; the TB knows that if it has focus the user must have caused it, in which case the toggle restores the focus.

As far as I can tell, our choice is either:
- prevent clicks from transferring focus. We then can use focused/not-focused to choose whether to restore the focus on click. The downside: Only the shortcut can be used to transfer focus.
- Allow clicks to transfer focus. In this case I think we must hand focus back to the booklist if the node is clicked because we can't know if it was the click that caused the TB to get the focus. This is what the current codebase does.

I am good either way. Do you want me to push an implementation of the first choice for you to play with?

kovidgoyal:
So here is what I meant:

1) When the option to give the tag browser focus is enabled:
  - Clicking on the tag browser should always give it focus, even if a
    node is toggled
2) When the option to give the tag browser focus is disabled:
  - Clicking on the tag browser should never give it focus, and the
    shortcut to give it focus should have no effect.


So the tag browser should use Qt.ClickFocus|Qt.StringFocus when the
option is enabled and Qt.NoFocus when it is not. You will need to add
some code to refresh_gui() in preferences/look_feel.py to do this when
the option is changed.

To achieve (1) we would additionally need to have the toggle function or nodes not
give away focus when tag browser focus is enabled.

On Sun, Nov 01, 2020 at 08:57:23AM -0800, Charles Haley wrote:
> > Why does the toggle have to restore focus? Seems to me we could set the focus policy to nofocus instead when the setting for no focus is set.
> 
> I don't understand. For this question, setting the focus policy to noFocus is no different from setting it to tabFocus. In both cases clicks don't transfer the focus to the TB. (FWIW: I thought noFocus blocked *all* focus transfers. Turns out it blocks only implicit transfers. Explicit transfers work even if noFocus is set.)
> 
> The overarching goal is to restore the focus to the TB after a RETURN causes a toggle. By definition the TB haa the focus before the RETURN so we know we want to do the restore. The current codebase does this.
> 
> You pointed out that clicking on a focused node caused that node to toggle and lose the focus. Fixing this (restoring the focus to the node after the click) requires either a) knowing that the node had the focus before the click, or b) knowing that clicks never transfer the focus. I can't find any way to know (a). Preventing click-to-focus (b) solves the problem by providing the required certainty; the TB knows that if it has focus the user must have caused it, in which case the toggle restores the focus.
> 
> As far as I can tell, our choice is either:
> - prevent clicks from transferring focus. We then can use focused/not-focused to choose whether to restore the focus on click. The downside: Only the shortcut can be used to transfer focus.
> - Allow clicks to transfer focus. In this case I think we must hand focus back to the booklist if the node is clicked because we can't know if it was the click that caused the TB to get the focus. This is what the current codebase does.
> 
> I am good either way. Do you want me to push an implementation of the first choice for you to play with?
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/1271#issuecomment-720118633

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

cbhaley:
Ah, that is the option I thought you wouldn't want. :)

I will submit it as a new PR

PR #1273, opened 2020-11-02
polyglot: don't try to import plistlib.Data

sgn:
plistlib.Data is deprecated from Python 3.4 and fully removed in Python
3.9.

Skimming over the code, it's look like nothing uses this data structure.

Let's remove it.


eli-schwartz:
https://github.com/kovidgoyal/calibre/commit/764e8bff7e77ef83362c03211fdd08ee1f625b8d#diff-06647ece60517479fdc400725654db46c4893ab75c08ed733da850c80f29366f

It was incorrectly retained from the py2 branch when removing py2 code.

PR #1275, opened 2020-11-03
#1902582: Enhancement:  Add a submenu to the Search for books action …

cbhaley:
…in the Tag browser context menu

I put all the search actions into the new submenu. There are as many as 5.

FWIW: I am indifferent to the presence or absence of the separator between "Manage X" and "Hide category X".

### DIFF
--- a/src/calibre/gui2/tag_browser/view.py
+++ b/src/calibre/gui2/tag_browser/view.py
@@ -634,6 +634,7 @@ class TagsView(QTreeView):  # {{{
                 m.addAction(_('All categories'),
                         partial(self.context_menu_handler, action='defaults'))
 
+        search_submenu = None
         if index.isValid():
             item = index.data(Qt.UserRole)
             tag = None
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I have merged, while testing I noticed a "Search using saved search expression" menu entry. That fails for most things, maybe only add it to the menu if it is not going to fail.

cbhaley:
I added that option in 2/2019 in response to an enhancement request. Looking again at the ER and looking at what that option does, I think it was a mistake to add it. But it is there, so ...

I will do another PR not adding the line if there is no saved search with that item name.

cbhaley:
Actually, that action should show only on the context menu for a saved search. I will fix it.

kovidgoyal:
Sounds good, thanks :)

On Tue, Nov 03, 2020 at 02:26:51AM -0800, Charles Haley wrote:
> Actually, that action should show only on the context menu for a saved search. I will fix it.
> 
> -- 
> https://github.com/kovidgoyal/calibre/pull/1275#issuecomment-721030552

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

PR #1277, opened 2020-11-04
calibre-gui.desktop: rename "calibre" to "Calibre"

MicroJoe:
Rationale: all other .desktop files created by linux.py script are
capitalizing the application names but this one. Also, as Calibre uses
"Calibre" in its titlebar, it would make it more consistent.


kovidgoyal:
The correct capitalization for calibre is lowercase. Lowercase is used
everywhere, including the titlebar, see line 767 or gui2/ui.py

PR #1280, opened 2020-11-07
Bug 1903363: Bulk Metadata Edit: "Set to Undefined" missing on intege…

cbhaley:
…integer columns.

Also added clear on float columns, and added the buttons to single metadata edit.


kovidgoyal:
merged, however the tooltip for the clear buttons just have the column
name, they should probably say "Clear coloumn name"

PR #1283, opened 2020-11-10
Use and prefer XDP interface for checking for network online status

ntninja:
Like #1268, but for the other D-Bus API that can be trivially be converted to a sandbox compatible alternative.

### DIFF
--- a/src/calibre/utils/network.py
+++ b/src/calibre/utils/network.py
@@ -7,30 +7,71 @@ from calibre.constants import iswindows, islinux, isbsd
 from calibre.utils.config_base import tweaks
 
 
-class LinuxNetworkStatus(object):
+class LinuxNetworkStatus:
+
+    # Map of NetworkManager connectivity values to their XDP/GLib equivalents
+    NM_XDP_CONNECTIVITY_MAP = {
+        0: 4,  # NM_CONNECTIVITY_UNKNOWN → Full network
+        1: 1,  # NM_CONNECTIVITY_NONE → Local only
+        2: 3,  # NM_CONNECTIVITY_PORTAL → Captive portal
+        3: 2,  # NM_CONNECTIVITY_LIMITED → Limited connectivity
+        4: 4,  # NM_CONNECTIVITY_FULL → Full network
+    }
 
     def __init__(self):
+        # Prefer desktop portal interface here since it can theoretically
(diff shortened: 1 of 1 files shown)


src/calibre/utils/network.py
```diff
@@ -7,30 +7,71 @@
...
         try:
-            connections = self.manager.Get("org.freedesktop.NetworkManager",
-                        "ActiveConnections")
-            return len(connections) > 0
+            # Meanings of returned XDP/GLib connectivity values:
+            #   * 1: Local only. The host is not configured with a route to the internet.
+            #   * 2: Limited connectivity. The host is connected to a network, but can't reach the full internet.
+            #   * 3: Captive portal. The host is behind a captive portal and cannot reach the full internet.
+            #   * 4: Full network. The host connected to a network, and can reach the full internet.
+            return self.get_connectivity() >= 2
```
ntninja, line 70: @kovidgoyal: Is reporting network connectivity in all but case 1 here the right thing to do, or should this check be different here?

kovidgoyal: This status is used for deciding whether to trigger a scheduled news download for hold off. Therefore it should return true if making connections to http servers over the internet is possible

ntninja: @kovidgoyal: Sorry, I forgot about this for a few days. Could you be more precise on what “making connections […] over *the internet*” means for you?

In particular: If we are reported “limited connectivity”, this may imply anything from »there is only a single other network device and none of the pre-defined news sources reachable« to »every device on the public internet other than the one used to check for “internet” connectivity is reachable«. A well-known case of the latter is the thing were Google is blocked in China so all Google Android device report “no internet” (aka “limited connectivity”) there, but a significant portion of globally available websites is reachable in the chinese Internet nonetheless. Should Calibre consider such a case as “being online”?

Captive portal is easier to decide, since it means that access to most content beyond the local network is intentionally blocked by the ISP to force some kind of authorization/compliance procedure. Calibre probably shouldn't consider this as “being online”, no?

Based on the above, please tell me which of these states you consider to be “online enough” and I'll update the expression appropriately!

kovidgoyal: >= 4 should be correct.

ntninja: Done! (There are no values `> 4`)

PR #1284, opened 2020-11-14
RFC: load plugins from system-wide directory

t-8ch:
As discussed in: https://www.mobileread.com/forums/showthread.php?t=333893
Open points:
- [x] disable updates
- [x] handle load failures
- [x] prevent removal
- [x] prevent replacement
- [x] mention in the UI that this is managed system-wide
- [ ] new translations
- [ ] documentation

### DIFF
--- a/src/calibre/customize/ui.py
+++ b/src/calibre/customize/ui.py
@@ -4,14 +4,14 @@ __copyright__ = '2008, Kovid Goyal <kovid at kovidgoyal.net>'
 
 import os, shutil, traceback, functools, sys
 from collections import defaultdict
-from itertools import chain
+from itertools import chain, repeat
 
 from calibre.customize import (CatalogPlugin, FileTypePlugin, PluginNotFound,
                               MetadataReaderPlugin, MetadataWriterPlugin,
                               InterfaceActionBase as InterfaceAction,
                               PreferencesPlugin, platform, InvalidPlugin,
                               StoreBase as Store, EditBookToolPlugin,
-                              LibraryClosedPlugin)
+                              LibraryClosedPlugin, PluginInstallationType)
 from calibre.customize.conversion import InputFormatPlugin, OutputFormatPlugin
 from calibre.customize.zipplugin import loader
 from calibre.customize.profiles import InputProfile, OutputProfile
@@ -21,11 +21,15 @@ from calibre.ebooks.metadata import MetaInformation
 from calibre.utils.config import (make_config_dir, Config, ConfigProxy,
                                  plugin_dir, OptionParser)
 from calibre.ebooks.metadata.sources.base import Source
-from calibre.constants import DEBUG, numeric_version
+from calibre.constants import DEBUG, numeric_version, system_plugins_loc
 from polyglot.builtins import iteritems, itervalues, unicode_type
 
 builtin_names = frozenset(p.name for p in builtin_plugins)
 BLACKLISTED_PLUGINS = frozenset({'Marvin XD', 'iOS reader applications'})
+
+
+def zip_value(iterable, value):
+    return zip(iterable, repeat(value))
 
 
 class NameConflict(ValueError):
@@ -341,10 +345,19 @@ def reread_metadata_plugins():
             for ft in plugin.file_types:
                 _metadata_writers[ft].append(plugin)
 
-    # Ensure custom metadata plugins are used in preference to builtin
-    # ones for a given filetype
+    # Ensure the following metadata plugin preference is used:
+    # external > system > builtin
     def key(plugin):
-        return (1 if plugin.plugin_path is None else 0), plugin.name
+        installation_type = plugin.installation_type
+        if installation_type is PluginInstallationType.BUILTIN:
+            order = 2
+        elif installation_type is PluginInstallationType.SYSTEM:
+            order = 1
+        elif installation_type is PluginInstallationType.EXTERNAL:
+            order = 0
+        else:
+            raise ValueError(installation_type)
+        return order, plugin.name
 
     for group in (_metadata_readers, _metadata_writers):
         for plugins in itervalues(group):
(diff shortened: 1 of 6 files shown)


kovidgoyal:
This doesn't need integration with any of those things, they have nothing to do with plugins. Plugin import machinery is in zipplugin.py, in particular the load() method.

Regarding the actual concept, getting calibre to load/import the plugins is easy. The hard part is integrating into the rest of the plugin machinery. Disabling updates, dealing withing plugin load failures/incompatibilities, etc.

Freso:
In addition to the plugin that @t-8ch wants to package for the AUR, I would love to make a PKGBUILD for https://gitea.multiname.org/ibu/calibre-tal and was trying to figure out how to do this when I came upon the forum thread and then this PR.

t-8ch:
@kovidgoyal Sorry for the long delay. I'd like to move this forward again.
I think most cases should now be covered. Could you take another look?

src/calibre/customize/ui.py
```diff
@@ -473,7 +477,10 @@ def add_plugin(path_to_zip_file):
     if plugin.name in builtin_names:
         raise NameConflict(
             'A builtin plugin with the name %r already exists' % plugin.name)
-    plugin = initialize_plugin(plugin, path_to_zip_file)
+    if plugin.name in _list_site_plugins:
```
kovidgoyal, line 480: Should be _list_site_plugins()

src/calibre/customize/ui.py
```diff
@@ -676,6 +684,21 @@ def has_external_plugins():
     return bool(config['plugins'])
 
 
+def _list_site_plugins():
+    if not site_plugins_loc:
+        return []
```
kovidgoyal, line 689: Should be a plain return, its a generator

src/calibre/customize/ui.py
```diff
@@ -676,6 +684,21 @@ def has_external_plugins():
...
+
+    try:
+        plugin_file_names = os.listdir(site_plugins_loc)
+    except FileNotFoundError:
```
kovidgoyal, line 693: Should catch OSError not just FileNotFoundError

src/calibre/customize/ui.py
```diff
@@ -676,6 +684,21 @@ def has_external_plugins():
...
+        return []
```
kovidgoyal, line 694: Should be a plain return

src/calibre/customize/ui.py
```diff
@@ -691,21 +714,29 @@ def initialize_plugins(perf=False):
         from collections import defaultdict
         import time
         times = defaultdict(lambda:0)
-    for zfp in list(external_plugins) + builtin_plugins:
+
```
kovidgoyal, line 717: There needs to be support for BLACKLISTED_PLUGINS for system-wide plugins as well. Also, in add_plugin() you have disabled the ability to ad dplugins whose names conflict with system plugins, however, this code prioritizes installed plugins over system plugins.

kovidgoyal (changes requested):

src/calibre/customize/ui.py
t-8ch: As the external plugins do not have their name available before being loaded from disk they can't use the same conflict detection logic as the normal plugins who get their name preloaded from `customize.py`.

Would it be fine to change the logic to:
```py
for p in chain(builtin, system, external):
    plugin = _load_plugin(p)
    if already_initialized(plugin) or blacklisted(plugin):
        continue
```

This would have the following effects:
* Even blacklisted plugins are loaded and would be able to execute code.
* A system-plugin that could not be loaded would allow the same plugin in the users installation to be loaded.

It would also need some official decision if it is fine that a system-wide plugin can completely prevent the user from installing the same plugin on their own.
For me personally this fits my usecase perfectly, but I could imagine that it may not be the most correct solution in general.

kovidgoyal: I would prefer blacklisting to work without loading plugin code. To
achieve that simply require the plugin file name in the system plugins
directory be the plugin name. I believe you require that already for the
name conflict check in add_plugin(). 

Coming to policy. Currently I believe external plugins win over builtin
ones in cases where plugins are used for file types/metadata reading
etc. The natural extension of this is

external plugins > system wide plugins > builtin plugins

To achieve that you have to poulatin _initialized_plugins with external
plugins, then system wide and then finally builtin. In addition there
are some functions in customize/ui.py that will need to be adjusted,
such as reread_metadata_plugins() which performs its own sorting on the
plugin list.

t-8ch: > I would prefer blacklisting to work without loading plugin code.

Ok.

> To achieve that simply require the plugin file name in the system plugins
directory be the plugin name.

Does `require` mean, that it should be validated in code after the plugin has been loaded, or an assumption can be made that it is like that?

> Coming to policy. Currently I believe external plugins win over builtin
ones in cases where plugins are used for file types/metadata reading
etc.

I think builtin plugins win over external ones:
https://github.com/kovidgoyal/calibre/blob/80364863a9d2f8a8c78222d217c7223204147117/src/calibre/customize/ui.py#L682-L685

https://github.com/kovidgoyal/calibre/blob/80364863a9d2f8a8c78222d217c7223204147117/src/calibre/customize/ui.py#L488-L506

kovidgoyal:
require means assume the name is the filename. 

And yes builtin plugins with a given *name* win over custom ones. But if
there are two different plugins that say both provide metadata reading
for the XYZ plugin or conversion to the XYZ format, the custom one wins.

t-8ch:
Thanks for the clarification!

t-8ch:
@kovidgoyal I added an explicit attribute to track the how a given plugin was installed instead of having to infer it from the plugin path. (And used it were appropriate). Is this fine? Should it be split into a dedicated commit?

t-8ch:
@kovidgoyal thanks!
@eli-schwartz Ping for you to set `sys.system_plugins_location` in the ArchLinux package for the next release.

kovidgoyal:
On Fri, Apr 09, 2021 at 06:53:29AM -0700, Thomas Weißschuh wrote:
> @eli-schwartz Ping for you to set `sys.system_plugins_location` in the ArchLinux package for the next release.

I have added an option to ./setup.py install to set it.

eli-schwartz:
Thanks! Would `resources_location + '/plugins'` cause issues due to codepaths using `P()` somehow?

kovidgoyal:
I recommend against that. I doubt it will cause issues, but in the
future I might want to add a plugins directory to resources. If you want
to put it in a similar location I suggest using

<base_directory>/calibre/resources
<base_directory>/calibre/system-plugins

or similar.

eli-schwartz:
Yeah, system-plugins seems like a very elegant way to guarantee no future conflicts will arise. I'll use that.

I just didn't want to end up with /usr/share/calibre and /usr/share/calibre-plugins both taking up space in /usr/share. :)

PR #1285, opened 2020-11-14
Add automerge book command to CLI in order to automate maintenance for multilanguages libraries

juampe:


kovidgoyal:
I'm not sure I understand why this needs to be a separate command. If you want adding + merging, the place to do it is in the existing add command. Add an option to it to merge detected duplicates instead of ignoring them.

juampe:
Thanks Kovid. You are right. The right place is the add command as an option, but first I need to study the add command code to not broken anything. In order to see if this functionality can be a reality (I have tons of books willing to be merged), It was easyer for my programming skills, in this case, to add than to merge. I'll do a try, thanks again.

kovidgoyal:
Feel free to open a new PR if and when you have something for the add command.

PR #1287, opened 2020-11-14
Add an option to remove NCX TOC file when upgrading book internals

thiagoeec:


kovidgoyal:
You cant use an editor setting directly inside polish code. Polish code is called from other places as well, like the polish books tool. Instead, add it to CUSTOMIZATION in polish/main.py and have the editor set that option like it does for the others.

thiagoeec:
Got it! I'll fix it.

PR #1290, opened 2020-11-18
Add automerge book option to add in CLI

juampe:
As commented in https://github.com/kovidgoyal/calibre/pull/1285 


kovidgoyal:
Can you please squash these down to a single commit.

juampe:
Ok, I need additional git superpowers.

PR #1291, opened 2020-11-19
Add automerge option to add command CLI v3

juampe:
As commented in #1285 and #1290


kovidgoyal:
Thanks, I will review it when I have some time.

PR #1292, opened 2020-11-19
Follow link from keyboard

chasecarl:
I implemented functionality that allows to highlight on-screen links and choose one by entering its number (similar to Zathura pdf viewer if you're familiar) that enables fully mouseless viewing experience. I used a relatively simple Alt+f shortcut for this because history-related actions (back, forward) also have Alt shortcuts.
I'm new to RapydScript so the styles used for highlighting are created dynamically for each link - I think there is a better way to do this. And overall highlighting and numbering styles are not final.
Invalid link numbers for now just throw an exception.
FollowLinkOverlay and follow_link.pyj basically used SearchOverlay and search.pyj as a template.
Icon for the follow link action is the Reference mode icon for now.
Besides that, I found 2 issues:
1. If you switch chapters while links are highlighted, the highlighting is gone but you can still jump using numbers that were displayed.
2. If you open a footnote using follow link, then footnote's header will also contain the number assigned to the link by the follow link action.


kovidgoyal:
Thanks, I will review it when I have some time.

kovidgoyal:
Some comments:

1) There is no need for an overlay and a text input at all. Just intercept key events in the iframe and select links according to the key presses

2) The link numbering checkboxes are rather ugly. For better versions, see https://github.com/kovidgoyal/vise  which implements this functionality for web browsing (also in RapydScript)

3) When a link is activated it should be zoomed using a css transform to give visual feedback

kovidgoyal:
And link numbering should happen using the full set of 1-0 and a-z keys, agains see vise linked above for code to implement that.

kovidgoyal:
I decided to implement this myself. It's mostly done just need to add the code to select the hint by keypresses and activate it.

PR #1301, opened 2020-12-11
Fix build with ICU >= 68

heirecka:
ICU stopped defining the macros FALSE and TRUE in its headers with 68:
https://github.com/unicode-org/icu/commit/c3fe7e09d844


heirecka:
Tested locally with icu 67 and 68.

PR #1302, opened 2020-12-12
Bug #1907856: Typing in number fails to replace undefined integers

cbhaley:


cbhaley:
Added an unrelated change: Throw away empty icon specifications in icon rule evaluation.

PR #1307, opened 2020-12-19
Recipe for Substack newsletters

topynate:
This recipe uses the RSS feeds which Substack provides for each newsletter. Unlike the built-in RSS recipe, this one also lets Calibre download all posts from your paid subscriptions.

If you don't add a username and password, it will work exactly like the built-in recipe.

### DIFF
--- a/recipes/substack.recipe
+++ b/recipes/substack.recipe
@@ -23,7 +23,7 @@ from mechanize import Request
 
 class Substack(BasicNewsRecipe):
     title          = 'Substack'
-    __author__     ='topynate'
+    __author__     = 'topynate'
     oldest_article = 7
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Then just replace the built-in recipe, no need for a separate one, since
this is a superset.

topynate:
As far as I know, this is the first recipe for Substack. The "built-in RSS recipe" to which I was referring is simply `BasicNewsRecipe`.

kovidgoyal:
Ah I see, ok. Well add needs_subscription = 'optional' and also __author__ and laguage to the recipe.

topynate:
Yep, added.

PR #1310, opened 2020-12-20
Get a past edition of the Economist (e.g., date='2020-11-28').

shinozukayohei:
A new function edition(date) allows the user to specify a past edition. The default, date='', takes the current edition. The proper cover URL appears in the Calibre desktop app, though not on Kindle Oasis -  a minor remaining issue. Much respect and admiration for Kovid Goyal's work.


kovidgoyal:
Some comments:

1) There are two identical economist recipes (for historical reasons) you should make changes to both economist.recipe and economist_free.recipe

2) dont use urllib to make requests, instead do self.index_to_soup() if you need unparsed html pass raw=True to it.

PR #1315, opened 2020-12-25
Bug #1909293:Composite tag-like column: problem with case

cbhaley:


cbhaley:
Also includes fix for Bug #1909258: Fix for tags-like composite columns "shown in the tag browser" not considered categories in quickview.

PR #1316, opened 2020-12-26
Bug #1909332: Duplicated rules appear to overwrite next item in list …

cbhaley:
…Edit

Actually the duplicated rule was exchanged with the last rule.


cbhaley:
Added Bug 1909339: Enhancement Request: Rules editors: Button to open icons folder

PR #1322, opened 2020-12-29
replace small image src with srcset in guardian.recipe

hehonghui:
This change is [Reviewable](https://reviewable.io/reviews/kovidgoyal/calibre/1322)


kovidgoyal:
This is already done in preprocess_html() if you want to pick some URL other than the first change the logic there.

PR #1333, opened 2021-01-11
Update the mediapart newsfeed

lhoupert:
Hi!
I am a reader of Mediapart and I was frustrated that the current version of the recipe only provided the categories "La Une" and "Brèves".  I updated the recipe to add new categories from the website [Mediapart](https://www.mediapart.fr/)


lhoupert:
@kovidgoyal, Do you know how I can modify this recipe code so the .pobi displayed in the kindle menu has a simple cover image?  For example this one:
https://upload.wikimedia.org/wikipedia/fr/thumb/8/8a/Logo-crieur.svg/800px-Logo-crieur.svg.png

Thank you!

kovidgoyal:
If you want to change the cover image, set cover_url in the recipe

lhoupert:
> If you want to change the cover image, set cover_url in the recipe

Thank you @kovidgoyal I have two question about the mediapart cover:
1. Is there an easy way to extend the "frame of the cover" so it has the size of a book cover?   The current dimension of the logo used for the cover is 800×600. It looks a little bit weird when displayed on the kindle menu. Is there a way to center this logo on a 800x1040 frame so it looks like a proper newspaper cover? Or the easiest would be to create a .png file at the proper dimension, host it somewhere and change the cover_url?
2. The cover image is not displayed on the kindle when the news is sent to the kindle email address. It appears briefly after plugging the kindle to the computer by usb (with calibre on) but disappears after a few seconds. Is it related to the size and format of the cover image ? Is there a way to fix that? I don't have this problem when I fetch the "New Yorker", the cover image is directly displayed on my kindle after sending the email (I don't need to connect it to Calibre)

lhoupert:
> > If you want to change the cover image, set cover_url in the recipe
> 
> Thank you @kovidgoyal I have two question about the mediapart cover:
> 
>     1. Is there an easy way to extend the "frame of the cover" so it has the size of a book cover?   The current dimension of the logo used for the cover is 800×600. It looks a little bit weird when displayed on the kindle menu. Is there a way to center this logo on a 800x1040 frame so it looks like a proper newspaper cover? Or the easiest would be to create a .png file at the proper dimension, host it somewhere and change the cover_url?
> 
>     2. The cover image is not displayed on the kindle when the news is sent to the kindle email address. It appears briefly after plugging the kindle to the computer by usb (with calibre on) but disappears after a few seconds. Is it related to the size and format of the cover image ? Is there a way to fix that? I don't have this problem when I fetch the "New Yorker", the cover image is directly displayed on my kindle after sending the email (I don't need to connect it to Calibre)

So after doing a quick check, it seems that setting up a cover at the .jpeg format fix the problem. The cover does't disappear from the kindle menu. I will update the recipe with a new PR

PR #1335, opened 2021-01-14
cover update2

lhoupert:
Hi @kovidgoyal ,

Another PR on my side. This one is not finished yet. I wanted to ask you if there is an easy way to add the date on the Cover image. 
I saw that something like that exist in the `default_cover` function and I was wondering if you could help me implement it. 

Right now the cover_url is https://raw.githubusercontent.com/lhoupert/calibre_contrib/main/mediapart.jpeg

I would like to display the date in bottom part of the image (above the pixel 250 of the bottom).

### DIFF
--- a/recipes/mediapart.recipe
+++ b/recipes/mediapart.recipe
@@ -299,7 +299,8 @@ class Mediapart(BasicNewsRecipe):
             p.setFont(font)
 
             # Add date
-            p.drawText(125, 750, date)
+            r = QRect(0, 600, 744,200)
+            p.drawText(r, Qt.AlignmentFlag.AlignJustify | Qt.AlignmentFlag.AlignVCenter | Qt.AlignmentFlag.AlignCenter, date)
             p.end()
             print(f'create_cover_mediapart\nDate: {date}\n')
             return pixmap_to_data(img)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
there are many apis you can use for this. calibre comes with PyQt which
has a painter api you can use to draw on images. If you want something
more out of the box, see the calibre_cover2() function

recipes/mediapart.recipe
```diff
@@ -243,21 +248,71 @@ class Mediapart(BasicNewsRecipe):
...
+            p.setPen(pen)
+
+            font = QFont()
+            font.setFamily('Times')
+            font.setBold(True)
+            font.setPointSize(78)
+            p.setFont(font)
+
+            # Add date
+            p.drawText(125, 750, date)
```
lhoupert, line 302: Is there a way to center horizontally the text  ?

lhoupert:
Thanks for your quick reply. I followed your advice and managed to find a way to add the date on top of the cover

It is quite basic but it works :-) See example below. I have to manually define the position of the text with 
```python3
 p.drawText(125, 750, date)
```
I was wondering if there was a better way to center the text horizontally?

Thank you! 
Screenshot 2021-01-14 at 22 48 16

kovidgoyal:
Use the drawText overload that takes a QRect and set the alignment flags to center.

lhoupert:
> Use the drawText overload that takes a QRect and set the alignment flags to center.

Thank you very much for the advice! I updated it in the last It in https://github.com/kovidgoyal/calibre/pull/1335/commits/d0ced3eb48ef4dbdf287af29929676d5fe462fd9 but it seems that one of the test failed. Any idea why?

PR #1336, opened 2021-01-15
add edition on cover

lhoupert:
Hi @kovidgoyal  I didn't finish the previous PR. I wanted to add information about the weekday and the time on the cover. This new PR should fix that see the results below:

Screenshot 2021-01-15 at 11 24 59

The test may be failing like previously. It may be due to Qt Alignment? I got a depreciation warning on my computer:
```
 qp.drawText(r, Qt.AlignmentFlag.AlignJustify | Qt.AlignmentFlag.AlignVCenter | Qt.AlignmentFlag.AlignCenter, date)
test_print_cover_img.py:62: DeprecationWarning: an integer is required (got type Alignment).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
```

### DIFF
--- a/recipes/mediapart.recipe
+++ b/recipes/mediapart.recipe
@@ -8,7 +8,7 @@
 #   3) Fix the cover image so it doesnt disappear from the Kindle menu
 # ( cover image format is changed to .jpeg)
 # 14 Jan 2021 - Add Mediapart Logo url as masthead_url and change cover
-#   by adding overlaying the date on top of the Mediapart cover
+#   by overlaying the date on top of the Mediapart cover
 from __future__ import unicode_literals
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
The failing test had nothing to do with you, dont worry about it.

lhoupert:
Hi @kovidgoyal ,

I just realised that some of the articles I am fetching with this recipe extend on several pages. Therefore I will have to adapt the recipe to 1) detect if the the article spread on several pages, 2) loop through the pages and merged the content. 

It should not be too complicated to do, but I was wondering if you know some recipes which are doing that? If yes, could you forward me some example  so I don't '' reinvent the wheel ''?

PR #1338, opened 2021-01-15
ENH: Added surname prefixes to author sort

madphysicist:
If an author last name is preceded by von, van, di, la, le, da, de, etc. include that in the last name.

So `Leonardo da Vinci` -> `da Vinci, Leonardo` instead of `Vinci, Leonardo da`. This seems to generally be the more acceptable approach. A notable exception is Ludwid van Beethoven. However, he was not a popular author of literature, and most of the examples I've come across sorting my library do follow the rule.

I've also simplified / sped up the tokenizing code. There is no need to allocate a new list for every prefix and suffix that is removed. Simply looping over the allowed range and using the `else` clause of the `for` loop does the same thing, but simpler. In the same spirit, I removed the need to recount all the Counter values when stripping parentheses.

My one concern is that I'm not sure where to put unit tests for the changes.


kovidgoyal:
These kinds of things are *highly* culture dependant. I'm pretty sure you will find cultures on earth where de van etc are treated as normal names. As such if you want to do this, it needs to be configurable. Add it to the tweaks after author_name_copywords

kovidgoyal:
As for unit tests, just create a file named metadata/author_name_tests.py add the tests there and add it to the list of tests loaded in under misc. Look at any of the other tests there for examples.

madphysicist:
I totally agree that the selection here is very Europe-centric, and even more so just a subset of what I found while skimming the database of my library. In other cultures it may be suffixes instead of prefixes, and I'm not sure if `author_surname_prefixes` an entirely appropriate name for the key. Either way, I moved it after `author_name_copywords`.

Will add unit tests shortly. Thanks for the pointer.

madphysicist:
It looks like I got the hang of the unit tests finally. One new question arose: We remove brackets in the variable `sauthor`. Right now, the bracketed portions are effectively replaced by the empty string. However, it may be worth replacing them with a space instead. What was the original use-case for removing brackets? Is it worth considering putting a space in their place?

kovidgoyal:
Why a space? bracketed bits are typically role information like (ed.)
for editor or (illus) for illustrator. No culture I know of in the world
uses brackets in people's names.

madphysicist:
That makes sense. I don't know of any such culture either (although it wouldn't surprise me if one turned up eventually). In that case, I think the PR is complete unless you have any further suggestions.

kovidgoyal:
The code looks mostly ok, however, thinking about it, I am not at all convinced that sorting Leonardo da Vinci as da Vinci, Leonardo is the smart thing to do. It just means Lenoardo da Vinci will sort the same as Benny da Gangster. I would prefer having them sort under V and G respectively. Seems much more useful than shoving all such names under one bucket. 

Why do you feel this is a desirable thing to do? And if you really want to do this, then it needs to be an option, defaulting to off (you can just add another tweak, a boolean controlling wheter to use author_surname_prefixes.

madphysicist:
In almost all the cases that I've seen, names with a prefix like that include the prefix as part of the official surname. I'll add the tweak to turn it on, make it off by default, and document it.

kovidgoyal:
I dont doubt you, but the purpose of the sort name isn't to identify the
surname but to make it easier to sort meaningfully. In any case let me
know when you add the boolean and I will merge.

madphysicist:
I've added a boolean `author_use_surname_prefixes` as you requested. On second thought, I'm not sure why that's preferable to just leaving the tuple empty. I also took the liberty of adding `'National'` to `author_name_copywords`. I don't think anyone is likely to have that in their name, and it takes care of `"National Lampoon"` and similar author groups.

kovidgoyal:
I want a separate boolean so that there can be a default, prepoulated
list.

PR #1362, opened 2021-02-23
Fix f-string syntax error: remove equal symbol

fredgan:


kovidgoyal:
That's not a syntax error. Your version of python is too old to support the {name=} syntax in f-strings.

kovidgoyal:
If for ome reason you are stuck on python 3.7 the correct fix is to replace {slug=} with slug={slug}

fredgan:
Well，my python is 3.7.6 indeed.  I didn't see the `python 3.8 at least` in the homepage readme.

kovidgoyal:
The versions of all software used by calibre is specified in
sources.json. However, you are welcome to send another PR properly
fixing this issue as I described above.

fredgan:
By the way, I want to submit an issue or feature request, then I go the the `launchPad` to register an account. Then I received a validation email. However, the link in the validation is unable to validate my email address. It prompts "Invalid address validation link". In this repo, the issue tracker is not open either. So I am not able to submit issues.

kovidgoyal:
I'm afraid I dont have control over launchpad, so cant help you with
that. If you need to submit issues, there are many ways you can do it,
not just launchpad, use the calibre forums for example at mobileread.com

PR #1375, opened 2021-03-11
Bug #1918428: virtual_libraries() performance improvement

cbhaley:


kovidgoyal:
Merged, with a few more speed improvements.

kovidgoyal:
Although, maybe the cache should also be invalidated when the viral
libraries pref is set.

PR #1378, opened 2021-03-13
feat: add primary-writing-mode option ePub/azw3 output

digglife:
Add `primary-writing-mode` option for supporting right-to-left language ebooks such as Japanese Manga.

Feature request here:

https://bugs.launchpad.net/calibre/+bug/1919006


kovidgoyal:
Some comments:
1) In the docs for the option mention that the default is to use the value from the input document, if present.
2) You need to add the option to a couple more places. The calibre GUI and the content server look in the gui2/convert folder and conversion_wdgets.pyj

luzpaz:
Is there an ETA on this ?

PR #1382, opened 2021-03-18
Add API to show/hide categories in the tag browser.

cbhaley:


kovidgoyal:
Is there a reason operation is a string and not a bool?

cbhaley:
I thought a string was more readable if not using named parameters. It also allows adding other operations if desired, such as 'toggle' or visibility queries.

kovidgoyal:
I'm ok with using a string, though then the function should be renamed to not have visibility in the name as otherwise it doesnt really make sense to have operations other than show/hide.

cbhaley:
I changed the name to tb_category_visibility. The operations supported are now:
        - 'show' to show the category in the tag browser
        - 'hide' to hide the category
        - 'toggle' to invert its visibility
        - 'is_visible' returns True if the category is currently visible, False otherwise

PR #1383, opened 2021-03-20
Bug #1920592 and Enhancement #1920576

cbhaley:


cbhaley:
Kovid, don't commit these for the moment. I have having some problem with elision that I want to fix, assuming I figure out how.

cbhaley:
OK, ready to commit.

cbhaley:
Another commit to fix an exception when launching the template editor from the custom column definition dialog.

PR #1385, opened 2021-03-21
Enhancement #1920574: allow changing fonts in the template tester

cbhaley:


kovidgoyal:
Why do you need to set the writing system? As far as I knwo Qt picks
that up automatically from the text.

cbhaley:
Without it I was getting Japanese and other non-latin fonts, which don't work for templates.

kovidgoyal:
I think the correct way to do it is to not use the font family nae, but
rather f = QFont(self) f.setFixedPitch(True) self.setFont(f)

cbhaley:
But some people want non-fixed pitch fonts. I do, for example.  Or are you saying do that instead of "monospace"? I did that because that is what the template editor did before .

kovidgoyal:
Yup instead of monospace. Generaly specify the characteristic you want
rather than a font family name.

On Sun, Mar 21, 2021 at 08:47:19AM -0700, Charles Haley wrote:
> But some people want non-fixed pitch fonts. I do, for example.  Or are you saying do that instead of "monospace"? I did that because that is what the template editor did before .
> 
> -- 
> You are receiving this because you commented.
> https://github.com/kovidgoyal/calibre/pull/1385#issuecomment-803607596

-- 

Dr. Kovid Goyal 
https://www.kovidgoyal.net
https://calibre-ebook.com

cbhaley:
Changes made.

PR #1386, opened 2021-03-21
I removed the duplicate `import` of `collections`.

0verk1ll:
I made this change to address a suggestion from LGTM (https://lgtm.com/projects/g/kovidgoyal/calibre?mode=tree&severity=recommendation&ruleFocus=1818040193).

### DIFF
--- a/src/calibre/gui2/ui.py
+++ b/src/calibre/gui2/ui.py
@@ -17,9 +17,7 @@ import re
 import sys
 import textwrap
 import time
-from collections import (
-    OrderedDict, deque
-)
+from collections import OrderedDict, deque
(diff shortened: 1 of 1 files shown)


src/calibre/gui2/ui.py
```diff
@@ -10,15 +10,16 @@
...
 import gc
 import os
 import re
 import sys
 import textwrap
 import time
-from collections import OrderedDict
+from collections import (
+    OrderedDict, deque
+)
```
eli-schwartz, line 22: This import is short enough it should be online line rather than a multiline parenthesized import. See the style in the rest of this file (and elsewhere).

PR #1393, opened 2021-03-27
get the economist cover url from json

xxyzz:
The `economist_free.recipe` file is for backward compatibility and it can't be a symbolic(or hard) link?


kovidgoyal:
Umm why? The current recipe fetches the cover fine. And you cant use python 3.8 idioms in recipes. They run on python versions all the way back to python 2.7. And no I dont trust git to get symbolic links right on all operating systems.

xxyzz:
The cover URL is already in the `script` tag so we don't need to parse another web page to get it. Thus save some time.

PR #1396, opened 2021-03-30
Template formatter and tester changes:

cbhaley:
Formatter:
- Clean up line number processing to make them more accurate
- Provide more detail in breakpoint labels
- Ensure all error strings are translatable
- Clean up error strings to distinguish between translated and untranslated parts

Tester:
- Remove context menu, replacing it with 2 small buttons for load and store


cbhaley:
Kovid, in case it matters, I am finished doing pushes

PR #1397, opened 2021-03-31
Bug #1922131: Advanced rules editor fails after new library with …

cbhaley:
…structure


cbhaley:
This bug is in fact a regression.

PR #1403, opened 2021-04-04
restore cover iff the thumbnail is the non-available image

xxyzz:
Sometimes Kindle downloads newer cover but calibre always restore the older cover. This pull request will restore the cover if and only if Kindle downloads the "no image available" image.


kovidgoyal:
I dont think this is a good idea. The no cover available graphics could
change across firmware revisions and be of a different size and
therefore different hash on different models.

PR #1422, opened 2021-04-18
Two commits: 

cbhaley:
- Fix a serious regression breaking calling stored templates introduced on 12/4. This easily could be serious enough to provoke an interim release. Apologies.

- Add a tweak to set the tab stop width in the template editor.

### DIFF
--- a/resources/default_tweaks.py
+++ b/resources/default_tweaks.py
@@ -570,3 +570,8 @@ exclude_fields_on_paste = []
 # Useful if for some reason your operating systems network checking
 # facilities are not reliable (for example NetworkManager on Linux).
 skip_network_check = False
+
+#: Tab stop width in the template editor
+# Sets the width of the tab stop in the template editor in "average characters".
+# For example, a value of 1 results in a space the width of one average character.
+template_editor_tab_stop_width = 4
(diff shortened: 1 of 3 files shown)


kovidgoyal:
what're the implications of the breakage?

cbhaley:
Calling a stored template always raises an exception unless you are using the new debugger.

The fact that we haven't already heard about it seems to say that not many people use stored templates.

kovidgoyal:
On Sun, Apr 18, 2021 at 09:15:28AM -0700, Charles Haley wrote:
> Calling a stored template always raises an exception unless you are using the new debugger.

So this would mean calibre would not start if a stored template is used
for a custom column for instance?

> 
> The fact that we haven't already heard about it seems to say that not many people use stored templates.

Yeah, lets get at least one bug report, since this seems to be more of a
denial-of-service than data corruption bug as far as I can see.

cbhaley:
No, calibre starts. If the template is used in a rule then the rule silently doesn't work. If it is used in a composite column you see an exception message as the template value. I don't know about other contexts.

And yes, it probably isn't a corruption problem, but it might be in a template based global search & replace or in an Action Chain where the template returns the wrong value.

kovidgoyal:
OK I will take a decision on whether to make a new release tomorrow
morning, as its getting late here.

cbhaley:
I don't think you need to make a release just for this problem. It seems that people (if any) who are using stored templates haven't upgraded. I announced the problem and its fix in the template thread so people are warned. If someone does upgrade and has problems then they can downgrade to 5.14.

Of course, if you are making a release anyway ... :)

PR #1430, opened 2021-04-27
Revert "Fix building on Arch"

eli-schwartz:
This reverts commit b8d8cf27e290f75791679cb3bce3f250904f2367.

libraqm (an obscure PIL functionality) was recently switched to be directly linked by Pillow, rather than dlopened, which meant it was a bit of a hidden dependency since calibre uses the imagingft module. But it is now directly depended on by the Arch packaging, so it no longer needs to be installed to get a functional Pillow.

### DIFF
--- a/setup/arch-ci.sh
+++ b/setup/arch-ci.sh
@@ -5,7 +5,7 @@
 
 set -xe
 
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Repos are synced, test_imaging still passes ok :)

PR #1432, opened 2021-04-30
Add ASIN to MOBI EXTH header and metadata

xxyzz:
This pull request adds two features:

1. Save ASIN from EXTH header to the 'mobi-asin' metadata.
2. Save ASIN to EXTH headers if the book's metadata has 'mobi-asin' record, this can be used by plugins to update a book's ASIN for Word Wise and X-Ray.

### DIFF
--- a/src/calibre/ebooks/metadata/mobi.py
+++ b/src/calibre/ebooks/metadata/mobi.py
@@ -325,7 +325,7 @@ class MetadataUpdater(object):
             stop, = unpack('>I', self.data[offoff + 8:offoff + 12])
         return StreamSlicer(self.stream, start, stop)
 
-    def update(self, mi):
+    def update(self, mi, asin=None):
         mi.title = normalize(mi.title)
 
         def update_exth_record(rec):
@@ -399,6 +399,11 @@ class MetadataUpdater(object):
                 not added_501 and not share_not_sync):
             from uuid import uuid4
             update_exth_record((113, unicode_type(uuid4()).encode(self.codec)))
+
+        if asin is not None:
+            update_exth_record((113, asin.encode(self.codec)))
+            update_exth_record((504, asin.encode(self.codec)))
+
         # Add a 112 record with actual UUID
         if getattr(mi, 'uuid', None):
             update_exth_record((112,


kovidgoyal:
DO you have any examples of books in the wild which use 504 and not 113?
I am ok with reading mobi-asin from it, but then you have to decide
which takes priority in case they conflict.

As for writing mobi-asin, I'm afraid thats not going to happen. calibre
very deliberately chooses not to do that, because amazon uses that id
for various internal services, therefore re-using the id from an
existing book after that book has been converted/edited in calibre is
not a goo didea.

xxyzz:
`headers.py` seems not used when a book is added to the library, `mobi-asin` is valid ASIN for some books. I can remove the first commit so the ASIN record will only be updated if 'mobi-asin' is changed.

kovidgoyal:
I am fine with adding an optional parameter defaulting to off to gate these changes. headers.py is definitely used when reading metadata see metadata/mobi.py

xxyzz:
>headers.py is definitely used when reading metadata see metadata/mobi.py

If that's the only place to set the mobi-asin metadata, all book's mobi-asin should be uuid but that's not the case. I guess there must exist some code to set the value from the header but couldn't find it.

Another unrelated  question: I see lots of `lopen()` in the calibre code but couldn't find where it defined. Why use `lopen()` instead of `open()`?

kovidgoyal:
On Fri, Apr 30, 2021 at 09:42:36PM -0700, xxyzz wrote:
> >headers.py is definitely used when reading metadata see metadata/mobi.py
> 
> If that's the only place to set the mobi-asin metadata, all book's mobi-asin should be uuid but that's not the case. I guess there must exist some code to set the value from the header but couldn't find it.

No 113 is set when the book is changed, i.e. edited or converted. Not
when updating metadata alone.

> 
> Another unrelated  question: I see lots of `lopen()` in the calibre code but couldn't find where it defined. Why use `lopen()` instead of `open()`?

lopen is a wrapper for open that was needed for python 2.x because
python 2.x did not create non-inheritable file descriptors.

PR #1437, opened 2021-05-05
Bug #1927179: VL set to wildcard crashes Calibre. …

cbhaley:
…Note that the problem had nothing to do with wildcards. It was triggered by an unanchored search specification (one with no column), which caused all the composite columns to be evaluated.

### DIFF
--- a/src/calibre/db/cache.py
+++ b/src/calibre/db/cache.py
@@ -674,8 +674,23 @@ class Cache(object):
         cover_as_data is True then as mi.cover_data.
         '''
 
+        # Check if virtual_libraries_for_books rebuilt its cache. If it did then
+        # we must clear the composite caches so the new data can be taken into
+        # account. Clearing the caches requires getting a write lock, so it must
+        # be done outside of the closure of _get_metadata().
+        vl_cache_was_none = self.vls_for_books_cache is None
         with self.safe_read_lock:
             mi = self._get_metadata(book_id, get_user_categories=get_user_categories)
+        if vl_cache_was_none and self.vls_for_books_cache is not None:
+            try:
+                with self.write_lock:
+                    self.clear_composite_caches()
+            except DowngradeLockError:
+                # We can't clear the composite caches because a read lock is set.
+                # As a consequence the value of a composite column that calls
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I'm a little concerned about clearing the cache without a write lock. Is it really always safe for multiple threads to concurrently access the cache while it can be modified? Is there some reason we cant modify virtual_libraries_for_books to use more granular locking. i.e. use `@api` and then acquire the read and write locks as needed inside the function body

cbhaley:
I would be happier doing it right. The problem is that I don't know how. I will try to figure out granular locking.

kovidgoyal:
Just put the sections that requir read locking in 

with self.read_ock:
  ...

and then clear the cache in a wite self.write_lock block

the only subtety would be to make sure that releasing the locks outside
the blocks is not an issue.

cbhaley:
There is no way to acquire a write lock in virtual_libraries_for_books() because it is called by get_metadata() that has already acquired a read lock.

I have another idea for how to do it -- clear the cache in view.virtual_libraries_for_books() instead of in cache.virtual_libraries_for_books(). More later ...

cbhaley:
I ended up clearing the composite cache in cache.get_metadata().

kovidgoyal:
On Wed, May 05, 2021 at 08:48:40AM -0700, Charles Haley wrote:
> I ended up clearing the composite cache in cache.get_metadata().

I have merged, changing the code slightly to hold a read lock when
checking if the cache needs to be rebuilt. Also I dont quite follow why
we need to check for a DowngradeLockError. A DowngradeLockError happens
when you try to acquire a read lock while holding a write lock. This is
what safe_read_lock guards against.

cbhaley:
The DowngradeLockError was a mistake. I misremembered the exception I was seeing when I tried to use write locks in virtual_libraries_for_books().

However, the change you made points to another problem. The method virtual_libraries_for_books() is protected by a shared read lock. This means that in theory multiple threads could be refilling the cache because the test for recursion is not protected. What a newbie mistake. I will fix it to use a gateway lock then make another PR.

PR #1438, opened 2021-05-05
Ensure that updating the vl_book cache is both thread safe and …

cbhaley:
…recursion safe.

### DIFF
--- a/src/calibre/db/cache.py
+++ b/src/calibre/db/cache.py
@@ -15,6 +15,7 @@ import traceback
 from collections import MutableSet, Set, defaultdict
 from functools import partial, wraps
 from io import BytesIO
+from threading import Lock
 from time import time
 
 from calibre import as_unicode, isbytestring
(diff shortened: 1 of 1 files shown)


cbhaley:
Although I think that making the vls for books cache thread safe, I realized that at the moment non-GUI threads don't use it. The virtual_libraries() template function only works on the GUI thread. Of course, that doesn't prevent "other thread" jobs that involve plugins from directly using the method.

I am not suggesting that my changes be ignored. Using a thread-safe lock is a better solution than system that only detects recursion on the same thread. I suppose I am musing about the complexity of threading questions.

BTW: I looked at using "with lock()" but setting up a context manager that properly dealt with being unable to obtain the lock seemed more complicated than doing in the "old" way.

kovidgoyal:
I have merged, changing to use a context manager and to hold the lock
while checking if the cache exists. There is probably still some
raciness in get_metadata() where the cache is checked and possibly
cleared holding different sets of locks. But I dont see a good way of
fixing that without risking deadlocks.

cbhaley:
Your change to hold the lock before checking is equivalent to putting the check outside the try_lock() because the lock may or may not have been obtained at the point of check. The attached patch fixes that.
[patch.txt](https://github.com/kovidgoyal/calibre/files/6433477/patch.txt)

kovidgoyal:
right, thanks.

PR #1439, opened 2021-05-08
Filter read only device on macOS and add test of os.makedirs()

xxyzz:
macOS seems to ignore the dirty bit in FAT and most of the time it works fine but last night I encountered an `OSError`(I/O error) at https://github.com/kovidgoyal/calibre/blob/43c70c10d07ec8a9fd8ecf1624c44685b8e41aaf/src/calibre/devices/usbms/device.py#L995

when I sent a book, so I think it would be nice to add the tests.

### DIFF
--- a/src/calibre/devices/usbms/device.py
+++ b/src/calibre/devices/usbms/device.py
@@ -424,6 +424,7 @@ class Device(DeviceConfig, DevicePlugin):
             return ans
         self._card_a_prefix = get_card_prefix('carda')
         self._card_b_prefix = get_card_prefix('cardb')
+        self.filter_read_only_mount_points()
 
     def find_device_nodes(self, detected_device=None):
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Why test makedirs? Is it possible for makedirs to fail while creating a file doesnt? Also unlike open() makedirs() will fail if the directory exists.

xxyzz:
It's a really wired error... I sent a book, no error occurred. Then I ejected my Kindle but the book wasn't there. I reconnected the Kindle and tried again then I got the IO error but I when I sent another book it works as normal.

I rebooted both my Kindle and mac but the error still exists until I cleared the dirty bit via `fsck`. Maybe something wrong happened when `os.makedirs` was called and it would fail again with the same path.

I think the right fix should be to detect the dirty bit. I will see if that's possible, sorry to bother.

PR #1449, opened 2021-06-10
Fix & improve The Hindu recipe

shivaprsd:
Summary:
* Fix: loading lead image of articles
* Fix: avoid duplicating of subheading
* Add: article summary to show up in TOC

Explanation:

The lead image markup seems to have changed since the last edit to this
recipe. Hindu now uses a `<picture>` element with a placeholder  and
the actual URL in many `<source>` tags. All img's with `data-src-template`
now lie outside the included tags of the soup (the removed for loop was
having no effect).

Articles contain identical `h2.intro` tags at two different places. Only
one is displayed on the website based on screen size, but the recipe is
including both, making subheadings appear twice.

Even for articles without an `h2.intro` tag, a summary is present in the
standard meta tag; enough for a more informative table of contents (e.g.
Highlights view of Kindle is showing only a "No details available" below
each title now).

### DIFF
--- a/recipes/hindu.recipe
+++ b/recipes/hindu.recipe
@@ -3,7 +3,7 @@ __license__ = 'GPL 3'
 __copyright__ = '2009, Kovid Goyal <kovid@kovidgoyal.net>'
 
 from calibre.web.feeds.news import BasicNewsRecipe
-import string
+import string, re
 
 
 def classes(classes):
@@ -30,9 +30,25 @@ class TheHindu(BasicNewsRecipe):
     ]
 
     def preprocess_html(self, soup):
-        for img in soup.findAll('img', attrs={'data-src-template': True}):
-            img['src'] = img['data-src-template'].replace('BINARY/thumbnail', 'alternates/FREE_660')
+        img = soup.find('img', attrs={'class': 'lead-img'})
+        try:
+            src = img.parent.find('source').get('srcset')
+            img['src'] = re.sub(r'(ALTERNATES)/.+?/', r'\1/FREE_660/', src)
+        except (TypeError, AttributeError):
+            pass
+        # Remove duplicate intro
+        for h in soup.findAll('h2', attrs={'class': 'intro'})[1:]:
+            h.extract()
         return soup
+
+    def populate_article_metadata(self, article, soup, first):
+        try:
+            desc = soup.find('meta', attrs={'name': 'description'}).get('content')
+            if not desc.startswith('Todays paper'):
+                desc += '...' if len(desc) >= 199 else ''   # indicate truncation
+                article.text_summary = article.summary = desc
+        except AttributeError:
+            return
 
     def articles_from_soup(self, soup):
         ans = []


shivaprsd:
Hi @kovidgoyal, any idea how to show the author's name in the TOC?
I tried setting `article.author` but it does not show up on my Kindle. So
had to use the workaround of adding it to the summary as well.

Curiously, when I add an img with `add_toc_thumbnail` it shows up below
the thumbnail. So, that property is definitely registered.

Is this the way Kindle does things? Or has it to do with how Calibre
generates the TOC?

kovidgoyal:
As far as I remember, kindles dont support showing author metadata, so you need to add it in description. However it has been a very long time since I looked at that, so I could be misremembering.

shivaprsd:
Okay, I debugged it with `ebook-convert --debug-pipeline` and checked
`input/index.ncx`. When an img is added with `add_toc_thumbnail`, tags:
`<calibre:meta name="toc_thumbnail">` & `<calibre:meta name="author">`
are present in the ncx, and the TOC shows author's name in Kindle. But
when img is not added both the meta tags disappears.

So it looks like a bug in the way Calibre generates the ncx, seeing how
Kindle shows the author's name in the case when the meta tag is present.
Can you look into this? I can't figure out where & how TOC is generated.

recipes/hindu.recipe
```diff
@@ -30,10 +30,26 @@ class TheHindu(BasicNewsRecipe):
...
+        for h in soup.findAll('h2', attrs={'class': 'intro'})[1:]:
+            h.extract()
         return soup
 
+    def populate_article_metadata(self, article, soup, first):
+        try:
+            desc = soup.find('meta', attrs={'name': 'description'}).get('content')
+            if not desc.startswith('Todays paper'):
+                desc += '...' if len(desc) >= 199 else ''   # indicate truncation
+                article.text_summary = article.summary = desc
```
shivaprsd, line 49: BTW, `article.text_summary = clean_ascii_chars(desc)` needed?

kovidgoyal:
The ToC is generated in feeds/news.py just search for toc_thumbnail an dis rendered in metadata/toc.py in the render() method. From a quick read, I dont see how author would be present only when thumbnail is present in either of those.

shivaprsd:
Sorry, my bad. I was running the wrong recipe the second time.

I ran it properly and `index.ncx` is generated correctly. But I do see
that the final `toc.ncx` is identical in both cases. Neither has any
info about the author or thumbnail. Then how does it know to display the
thumbnail in the first case? This additional info is stored somewhere
else?

kovidgoyal:
kindle formats dont actually use ncx files. Look at
create_periodical_index() in writer2/indexer.py to see where thumbnail
information is added to the index in a MOBI file.

shivaprsd:
Right. I can't see what could be done there.

I guess the TOC layout is also baked into the .mobi? Or is there any css
that controls it?

kovidgoyal:
the toc is rendered by the firmware, from index entries, there is no
markup for it at all.

recipes/hindu.recipe
shivaprsd, line 49: One last thing, `article.text_summary = clean_ascii_chars(desc)`
needed?

kovidgoyal: It's safer, but not strictly needed.

shivaprsd:
So `extra_css` won’t have any effect
on articles if the output is .mobi?

kovidgoyal:
It depends on what the CSS is, calibre conversions translate CSS into html 3.2 wherever
possible when converting to MOBI.

PR #1450, opened 2021-06-11
Use Pixel-Based Scrolling in the Main Library View

elsiehupp:
The fact that Qt defaults to `ScrollPerItem` is a perennial annoyance for me. The changes here allow the main library view to scroll smoothly, which is the norm on macOS.

(If you prefer `ScrollPerItem` on Windows and/or Linux, this change could probably be made Mac-only. I’m not sure what the norms are on Windows and on Linux.)

The scroll increment could probably stand to be finessed, and there are probably other places where smooth scrolling could be turned on.

(The righthand sidebar has smooth scrolling, but the lefthand sidebar does not. The main Preferences window has smooth scrolling, too. I’m not sure where else to check.)

I’m not sure how to rebase my merge request. It would be easiest if you have a bot that lets me just type `/rebase`.

Anyway, thoughts?

### DIFF
--- a/src/calibre/gui2/library/views.py
+++ b/src/calibre/gui2/library/views.py
@@ -227,6 +227,12 @@ class BooksView(QTableView):  # {{{
         QTableView.__init__(self, parent)
         self.pin_view = PinTableView(self, parent)
         self.gesture_manager = GestureManager(self)
+
+        self.setHorizontalScrollMode(QAbstractItemView.ScrollMode.ScrollPerPixel)
+        self.setVerticalScrollMode(QAbstractItemView.ScrollMode.ScrollPerPixel)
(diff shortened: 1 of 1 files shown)


eli-schwartz:
> I’m not sure how to rebase my merge request. It would be easiest if you have a bot that lets me just type `/rebase`.

That's not how Github works, sadly. Gitlab does. But in this case the problem is you've got a strange commit 5616310eba099a719a4b20d7e647c7ddc13ec362 which should not exist...

From your git shell, run the following:

```
git rebase origin/master
git rebase --skip # at the first sign of conflict, simply delete the offending commit
git push -f <... your configured personal remote and branch...> 
```

elsiehupp:
It looks like I figured it out.

PR #1458, opened 2021-06-20
Update Atlantic Recipes

thekabistro:
- Fix processing of Atlantic recipes as they were producing blank articles
- Seperate out magazine & website logic to respective recipes


kovidgoyal:
I keep them identical deliberately, so changes to one can be made to the other easily. Apart from the index processing the actual article markup processing should be identical in the two recipes. Could you please resubmit your PR with just the changes you need without splitting out the magazine/web parts.

kovidgoyal:
Oh and the issue with the empty articles should be already fixed by https://github.com/kovidgoyal/calibre/commit/6c3f1ebb5f060895bd8dc79f193479ee07d0a26e

PR #1467, opened 2021-06-30
Change MOBI EXTH cdeType metadata to EBOK if it's not EBOK

xxyzz:
Only EBOK type book can use Word Wise and X-Ray. If the book is a personal document or doesn't have a type, it would be meaningless to add the ASIN without changing the 501 record.

### DIFF
--- a/src/calibre/ebooks/metadata/mobi.py
+++ b/src/calibre/ebooks/metadata/mobi.py
@@ -395,7 +395,7 @@ class MetadataUpdater(object):
             update_exth_record((202, pack('>I', self.thumbnail_rindex)))
         # Add a 113 record if not present to allow Amazon syncing
         if (113 not in self.original_exth_records and
-                self.original_exth_records.get(501, None) == 'EBOK' and
+                self.original_exth_records.get(501, None) == b'EBOK' and
                 not added_501 and not share_not_sync):
             from uuid import uuid4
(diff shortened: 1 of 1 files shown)


kovidgoyal:
No, ASINs can be used for syncing as well, even for personal docs. that was the original reason calibre inserted made up asins in the first place.

Hoever, the change from 'EBOK' to b'EBOK' is correct, I have done that

xxyzz:
I changed book type to 'EBOK' only when the `asin` parameter is passed, it won't affect calibre's code.

kovidgoyal:
How are you sure that calibre code never passes in an ASIN?

xxyzz:
Because I added that parameter in https://github.com/kovidgoyal/calibre/pull/1432 two months ago...

kovidgoyal:
Ah, ok then send a new PR for it and I will merge it.

kovidgoyal:
Or force push to this one either way is fine

xxyzz:
But this pull request has no conflicts with the master branch as GitHub says. What should I change with the new commit?

kovidgoyal:
I already merged part of it, in https://github.com/kovidgoyal/calibre/commit/f8930134350fc7eb2e886d08ec143101528226ec

xxyzz:
Thanks!

PR #1475, opened 2021-07-18
Change the documentation for the new tag_browser_context_action() …

cbhaley:
…method so that it is clearer that the method is not required to yield actions.


cbhaley:
BTW: thanks for doing the work to add the feature to the plugin definition.

kovidgoyal:
You're welcome!

PR #1476, opened 2021-07-28
Fix custom series indicies being preserved when a custom series is …

cbhaley:
…cleared, making the column sort correctly. The problem is that the series index is still in the table cache for a book after the series is removed.

See https://www.mobileread.com/forums/showthread.php?p=4142361#post4142361

I freely confess that this might not be the best fix. :)


kovidgoyal:
I changed the implementation slightly, to make it more efficient.

cbhaley:
Thanks. Your implementation is clearer as well.

PR #1479, opened 2021-08-20
Add publish date to title for Economist Recipe

n194:
This PR adds the date to the title of Economist weekly editions for organization purposes. 

Edit: Not sure if this is wanted though because it can break automatic grouping in some e-readers.


kovidgoyal:
recipe titles must no contain dates, they are added automatically by the
recipe download system, unless you have specified the kindle output
profile, since the kindle displays dates automatically.

PR #1481, opened 2021-09-13
Make cache.set_field check if the value of the field is really …

cbhaley:
…changing, improving the performance of cache.set_metadata() that updates all fields.

This change also ensures that change events aren't triggered and books aren't marked as dirty if nothing actually changed.

Checking if a field has changed does introduce some new processing. I don't think it is a large amount and it could be offset by the savings.

### DIFF
--- a/src/calibre/db/cache.py
+++ b/src/calibre/db/cache.py
@@ -1230,8 +1230,8 @@ class Cache(object):
                 else:
                     v = sid = None
                 if sid is None and name.startswith('#'):
-                    extra = self._fast_field_for(sfield, k)
-                    sid = extra or 1.0  # The value to be set the db link table
+                    sid = self._fast_field_for(sfield, k)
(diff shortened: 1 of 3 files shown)


cbhaley:
Kovid, your changes broke db.write.many_many(). The book_id_item_id_map contains empty sets/tuples if nothing is being set for the book. An item is missing in table map (table.book_col_map.get) if nothing is set so the get returns the default, which you changed to 'missing'. As such the != comparison succeeds when in fact nothing changed.

Setting line 442 to the following works.
    book_id_item_id_map = {k:v for k, v in book_id_item_id_map.items() if v != g(k, tuple())}

kovidgoyal:
Thanks, fixed.

kovidgoyal:
You might want to add some tests to db.tests.write to check for dirtying of the various column types.

PR #1485, opened 2021-09-26
Bug #1945098: 5.28 Forgets or reverts identifiers.

cbhaley:
See https://bugs.launchpad.net/calibre/+bug/1945098 for the rationale.


kovidgoyal:
I fixed it slightly differently to avoid a double copy of default_value

PR #1506, opened 2021-10-16
Fix potential crash in ebook-edit due to type mismatch.

sprocklem:
If the ImageDelegate's font has a pixelSize of less than 5, it will calculate a floating point value, from its pointSize. QSize can only take an int, due to its dispatching to a C++ method.

I've changed some graphics settings recently, but am not sure what change caused ebook-edit to crash. It may make more sense to round() the result, but I used flooring division to keep consistent with the few other calculated QSize arguments, which are simply converted to int (by flooring) if they are not already.

### DIFF
--- a/src/calibre/gui2/tweak_book/editor/insert_resource.py
+++ b/src/calibre/gui2/tweak_book/editor/insert_resource.py
@@ -97,7 +97,7 @@ class ImageDelegate(QStyledItemDelegate):
         f = self.parent().font()
         sz = f.pixelSize()
         if sz < 5:
-            sz = f.pointSize() * self.parent().logicalDpiY() / 72.0
+            sz = f.pointSize() * self.parent().logicalDpiY() // 72
         self.title_height = max(25, sz + 10)
         self.item_size = self.cover_size + QSize(2 * self.MARGIN, (2 * self.MARGIN) + self.title_height)
         self.calculate_spacing()


eli-schwartz:
This is similar to #1460 and #1461, because python deprecated and then eventually removed (in python 3.10) the ability to pass floats to a C method and have them be automatically cast to int.

I guess neither of us found this one because it only sometimes happens? Huh.

kovidgoyal:
Yup. The python ecosystem's playing fast and loose with backward compatibility is really starting to get on my nerves.

eli-schwartz:
Well, it's technically a good thing in many cases where the loss of precision is unexpected (rather than Qt's use where it is expected).

Unfortunately, actually getting python to print deprecation warnings can be complicated at times.

kovidgoyal:
On Sat, Oct 16, 2021 at 06:40:09PM -0700, Eli Schwartz wrote:
> Well, it's technically a good thing in many cases where the loss of precision is unexpected (rather than Qt's use where it is expected).

Sure, but you really don't make this kind of change in a point release,
without any tooling to help identify and correct the breakage you cause,
beyond the fairly useless deprecation warning.

If you make a mistake in a public, widely consumed, API, you need to take
some responsibility for it, you don't just fix it and throw it over the
wall whenever you like.

> 
> Unfortunately, actually getting python to print deprecation warnings can be complicated at times.

Deprecation warnings are useless for this kind of thing, because they
only work on executed code paths. In any non-trivial application there
is no way you are going to cover all cases of something like this.

PR #1521, opened 2021-11-06
Bug #1950033: Book Details: Missing copy options on composite columns

cbhaley:
Added the copy options as well as search and remove to composites, bools, numbers, etc

### DIFF
--- a/src/calibre/ebooks/metadata/book/render.py
+++ b/src/calibre/ebooks/metadata/book/render.py
@@ -60,9 +60,9 @@ def search_action(search_term, value, **k):
     return action('search', term=search_term, value=value, **k)
 
 
-def search_action_with_data(search_term, value, book_id, field=None):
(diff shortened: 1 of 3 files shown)


kovidgoyal:
You should make equivalent changes to render_metadata() in
book_details.pyj as well so the metadata display on the content server
also has the same appearance.

cbhaley:
I think I did what you asked for.

Rendering bools does not take the booleans_are_tristate config setting into account. I couldn't figure out how to get that value from db.prefs.

cbhaley:
Note that I did change the behavior. Before it assumed booleans_are_tristate == False. Now it assumes it is True.

kovidgoyal:
I have implemented bools_are_tristate: https://github.com/kovidgoyal/calibre/commit/6322ba9a1b2ba8ae3560702725ebf9504469458f

PR #1522, opened 2021-11-07
manual/customizing: reword intro, mention that plugins can be shared

eli-schwartz:
And while we are at it, add a simple disclaimer that plugins and themes are not part of calibre and support for them is handled externally via their support threads.

### DIFF
--- a/manual/customize.rst
+++ b/manual/customize.rst
@@ -6,10 +6,18 @@ Customizing calibre
 ==================================
 
 calibre has a highly modular design. Various parts of it can be customized. You can learn how to create
-*recipes* to add new sources of online content to calibre in the Section :ref:`news`. Here, you will learn,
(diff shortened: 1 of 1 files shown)


eli-schwartz:
P.S.

I can't help but notice the themes.json (downloaded by the theme chooser) has all kinds of useful metadata in it, including the "url" for each theme, but I cannot actually figure out how to view that url from within calibre. Am I missing something?

The plugin updater *does* provide a link to the "Plugin homepage", and I think that themes should too, for support reasons so that users of the theme can contact the theme author for help.

kovidgoyal:
You are welcome to submit a PR to surface that URL, though other than our late lamented elsie I dont think anyone has ever asked for one.

kovidgoyal:
Actually I had a minute so I implemented it.

PR #1531, opened 2021-11-14
Google books metadata plugin: Workaround for API bug

soilheart:
For some books isbn query key needs to be in uppercase

As discussed here: https://www.mobileread.com/forums/showthread.php?t=342730

Have tested functionality with isbn:978-0-7869-5437-7 (Dragons of Autumn Twilight).


kovidgoyal:
Thanks, I implemented it slightly differently.

PR #1532, opened 2021-11-17
Add Kobo Clara HD and Kobo Aura H20v2 device profiles

BradleyDS2:


kovidgoyal:
There is no need for these, you can just use the tablet or generic eink
hd profiles they will work fine for modern kobos

BradleyDS2:
Tablet profiles result in the images being uncompressed, which is problematic for books that have very high resolution images.
GenericEinkHD has a resolution of 10k x 10k
The next largest is GenericEinkLarge at 600x999. That's about half the resolution of these devices.

There's specific models of other readers such as the SONY PRS-900, what's the harm in adding more variety to better cater to other devices?

If the issue is that it's a specific two models, I'm happy to add the rest of the Kobo models as well.

kovidgoyal:
Because at some point I am going to phase out output profiles entirely.
So there is no point adding more.

BradleyDS2:
Until that happens though, having more profiles would add value, and there's nothing additional to maintain. Is there any downside?

Is an alternative to output profiles going to be provided that allows for automatic image resizing?
If so, it'd be fantastic if the user could specify custom resolutions, instead of making changes to the source to add support and having said changes rejected.

kovidgoyal:
yes, it creates unnecessary churn and a bunch of profiles that dont
really do anything. And yes when I get rid of profiles, you will be able
to specify a max resolution for images.

And for the future, before spending time to make changes to the source,
ask if your changes are likely to be acceptable.

kovidgoyal:
There is no progress to be made it would be small patch to get rid of output profiles, I just have higher priority things to do.

NaruZosa:
Just checking in.
Has there been any progress or do you need any help with this?

kovidgoyal:
No, I dont need help, thanks.

kovidgoyal:
https://github.com/kovidgoyal/calibre/commit/c33df89ce2dc8037146f634ef8d528af3333cb68

PR #1534, opened 2021-11-20
Bug #1951673: Clicking on field Search Mode resets previous field

cbhaley:
Moved the mode field, which as a side benefit makes it more "important".


cbhaley:
I reviewed template_dialog.py. I made cosmetic changes in some variable names to make the code easier to understand and make it consistent with other calibre code. None of these changes affect functionality.

The commit ended up here ...

PR #1540, opened 2021-12-07
Add some guidance about ssl error

vmiheer:


kovidgoyal:
On Windows, setting the SSL_* env vars is not the recommended solution.
Instead the recommended solution is to fix the system certificate store,
which is typically done by visiting the failing site in Edge, which
installs missing certificates for you automatically. However sometimes
it requires more surgery, for example: https://www.mobileread.com/forums/showthread.php?t=342078

vmiheer:
I'll try this solution.

PR #1545, opened 2021-12-11
Fix Pocket/readitlater recipe

grdryn:
Hi! :wave: I think this is my first time interacting with this project, so there's a chance I might've missed something.

This recipe wasn't working (at least for me), where it would consistently fail with the following error when trying to download an article:

```
  File "calibre/utils/threadpool.py", line 100, in run
  File "calibre/web/feeds/news.py", line 1186, in fetch_obfuscated_article
  File "<string>", line 157, in get_obfuscated_article
  File "<string>", line 142, in get_textview
  File "re.py", line 201, in search
TypeError: expected string or bytes-like object
```

I believe this is because Pocket don't allow access to their "Article View" API by default:
https://getpocket.com/developer/docs/v3/article-view

This change uses the original URL of the article, rather than the pocket url for it (those `getpocket.com/a/read/<id>` URLs in the browser seem to just redirect me to `getpocket.com/my-list`). I've a feeling that the old way might have produced cleaner articles (Pocket cleanup & then Calibre cleanup), but I've never seen it work successfully.

We could alternatively try to convince Pocket to enable that API for the "app" we use here (I think this usage would qualify, as it's a "Pocket specific feature" (for Calibre)). That _might_ require adopting their full OAuth flow though, rather than using username/password. From reading their API docs with Calibre in mind, I think that would mean we'd have to have a web page somewhere that the user would have to access in their normal web browser, click a button to redirect them to the Pocket page to authorize the Calibre app to access their Pocket account, which would then redirect them back to our web page where we'd instruct them to enter the access token into the "password" field for the recipe in Calibre.

### DIFF
--- a/recipes/readitlater.recipe
+++ b/recipes/readitlater.recipe
@@ -1,5 +1,5 @@
 """
-Pocket Calibre Recipe v1.4
+Pocket Calibre Recipe v1.5
 """
 from calibre import strftime
 from calibre.web.feeds.news import BasicNewsRecipe
@@ -39,7 +39,7 @@ class Pocket(BasicNewsRecipe):
     # Settings people change
     oldest_article = 7.0
     max_articles_per_feed = 50
-    minimum_articles = 10
+    minimum_articles = 1
     mark_as_read_after_dl = True  # Set this to False for testing
     sort_method = 'oldest'  # MUST be either 'oldest' or 'newest'
     # To filter by tag this needs to be a single tag in quotes; IE 'calibre'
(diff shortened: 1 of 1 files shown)


kovidgoyal:
OAuth has always seemed like too many hoops to me for anything
non-browser based.

grdryn:
@kovidgoyal thanks for reviewing and merging!

I agree that OAuth is a little more inconvenient for the non-browser case, it's a pity it wasn't designed to be easier for those cases. At the same time, it can provide better security: when users get sent to the site that they're granting an app access to, the site will typically make it obvious which permissions the app is going to have (e.g. read-only access to some specific data, rather than giving the full login details). While I currently trust calibre with my login details for some sites, I wouldn't trust other apps with such data; and it would be better if I didn't _have_ to trust/provide any app/site with username/password for any other site.

Thinking about it a little more, if it ever became more compelling for calibre to work with OAuth, then starting an HTTP server on localhost and having that be the redirect URL, might be better than my suggestion above of hosting a web page somewhere else on the internet and using that. It's then kind of just part of the calibre desktop app, you're just interacting with it through a browser.

Anyway, maybe not compelling for now. :slightly_smiling_face:

PR #1547, opened 2021-12-12
Enhancement: add a tweak to provide the sort value for undefined …

cbhaley:
…numbers.

I couldn't find a case where numeric_sort_key() was called with a non-number but I decided to add the test anyway.

The change to tweaks.py fixes something that has annoyed me for a long time. I don't know if the problem exists on other platforms. If it doesn't then perhaps it could be conditioned with an 'if windows' test.

### DIFF
--- a/resources/default_tweaks.py
+++ b/resources/default_tweaks.py
@@ -594,3 +594,16 @@ skip_network_check = False
 # Sets the width of the tab stop in the template editor in "average characters".
 # For example, a value of 1 results in a space with the width of one average character.
 template_editor_tab_stop_width = 4
+
+#: Value for undefined numbers when sorting
+# Sets the value to use for undefined numbers when sorting.
+# For example, the value -10 sorts undefined numbers as if they were set to -10.
+# Use 'maximum' for the largest possible number. Use 'minimum' for the smallest
+# possible number. Quotes are optional if entering a number.
+# Examples:
+#   value_for_undefined_numbers_when_sorting = -100
+#   value_for_undefined_numbers_when_sorting = '2'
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Rather than initializing the constant at import time, initialize it in Field.__init__ that way changing the tweak doesn't require a restart. This is how the tweak for sorting times is also applied so I doubt it makes a significant performance difference.

As for the horizontal scrollbar, I cant get  horizontal scrollbar to show up at all on windows or linux. Since the text in the list of tweaks is word wrapped why would there be a horizontal scrollbar? Do you have a custom tweak with a very long name and no spaces in the name? If that's the case the correct fix would be to change the data() method to force wrap things longer than 40 chars.

cbhaley:
The horizontal scrollbar problem is related to font size. The width of the list and the number-of-characters word wrap point are hard coded, so the bar appears for bigger fonts. I have an implementation that computes the word wrap point based on the text and the font. I will push it shortly.

cbhaley:
As for the tweak, I changed it. However it seems that Preferences / tweaks (Tweaks.commit()) doesn't change the internal tweak dictionary so a restart is still required. This is true for the date tweak as well. However, it is worth doing so changing it can be moved easily to the gui if wanted.

kovidgoyal:
I have merged with a simpler solution for word wrapping in the tweaks
view, letting Qt handle it. The scrollbar doesnt show up for me anymore.

cbhaley:
Your change is much better! However, note that in some cases it uses ellipses instead of wrapping. See image. I am OK with that
Clipboard01
.

PR #1555, opened 2021-12-25
Bug #1955732: Hierarchical entries in user category may not merge …

cbhaley:
…correctly in tag browser

Note that this fix works with grouped search terms. 'Real' user categories will never be merged because the source category is part of the 'name'

### DIFF
--- a/src/calibre/db/categories.py
+++ b/src/calibre/db/categories.py
@@ -44,7 +44,8 @@ class Tag:
 
     @property
     def string_representation(self):
-        return '%s:%s:%s:%s:%s'%(self.name, self.count, self.id, self.state, self.category)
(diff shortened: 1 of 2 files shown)


cbhaley:
Kovid, a similar but not the same problem exists in the web server. Grouped searches act like user categories, but this doesn't work. Reason: duplicate keys are filtered out by get_categories()  so searches must use the gst key not the category key.

Hierarchies also are duplicated as in this bug.

I looked at fixing these but got quite lost in the code. Sorry.

kovidgoyal:
Merged, I dont care enough about this issue to implement it in the
server, if you do am happy to provide any needed guidance. Look first at
get_tag_browser in srv.handler.py, this makes the output of
get_categories available in the server. You will probably need to adjust
the render() method to add the gst metadata and then use it inthe server
pyj code that builds the tress.

PR #1576, opened 2022-01-23
Improve build support for NetBSD

Rhialto:
by adding it to some conditionals.

Hi! Here is a small improvement for building Calibre on NetBSD. It is far from complete but it's a small step. In [pkgsrc](https://github.com/NetBSD/pkgsrc/tree/trunk/misc/calibre) I have created more patches, but not all of them may be appropriate to upstream.

In particular I'm trying to find out why libusb isn't properly scanning usb devices, so if you're open to it, I can possibly supply more patches in the future.

### DIFF
--- a/setup/build.py
+++ b/setup/build.py
@@ -7,7 +7,7 @@ __docformat__ = 'restructuredtext en'
 import textwrap, os, shlex, subprocess, glob, shutil, sys, json, errno, sysconfig
 from collections import namedtuple
 
-from setup import Command, islinux, isbsd, isfreebsd, ismacos, ishaiku, SRC, iswindows
+from setup import Command, islinux, isbsd, isfreebsd, ismacos, ishaiku, SRC, iswindows, isnetbsd
 isunix = islinux or ismacos or isbsd or ishaiku
 
 py_lib = os.path.join(sys.prefix, 'libs', 'python%d%d.lib' % sys.version_info[:2])
@@ -115,7 +115,7 @@ def is_ext_allowed(ext):
     only = ext.get('only', '')
     if only:
         only = set(only.split())
-        q = set(filter(lambda x: globals()["is" + x], ["bsd", "freebsd", "haiku", "linux", "macos", "windows"]))
+        q = set(filter(lambda x: globals()["is" + x], ["bsd", "freebsd", "netbsd", "haiku", "linux", "macos", "windows"]))
         return len(q.intersection(only)) > 0
     return True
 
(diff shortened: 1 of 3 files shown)


eli-schwartz:
As a matter of curiosity, is your libusb issue related to the reason why it doesn't seem to be enabled on *free*BSD either?

kovidgoyal:
I'm fine in principle with adding support for netbsd however I dont understand the purpose of this PR, it introduces no new behavior based on isnetbsd, so what's the point of having that conditional? Is this PR meant to be a first step

Rhialto:
It enables the build of libusb and libmtp for netbsd. The change in `constants.py` and `extensions.py` allows the build, and the other parts are needed to make the netbsd check possible.

 I don't have a device handy to test if libmtp works, and I know that libusb doesn't work, but when it is compiled, it can hopefully be debugged.

Rhialto:
@eli-schwartz I found out that the libusb extension (using the libusb or libusb-1.0 libraries) doesn't find my plugged in reader device (a generic e-ink Android device, from Icarus). If these libraries on FreeBSD are similar enough to have the same issue, I couldn't say.

kovidgoyal:
I wouldnt be willing to enable it if it is known not to work. So to get your PR merged I suggest adding a runtime check for netbsd that disables the libusb/libmtp support at runtime, if you want them not disabled at build time.

Rhialto:
Currently the libusb extension detects no devices, which is kind of the same effect as being disabled, I suppose.

kovidgoyal:
Looking at devices/scanner.py freebsd uses the libusb based scanner. And
there is a note saying that device detection on netbsd is not
implemented. So I dont think building libusb is going to do anything for
you. What you need to do is implement device scanning on netbsd. Without
that, you will get nowhere.

Rhialto:
In a local patch I have
```
if isnetbsd:
    netbsd_scanner = libusb_scanner
```
but apart from that, there is this a few lines later:
```
        if self.scanner is None:
            self.scanner = libusb_scanner
```
so my patch isn't even needed.

Also, if the libusb extension isn't built, and if I remember correctly, something complains that is missing, while starting up calibre. That probably made me look at this in the first place.

The code in the libusb extension does build on netbsd, and does look like it should work.

But I get the feeling I'm too soon with my merge request.
I will try to get the device scanning working and then get back to you.

kovidgoyal:
On Sun, Jan 23, 2022 at 06:05:59AM -0800, Rhialto The M. wrote:
> In a local patch I have
> ```
> if isnetbsd:
>     netbsd_scanner = libusb_scanner
> ```
> but apart from that, there is this a few lines later:
> ```
>         if self.scanner is None:
>             self.scanner = libusb_scanner
> ```
> so my patch isn't even needed.

Indeed, when no system specific scanner is present calibre fallsback to
the libusb based one.

> 
> Also, if the libusb extension isn't built, and if I remember correctly, something complains that is missing, while starting up calibre. That probably made me look at this in the first place.

Yes, there will be an error message if libusb is not available.

> 
> The code in the libusb extension does build on netbsd, and does look like it should work.
> 
> But I get the feeling I'm too soon with my merge request.
> I will try to get the device scanning working and then get back to you.

Indeed if you can get device detection working I will be happy to 
include libusb/libmtp on netbsd.

PR #1578, opened 2022-01-25
Unsmarten guillemetes

Carmina16:
Add «guillemetes» to the unsmarten list.


kovidgoyal:
I'm not sure this is appropriate. unsmarten is supposed to be basically the opposite of smarten. And smarten doesnt know about non-english languages. Replacing quotes in languages other than english with quotes from english doesnt seem right to me.

PR #1591, opened 2022-02-06
Add ability to turn off completion in the EditWIthComplete widget via…

cbhaley:
… a context menu item. Use that ability in edit metadata single & bulk.

In the end I thought it would be easier for you to evaluate this stuff if you could see it.

### DIFF
--- a/src/calibre/gui2/complete2.py
+++ b/src/calibre/gui2/complete2.py
@@ -12,6 +12,7 @@ from qt.core import (
 
 from calibre.constants import ismacos
 from calibre.utils.icu import sort_key, primary_startswith, primary_contains
+from calibre.gui2 import gprefs
 from calibre.gui2.widgets import EnComboBox, LineEditECM
 from calibre.utils.config import tweaks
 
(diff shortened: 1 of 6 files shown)


kovidgoyal:
Whats the motivation for this: Why would one want to disable completion?

cbhaley:
I don't like completion and have wanted to get rid of it for years. Whenever I type something that is a prefix of something else it is hard to get out of the dialog without it completing without that something else. AFAIK my only choices are to click somewhere else or hit escape. The second is a dangerous habit because it can cancel the enture dialog. I have been known to get into a state where I hit escape until the dialog goes away then start over

However, now that I think about it, changing the behavior of tab in Completer would solve my problem more elegantly. A tab pressed while typing currently selects the top item, at which point I am stuck trying to get out of the list widget with what I typed instead of the completion. The fix: change (optionally) the behavior of tab so that if pressed without a selected completion it accepts what you have typed. This could properly be a tweak.

cbhaley:
I am closing this so I can submit the way with the tweak I described above.

cbhaley:
Also, I just found the preselect_first_completion tweak. My change interacts with that tweak by using tab to 'accept' what is typed then add a separator.

PR #1593, opened 2022-02-10
Enhancement #1960442: Remove line in Template functions window

cbhaley:


cbhaley:
I also pushed some changes to the preferences dialog:
1) Make the stored templates dialog show all the selected books instead of only the first.
2) Remember the last state help showing or not for both stored templates and template functions.

PR #1613, opened 2022-04-13
Enhancement #1968810: provide an interface ot mark books with …

cbhaley:
…arbitrary text.

NB: I have no skill with icons nor do I have advanced tools to create them. The one I provided has a circle drawn in the pin, intended to indicate the letter O. It is hard to see but at least it changes the color somewhat. Feel free to use something else. :)


cbhaley:
I changed the icon by flipping it to point the other way, so it is obviously different.

cbhaley:
I pushed a commit to add id_from_url() to the Google metadata source. It doesn't belong in the PR but I apparently can't create another one. Or at least I don't know how.

cbhaley:
One last commit (I hope) -- sometimes the rows weren't refreshed when all marks were removed or a first mark appeared. It usually was the selected rows but not always.

PR #1615, opened 2022-04-15
Mark w/text: Change the way history is shown in the buttons.

cbhaley:
After using it for a bit it I think it makes more sense to put the latest value into the combo box and show up to 4 older values in the buttons. This allows reusing the latest value by pressing return and doesn't duplicate values.


cbhaley:
BTW: the colored pin stuff is nice.

kovidgoyal:
On Fri, Apr 15, 2022 at 11:37:43AM -0700, Charles Haley wrote:
> BTW: the colored pin stuff is nice.

Yes, was easier to implement than I expected.

PR #1621, opened 2022-04-23
Enhancement #1970045: is_marked() text as tooltip

cbhaley:


kovidgoyal:
I have merged, it occurs to me you might want to add an "Unmark this
book" action to the context menu when right clicking a row number.

cbhaley:
I think that opens a can of worms. If Unmark then why not Mark? Why not Edit, view, and all the actions on the main context menu? Given that a right click a few mm away gives all of that, why go here?

kovidgoyal:
Mainly because the pins show up on the rows numbers, the rest of the things dont. But it was just a passing thought, not a big deal either way.

cbhaley:
That makes sense. I'll do it.

cbhaley:
Question: it is trivial (and easier) to make it toggle the mark instead of simply clearing it. Should it do that?

kovidgoyal:
yes, makes sense.

PR #1622, opened 2022-04-24
Set & clear marks on the row label context menu

cbhaley:


cbhaley:
I also added a "Clear marks on selected books" action to the marks action.

PR #1624, opened 2022-04-25
Update New Yorker recipe

yodha8:
* Has correct section names now
* Sections are in right order now
* Mail article is included now


recipes/new_yorker.recipe
```diff
@@ -3,7 +3,8 @@
 # License: GPLv3 Copyright: 2016, Kovid Goyal <kovid at kovidgoyal.net>
 from __future__ import absolute_import, division, print_function, unicode_literals
 
-from collections import defaultdict
+from collections import OrderedDict
+from typing import OrderedDict
```
ping, line 7: It looks to me that the import for typing.OrderedDict is not needed and the 2 imports with the same name is confusing.

I've removed this line locally and recipe works as expected.

PR #1640, opened 2022-05-13
Remove some repeated links to other articles

felagund:
On the web, under some articles, there are links to other articles or sections. These links are not present in the print edition, are repetetive and annoying. For example, soething like this: "For exclusive insight and reading recommendations from our correspondents in America, sign up to Checks and Balance, our weekly newsletter."

This commit removes them, as the ■ symbol marks end of the article.

The only downside is that applying the ■ symbol is not consistent - some articles do not end by it. One could target specific texts to remove as they are repetetive (like the one quoted above that appears under articles in the US sections).

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -161,6 +161,11 @@ class Economist(BasicNewsRecipe):
     keep_only_tags = [dict(name='article', id=lambda x: not x)]
     no_stylesheets = True
     remove_attributes = ['data-reactid', 'width', 'height']
+    preprocess_regexps = [
+        (re.compile(r'<span data-ornament="ufinish">■</span></p>.*</body>', re.DOTALL|re.IGNORECASE),
+        lambda match: '<span data-ornament="ufinish">■</span></p></body>'
+        )
+    ]
     # economist.com has started throttling after about 60% of the total has
     # downloaded with connection reset by peer (104) errors.
     delay = 1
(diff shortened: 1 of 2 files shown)


felagund:
I would understand it if you read the web version through a browser, but a recipe is used to download the weekly edition - possibly for offline use or into a e-reader and there the usefulness is questionable. And following such link will most likely take you from the application (certainly it takes you from Calibre's e-reader), as it will open the link via a browser, so it kind of defeats the purpose of downloading the issue in the first place, no?

I wonder how and in which appliation you read the magazine then.

My usecase is that I want to replicate the paperversion when I am travelling, but I assumed this kind of use is the idea behind recipes.

kovidgoyal:
On Fri, May 13, 2022 at 08:50:50AM -0700, Tomáš Hnyk wrote:
> I would understand it if you read the web version through a browser, but a recipe is used to download the weekly edition - possibly for offline use or into a e-reader and there the usefulness is questionable. And following such link will most likely take you from the application (certainly it takes you from Calibre's e-reader), as it will open the link via a browser, so it kind of defeats the purpose of downloading the issue in the first place, no?

Why? Looking up things in a web browser is no great burden. Even e-ink
Kindles can do that. Not to mention that links that point to other
articles in the same issue are rewritten to go to the downloaded
version.

> 
> I wonder how and in which appliation you read the magazine then.
> 
> My usecase is that I want to replicate the paperversion when I am travelling, but I assumed this kind of use is the idea behind recipes.

There is no reason to slavishly follow the paper version. If something
has a use beyond the paper version, there is no reason not to have it.

felagund:
Hm, my quite new Pocketbook 741 that even has colour can do that too but using its browser is pain in the ass. For me, these  texts increase the cognitive burden as I have to skip them (ok, it is not a huge deal). When they are links to other article in the issue, I will get to them anyway (but I can see that not everyone reads it in a linear fashion). Especilly the recurent one advertising various columns or megatopic like Ukraine or covid are of limited use I think. However, I can see that our mileage varies considerably.

PR #1641, opened 2022-05-13
Template language for i in range() with documentation.

cbhaley:


cbhaley:
Added a commit to improve user debugging of range()

PR #1642, opened 2022-05-13
Set per issue title

felagund:
This modifies the title of the downloaded e-book for the economist with the date. It is useful for telling appart different issues in calibre, in the filesystem (the filename is taken from the titile) and it also makes life easier for some e-readers, like this:
https://www.mobileread.com/forums/showthread.php?t=346484

I am sure, python-wise, this can be more elegant but this seems to work. The title of the downloaded magazine is then "The Economist from 2022-05-14" or something like that.

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -249,6 +249,7 @@ class Economist(BasicNewsRecipe):
             self.timefmt = ' [' + edition_date + ']'
         else:
             url = 'https://www.economist.com/printedition'
+        self.title = self.title + " from " + self.browser.open(url).geturl().split("/")[-1]
         raw = self.index_to_soup(url, raw=True)
         # with open('/t/raw.html', 'wb') as f:
         #     f.write(raw)
--- a/recipes/economist_free.recipe
+++ b/recipes/economist_free.recipe
@@ -249,6 +249,7 @@ class Economist(BasicNewsRecipe):
             self.timefmt = ' [' + edition_date + ']'
         else:
             url = 'https://www.economist.com/printedition'
+        self.title = self.title + " from " + self.browser.open(url).geturl().split("/")[-1]
         raw = self.index_to_soup(url, raw=True)
         # with open('/t/raw.html', 'wb') as f:
         #     f.write(raw)


kovidgoyal:
Download in EPUB not MOBI and the dates will be added to the title
automatically, in the format specified by datefmt in the recipe. For
MOBI they are deliberately removed since that is optimal for usage on
Kindles.

felagund:
That does not seem to me to be the case at all. See screenshot. I am talking about the title of the whole ebook, not individual articles. Right now, if I download Economist issues every week in calibre and keep them there, they canno be distinguished by title (one needs to open the cover and see the date on it). The date metadata refers (as far as I can see) to the date of the download, not the date of the issue, which would also be confusing for telling them appart.
Screenshot_20220514_234759

kovidgoyal:
Again you need to change the download format to EPUB and the output
profile to non-Kindle.

felagund:
Indeed. By the way, it has nothing to do with EPUB as a download format - Calibre will happily add date to EPUB too.

However, it adds the date of download (May 15), not the day when it was issued (May 14). This would be especially confusing if one was using the support for downloading old issues on lines 20 and 247-249.

I noticed there is a function `publication_date()` (undocumented here: https://manual.calibre-ebook.com/news_recipe.html ) but when I tried to modify that, it only changes `<dc:date>2020-01-02T00:00:00+00:00</dc:date>` in `content.opf` in the resulting EPUB file.

Would not some sort of API for setting the date used in the title with non-kindle-output be useful?

Another thing is that non-kindle output is putting text like this:

    This article was downloaded by calibre from https://www.economist.com/the-world-this-week/2022/05/14/politics
    | [Section menu](clbr://internal.invalid/feed_0/index.html#article_0) | [Main menu](clbr://internal.invalid/index_u1.html#feed_0) |

in the ebook, which I do not want. The last time I checked, the only way around it to use a kindle output format, no?

If there was a way to either turn on "date in title" for Kindle or  turn off the text for non-Kindle output, I would be happy with the current functionality.

kovidgoyal:
On Sun, May 15, 2022 at 03:46:54AM -0700, Tomáš Hnyk wrote:
> Indeed. By the way, it has nothing to do with EPUB as a download format - Calibre will happily add date to EPUB too.
> 
> However, it adds the date of download (May 15), not the day when it was issued (May 14). This would be especially confusing if one was using the support for downloading old issues on lines 20 and 247-249.
> 
> I noticed there is a function `publication_date()` (undocumented here: https://manual.calibre-ebook.com/news_recipe.html ) but when I tried to modify that, it only changes `<dc:date>2020-01-02T00:00:00+00:00</dc:date>` in `content.opf` in the resulting EPUB file.
> 
> Would not some sort of API for setting the date used in the title with non-kindle-output be useful?

Set timefmt in the recipe in parse_index.

> 
> Another thing is that non-kindle output is putting text like this:
> 
>     This article was downloaded by calibre from https://www.economist.com/the-world-this-week/2022/05/14/politics
>     | [Section menu](clbr://internal.invalid/feed_0/index.html#article_0) | [Main menu](clbr://internal.invalid/index_u1.html#feed_0) |
> 
> in the ebook, which I do not want. The last time I checked, the only way around it to use a kindle output format, no?
> 
> If there was a way to either turn on "date in title" for Kindle or  turn off the text for non-Kindle output, I would be happy with the current functionality.

No, there isnt. You are free to run from source with a patch.

felagund:
> Set timefmt in the recipe in parse_index.

I do not need to format it. I would like (and I think that should be the default anyway) the date in the title set for the date of the issue of the magazine (that is what my PR did). It would be useful especially when downloading older issues.

Would you accept a patch that would modify  create_opf ( https://github.com/kovidgoyal/calibre/blob/ae0f3a86dbb11326ca64843ac8c79837d8b84ab6/src/calibre/web/feeds/news.py#L1460 ) to use publication_date if set byt the recipe? If publication_date is not set, it default to now in both cases anyway?




> No, there isnt. You are free to run from source with a patch.
Would you accept a patch that would make the navbar optional via a setting?

kovidgoyal:
On Sun, May 15, 2022 at 12:48:00PM -0700, Tomáš Hnyk wrote:
> > Set timefmt in the recipe in parse_index.
> 
> I do not need to format it. I would like (and I think that should be the default anyway) the date in the title set for the date of the issue of the magazine (that is what my PR did). It would be useful especially when downloading older issues.

So you set timefmt to whatever the issue date is.

> 
> Would you accept a patch that would modify  create_opf ( https://github.com/kovidgoyal/calibre/blob/ae0f3a86dbb11326ca64843ac8c79837d8b84ab6/src/calibre/web/feeds/news.py#L1460 ) to use publication_date if set byt the recipe? If publication_date is not set, it default to now in both cases anyway?

Sure.

> 
> 
> 
> 
> > No, there isnt. You are free to run from source with a patch.
> Would you accept a patch that would make the navbar optional via a setting? 

No.

felagund:
> Would you accept a patch that would modify create_opf ( https://github.com/kovidgoyal/calibre/blob/ae0f3a86dbb11326ca64843ac8c79837d8b84ab6/src/calibre/web/feeds/news.py#L1460 ) to use publication_date if set byt the recipe? If publication_date is not set, it default to now in both cases anyway?
> Sure.


https://github.com/kovidgoyal/calibre/pull/1661

PR #1651, opened 2022-06-03
Bug #1977496: shortcut for OK buttons

cbhaley:


cbhaley:
FWIW: I tried changing the focus to the button box. Unfortunately I couldn't see the focus change (calibre light theme) so I couldn't tell that something happened. Thus CTRL-Return.

kovidgoyal:
Focus change on Tab is not desired anyway, since this is plain text and
tab characters are significant.

PR #1652, opened 2022-06-05
Enhancement #1977681: additional display options for yes/no columns

cbhaley:
Radio buttons that control showing icon, text, or both


cbhaley:
Second commit: I forgot the keyboard accelerators on the radio buttons

cbhaley:
The accelerator on "Show:" doesn't do anything, at least on windows.

Sorry about the non-translated strings.

kovidgoyal:
No worries, and yes I think it needs the widget added in add_for to set a focus proxy to a radio button for it to work. I mostly jsut added it to avoid having someone complain to me that one label doesnt have an accelerator.

PR #1653, opened 2022-06-05
Make the accelerator on &Show work.

cbhaley:
I am a bit concerned about why shift-tab works differently from tab. It clearly does, but perhaps because of a bug in Qt? In any event, without the code one gets stuck in the group box.


cbhaley:
BTW: setting the first radio button as the focus proxy for the group box works, but in a strange way. It gives the focus to that button even if it isn't selected. Left and right arrow then work, but they don't select the moved-to button until you pass over the selected button, at which point selection moves with the focus.

I think this is strange behavior, but if you prefer I can implement that instead of trapping focus events as I do in the submitted code.

cbhaley:
The last commit works the way I want, ensuring the checked button is focused while using 'vanilla' Qt. It doesn't work around the odd behavior of backtab. It does this by dynamically changing the focus proxy.

Apologies for the multiple commits. I don't know what happens if I remove commits in a pull request so I left them all.

PR #1658, opened 2022-06-15
Fix a potential wrong operator bug

LostBenjamin:
Hi,

This pull request is a fix to a potential wrong operator bug at `src/calibre/gui2/tweak_book/reports.py`. Please check the changes.

`self.rules` seems a tuple so using `in` would be more reasonable.

Best,
Jingxuan


kovidgoyal:
No, is is correct. internalPOinter() points to container objects, the topmost one being self.rules, so we test for object equality.

PR #1659, opened 2022-06-15
Update fonts MIME following IANA recommendation

lewer:
Hello.

MIME types for fonts are outdated.

According to https://www.iana.org/assignments/media-types/media-types.xhtml#font, MIME for ttf, otf and woff fonts should be respectively **font/ttf**, **font/otf**, **font/woff**.

These MIME are supported by EPUB 3 (https://www.w3.org/publishing/epub3/epub-spec.html#sec-core-media-types)

This PR changes MIME types accordingly.

### DIFF
--- a/resources/mime.types
+++ b/resources/mime.types
@@ -1378,10 +1378,11 @@ application/x-cbc                          cbc
 application/x-koboreader-ebook             kobo
 image/wmf                                  wmf
 application/ereader                        pdb
-# See http://idpf.org/epub/30/spec/epub30-publications.html#sec-core-media-types
-application/vnd.ms-opentype                otf
-application/font-woff                      woff
-application/x-font-truetype                ttf
+# See https://www.w3.org/publishing/epub3/epub-spec.html#sec-core-media-types
+font/otf                                   otf
+font/woff                                  woff
+font/woff2                                 woff2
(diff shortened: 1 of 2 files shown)


lewer:
Ok.

I'm trying to prevent the Check book tool of E-book editor from warning me 

> The file foo.ttf has its MIME type specified as font/ttf in the OPF file. The recommended MIME type for files with the extension "ttf" is application/vnd.ms-opentype.

Could I leave **OEB_FONTS** as it is, and replace the `if ans in {'application/x-font-truetype', 'application/vnd.ms-opentype'}` with something like:
```
if ans in {'font/woff', 'font/ttf', 'font/otf'}:
    versioned_mime = {
        (3,1) : {
            'font/ttf' : 'application/font-sfnt',
            'font/otf' : 'application/font-sfnt',
            'font/woff' : 'application/font-woff'
        },
        (3,0) : {
            'font/ttf' : 'application/vnd.ms-opentype',
            'font/otf' : 'application/vnd.ms-opentype',
            'font/woff' : 'application/font-woff'
        },
        ...

    opfversion = self.opf_version_parsed[:2]
    return versioned_mime[opversion][ans]
```

else, how can I fix the issue?

PR #1675, opened 2022-07-03
The template database functions we discussed.

cbhaley:


kovidgoyal:
No need, just do

get_gui.ans = some_object_with_a_current_db attribute that returns the db

PR #1676, opened 2022-07-04
Tests for the new formatter db functions, and fixes for the errors …

cbhaley:
…that the tests found.

I left in the code I used to select specific tests. Yank it out if you don't want it.


cbhaley:
I also just pushed fixes for typos I noticed in the manual.

kovidgoyal:
I added an entry point to easily run tests by name using calibre-debug

PR #1677, opened 2022-07-04
One difference: using the new run_test() I see all the debug output …

cbhaley:
…where before I didn't, for example exception stack traces in the formatter. Not a big deal, but it does make reading the output more difficult with the extraneous traces. This push fixes that.

NB: I just discovered the hard way that when I do a
    from calibre.constants import DEBUG
I don't see changes to DEBUG caused by calls to calibre.constants.debug(False). Instead I must
    from calibre.constants import is_debugging()
and use that.

### DIFF
--- a/src/calibre/utils/formatter.py
+++ b/src/calibre/utils/formatter.py
@@ -13,7 +13,7 @@ from functools import partial
 from math import modf
 
 from calibre import prints
-from calibre.constants import DEBUG
+from calibre.constants import is_debugging
 from calibre.ebooks.metadata.book.base import field_metadata
 from calibre.utils.formatter_functions import formatter_functions
 from calibre.utils.icu import strcmp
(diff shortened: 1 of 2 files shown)


kovidgoyal:
On Mon, Jul 04, 2022 at 07:49:27AM -0700, Charles Haley wrote:
> …where before I didn&#39;t, for example exception stack traces in the formatter. Not a big deal, but it does make reading the output more difficult with the extraneous traces. This push fixes that.

That's because I set buffer=False, this means output from print
statements is immediately visible, which is useful when developing
tests. With buffer=True such output is shown only if a test fails.

> 
> NB: I just discovered the hard way that when I do a
>     from calibre.constants import DEBUG

Yes, using a constant for this was a mistake from ye very old days.
However, a better fix for this would probably be to change the value of
debug before running find_tests() then in most places importing DEBUG
will get the right value.

cbhaley:
Moving the debug() works. I will make a new commit. I won't bother to undo the changes in the formatter.

PR #1685, opened 2022-07-12
The formatter functions database code we have been discussing

cbhaley:


kovidgoyal:
I dont understand why get_database() needs to be a context manager?
Since nothing is done after the yield anyway, why not just make it a
normal function returning db?

cbhaley:
I wanted the db reference to be guaranteed to go away, which it will do when the 'with' is exited.

I have no problem returning the db if that is what you would rather.

BTW: I pushed fixes for the failing test.

kovidgoyal:
maybe I am missing something, but why would it go away? For example:                                                                           
                                                                                                                                                   
with get_database() as db:                                                                                                                         
  pass                                                                                                                                             
print(db)                                                                                                                                          
                                                                                                                                                   
db is accessible outside the with block. Not to mention that, the code                                                                             
in the block could simply push the db into a list or dict or whatever                                                                              
to keep a reference to it.

cbhaley:
Clearly I misunderstood how they work. I thought the "as var" had the lifetime of the with.

I will change it then make a comment saying I did.

cbhaley:
Changes from a context manager to a function pushed.

kovidgoyal:
On Tue, Jul 12, 2022 at 07:10:02AM -0700, Charles Haley wrote:
> I wanted the db reference to be guaranteed to go away, which it will do when the 'with' is exited.

Umm maybe I am missing something, but why would it go away? For example:

with get_database() as db:
  pass
print(db)

db is accessible outside the with block. Not to mention that, the code
in the block could simply push the db into a list or dict or whatever
to keep a reference to it. 

> 
> I have no problem returning the db if that is what you would rather.

Yes, unless I am missing something about the with statement.

cbhaley:
I am confused. Why this comment? The changes you asked for were pushed and accepted months ago.

Perhaps an email sitting in an outbox queue?

kovidgoyal:
yes this mail was definitely not intended to be sent.

PR #1689, opened 2022-07-15
Yet another enum change problem. Sigh.

cbhaley:
See https://www.mobileread.com/forums/showthread.php?t=347919

### DIFF
--- a/src/calibre/gui2/custom_column_widgets.py
+++ b/src/calibre/gui2/custom_column_widgets.py
@@ -1280,10 +1280,10 @@ class BulkSeries(BulkBase):
 
     def getter(self):
         n = str(self.main_widget.currentText()).strip()
-        autonumber = self.idx_widget.checkState()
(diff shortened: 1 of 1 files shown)


cbhaley:
BTW: I wonder if line 33 in gui2.convert.heuristics has the same problem. Should the existing code:
        self.enable_heuristics(self.opt_enable_heuristics.checkState())
be
        self.enable_heuristics(self.opt_enable_heuristics.isChecked())
?

Same question for txt_input.py line 50.

kovidgoyal:
The enable_heuristics() function doesnt actually use the state
parameter. The txt input one is indeed a bug though one cant use
isChecked() as its a QListWidgetItem which doesnt have isChecked. Have
committed a fix. And another for one in the check library dialog. There
were a whole bunch of these I made a few days ago in 59541e36a5 but I
guess I missed a few checkState() calls.

PR #1690, opened 2022-07-16
Low effort port to charset_normalizer

hroncok:
See https://github.com/PyYoshi/cChardet/issues/77

Note that I have not updated `setup/arch-ci.sh` and `bypy/sources.json`, as I don't really understand the files.


kovidgoyal:
That's a pure python implementation. The whole point of using cchardet
is that it's a lot faster than pure python implementations (calibre used
to use the python chardet package). If cchardet is not being maintained,
I will simply fork it and maintain it myself as I do for many other
python packages.

kovidgoyal:
I have now switched calibre to directly using uchardet, which is what cchardet is an overly elaborate wrapper of.

hroncok:
Thank you, that gets the job done!

PR #1691, opened 2022-07-16
Fix leaving bits of the background selection sometimes visible when e…

cbhaley:
…diting sort and link columns. This change makes the editor widget fill the underlying cell.

As an aside, shouldn't EnLineEdit do setClearButtonEnabled(True) to be like other widgets? If no, then perhaps I should add it in this delegate. The EditWithComplete widgets in the authors column have it.


kovidgoyal:
As I recall I didnt put it in EnLineEdit because thats used all over the
place. Instead every individual use site has to decide if it wants it.

PR #1713, opened 2022-08-10
Support import ebook using localized path and name

noogel:
ASCII path is used by default, and localization path is also supported.


kovidgoyal:
The filenames in the calibre library are kept ASCII only by design, for
maximum robustness across platforms. This is not going to change.

PR #1717, opened 2022-08-19
Enhancement 1984017 - Add Apply button to Manage Tags

cbhaley:
There are two requests here.
1) Add apply button. Because the work is done after the dialog has closed, changing this is more work than I am willing to do.

2) Add toolbar / shortcut / context menu support for Manage categories. Done.

### DIFF
--- a/src/calibre/customize/builtins.py
+++ b/src/calibre/customize/builtins.py
@@ -1063,6 +1063,13 @@ class ActionMarkBooks(InterfaceActionBase):
     description = _('Temporarily mark books')
 
 
+class ActionManageCategories(InterfaceActionBase):
+    name = 'Manage categories'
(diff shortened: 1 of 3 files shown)


cbhaley:
Your change to use QToolButton.ToolButtonPopupMode.InstantPopup removed the ability to make a shortcut key. Any way to get it back?

kovidgoyal:
You would have to do what you did in show_menu() I think. Although that
doesn't really work with shortcuts. What if the button is hidden because
there are too many buttons on the toolbar? Or its not added to a toolbar
but instead to a context menu. On shortcut trigger you would need to
popup a dialog with the choices, it should be easy to write some code
that will iterate over the actions in the menu and create buttons in a
dialog or something similar form them.

PR #1720, opened 2022-08-23
Added way to make custom labeled page break in APNX file using ARIA

Vasolik:
Hi,

I am not a python programmer (and this is the first time I was writing any python code) so please inspect my PR. I tried my best to write code as clean as possible, but I might not know some python quirks so I might make some mistakes there. Anyway, this PR should add the following feature (I tested it and it is working perfectly for me):

Added way to make custom labeled page break in APNX file using ARIA ( Accessible Rich Internet Applications) tag in html.

Using <any_html_element *** pagebreak *** aria-label="some_label" ***> will produce page break location.
Possible usage
```
<span id="pg159" role="doc-pagebreak" aria-label="159"/>
<h1 id="pg_header1" role="doc-pagebreak" aria-label="Header 1">Header</h1>
<p role="doc-pagebreak" aria-label="§ 1 part 4 page 6 of 9">Text</p>
```


kovidgoyal:
I dont think adding custom markup is the right way to go for this.
Instead, what should be done is add support to the MOBI/AZW3 output
plugin to translate page-map.xml and the epub3 page map standard to a
PAGE record in the output file. The apnx code can then use this PAGE
record. See https://wiki.mobileread.com/wiki/APNX

Vasolik:
As I understood aria-labels are the standard way (kinda) to label non-visible elements in Html https://www.aditus.io/aria/aria-label/ . So that is the reason I chose that kind of label markings.

Adding an additional option to generate those custom labels from page-map.xml is also a nice addition. Is there a way to get file text location from page-map.xml? For APNX I would need such information.

But I would argue that we should keep also aria-labels marking solution as well because marking the place with one tag in html is much simpler than generating page-map.xml for manual editors. And aria-label is not out of the blue html attribute :). 

Even if we reject aria-labels part of the change (by commenting it out or deleting it), can we still accept this PR, because it contains refactoring to that section of the code which makes it more easier to expand with additional ways of generating APNX markings? Just populate the Pages class object with data, and you can generate APNX file.

Vasolik:
One more resource I chose aria-labels: http://kb.daisy.org/publishing/docs/html/dpub-aria/doc-pagebreak.html#

kovidgoyal:
I am fine with merging this PR for code re-organization. However, I do
think that having two ways to do the same thing is not optimal. Since
EPUB has settled on <nav> in EPUB 3 and (page-map.xml) in EPUB 2 for
this information. I think we should stick with that.

So I suggest you remove the code for reading the aria roles and rebase
your commits onto master, then I will merge and we can discuss how to
implement PAGE support from page-map.

page-map.xml if just a list of links to anchors in the html file. You 
just get that list, find the anchors and assign page numbers, it should
be just about as easy as your existing solution. You then store it in
the PAGE section of the MOBI file and the APNX code can be greatly
simplified, to just look for and use that information when sending the
file to the device.

Vasolik:
I removed aria pagebreak code and rebased.

kovidgoyal:
I looked over this, some comments:

1) Dont inherit from str and enum. You can get the str value of any enum member by doing member.value

2) Currently calibre is compatible with python 3.7, which does not support type subscripts on builtin container types. So use from typing import List, Dict, etc.

3) I am not a fan of staticmethod. Just make those top level functions in the module or in another module.

4) There is an unused import fo AriaPageBreakGenerator left over

5) In EXTRA_CUSTOMIZATION_CHOICES the call to APNXBuilder.generators.keys() should be wrapped in set() as keys returns an iterator not a concrete type

Other than these it looks ok, I havent reviewed the actual algorithms as I assume they are just copied form the original code, if that's not the case let me know and I will review them as well.

Vasolik:
1. I think that I tried with that, but I was not able (I think) to assign a noninteger value to enum in that case. I will recheck it.
2. I will add this. But I think I read somewhere that one more feature I have used was from 3.10. I am pretty new to python (this is the only code I have ever written in it) so I do not remember what it was and where.
3. This was an organizational thing, but I will move it.
4. Will be removed.
5. I am not sure what should I do here. Call set(APNXBuilder.generators.keys())

Algorithms are not changed, except that results are wrapped in Pages object.
I will fix these after work.
PS: Is there a way to make generators registrable by add-on? Would that be ok?

kovidgoyal:
On Wed, Sep 14, 2022 at 02:16:46AM -0700, Vaso Peras-Likodrić wrote:
> 1. I think that I tried with that, but I was not able (I think) to assign a noninteger value to enum in that case. I will recheck it.


class X(Enum):
   val1 = 'whatever'
   val2 = 'foo'

X.val1.value == 'whatever'

> 2. I will add this. But I think I read somewhere that one more feature I have used was from 3.10. I am pretty new to python (this is the only code I have ever written in it) so I do not remember what it was and where.

I didnt spot anything else but then I didnt read every line of code in
detail. We will see.

> 3. This was an organizational thing, but I will move it.
> 4. Will be removed.
> 5. I am not sure what should I do here. Call set(APNXBuilder.generators.keys())

Yes that's all that's needed.

> 
> Algorithms are not changed, except that results are wrapped in Pages object.
> I will fix these after work.
> PS: Is there a way to make generators registrable by add-on? Would that be ok?

You mean via calibre plugins? If so a fair bit of work will be required
as you will need a new plugin type.

Vasolik:
PR updated.
I changed the interpreter in Pycharm to 3.7 and I noticed that piped unions are also a feature of the later python version. So I changed those as well.

I think I addressed all other issues as well.

PR #1723, opened 2022-08-25
Prefs / Look & feel / Tag browser partition and hierarchy: change …

cbhaley:
…lists of keys to list widget with check boxes.

NB: the partition list was a gpref, which was wrong. It is now a db pref. I added migration code to the tag browser. It is possible that the server code must be changed to get the db pref (if it exists). I don't know how to do that, if it is needed.

### DIFF
--- a/manual/gui.rst
+++ b/manual/gui.rst
@@ -386,7 +386,7 @@ Two variants of equality searches are used for hierarchical items (e.g., A.B.C):
 
 *'Regular expression' searches*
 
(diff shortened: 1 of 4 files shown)


cbhaley:
Good.

I made the comment because I saw this on line 595 of srv.metadata.py
prefs = {
        'tags_browser_category_icons':gprefs['tags_browser_category_icons'],
        'tags_browser_collapse_at':opts.collapse_at,
        'tags_browser_partition_method': opts.collapse_model,
==> 'tag_browser_dont_collapse': opts.dont_collapse,
        'tag_browser_hide_empty_categories': opts.hide_empty_categories,
    }

kovidgoyal:
That comes for the query URL (which is used to get data to build the
display in the browser). The browser view has its own notion of collapse
that does not match that of the tag browser in the main program which
you can set via the preferences in the browser.

But thanks, this reminded me to check and I found that
tag_browser_category_order did indeed need to be migrated.

PR #1724, opened 2022-08-26
Three things related to tag_browser_category_order:

cbhaley:
1) do a proper migration of the tweak
2) Fix the server's usage of the migrated value
3) Fix an exception in the server related to hierarchical user categories


cbhaley:
The db pref is an ordered list of categories, not a dict of category w/sort order numbers like the tweak.

PR #1729, opened 2022-08-30
In Preferences: add Ctrl-Up and Ctrl-Down as keyboard shortcuts for …

cbhaley:
…lists with movement buttons.


cbhaley:
FWIW: I tried adding eventHandlers but they never saw keyboard events. The subclasses did.

PR #1734, opened 2022-09-10
Add strcmpcase() to template language

un-pogaz:
Following a [little discusion](https://www.mobileread.com/forums/showthread.php?t=348976) with @cbhaley, here is the implementation of a string comparator case-sensitive.
Although it is a marginal case, providing a build-in function would be useful and will be complementary the default behavior of calibre.


kovidgoyal:
This PR appears to be empty. Maybe you can re-create it witha  new PR

un-pogaz:
Oh, sorry.
It may not have been the best time to experiment and understand the Pull request, branche, the workflow of fork, etc. 😑
[Here the new](https://github.com/kovidgoyal/calibre/pull/1736)

PR #1738, opened 2022-09-16
Formatter function to_url()

un-pogaz:
return the string in a compliant URI Syntax (Percent-Encoding).

In the same way of the `to_the()`, this is useful when constructing calibre URLs, but in a more human like format.


kovidgoyal:
A proper name for such a function would be
quote_url_path()

that is what quote() does. URL's have different quoting rules for
different parts of the URL. quote() is meant just for quoting the path
components, it cannot be used, for example, to quote the URL query or
the hostname.

PR #1740, opened 2022-09-18
Enhancement #1989813: Option to quickly select filtered tag browser …

cbhaley:
…categories

Done using a new Virtual Field, in_tag_browser that is searchable using 'in_tag_browser:true' (or false or ...).

The feature is discoverable through a new menu item in Configure tag browser: Filter book lists. This menu item can be 'clicked' using a shortcut key.


cbhaley:
Pushed fixes for failing legacy tests

PR #1744, opened 2022-09-23
Use smooth scroll in viewer on mouse wheel scroll

yutotakano:
## Summary

Currently, the calibre viewer has smooth scrolling if you press the Up/Down keyboard keys, but not if you scroll using your mouse wheel. This is regardless of the orientation of the text. This PR attempts to use the same smooth animating system as keyboard press for mouse wheel events. This was previously requested [here](https://www.mobileread.com/forums/showthread.php?t=238177).

## Implementation

Instead of calling `scroll_by_in_block_direction` on mouse scroll, we now call into the ScrollAnimator class that is used already by the Up/Down keyboard shortcuts. To specify how much we want to scroll by, a new argument was added to `scroll_animator.start()`.

Furthermore, `scroll_by_and_check_next_page` checks whether the scroll resulted in a chapter flip. Unfortunately we can't use the same function anymore since by using ScrollAnimator, we lose access to calling arbitrary checks upon finishing the scroll. Fortunately though, ScrollAnimator already checks whether we need to move to the next/previous spine item, so we can remove the need to call `scroll_by_and_check_next_page`!

## Tested

Tested scrolling with keyboard and mouse on one vertical and horizontal book each.

### DIFF
--- a/src/pyj/read_book/flow_mode.pyj
+++ b/src/pyj/read_book/flow_mode.pyj
@@ -151,7 +151,15 @@ def flow_onwheel(evt):
     if Math.abs(di) >= 1:
         scroll_viewport.scroll_by_in_inline_direction(di)
     if Math.abs(db) >= 1:
-        scroll_by_and_check_next_page(db)
+        # Scroll in the direction of the block.
+        # The sign of db determines whether to scroll up or down, however
+        # scroll_animator expects a separate argument to specify this. This
+        # argument is also used for determining whether to go to the next or
+        # previous spine item at the ends, so it's important to set them correctly.
+        if db < 0:
+            scroll_animator.start(DIRECTION.Up, False, -db)
+        else:
+            scroll_animator.start(DIRECTION.Down, False, db)
 
 @check_for_scroll_end
 def goto_boundary(dir):
@@ -189,7 +197,7 @@ def is_auto_scroll_active():
 
 
 def start_autoscroll():
-    scroll_animator.start(DIRECTION.Down, True)
+    scroll_animator.start(DIRECTION.Down, True, line_height())
 
 
 def toggle_autoscroll():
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This makes scrolling with a touchpad much worse, as noted in the original MR thread. If you want to do this, it should only be done for DOM_DELTA_LINE. And possibly even then have an option to turn it off/on in the scrolling behavior section of the viewer preferences.

yutotakano:
Ouch, I hadn't tested with a touchpad. I'll look into adding a config option as you say, and likely limit it to _LINE only. May I ask why you think smooth scroll shouldn't be used with DOM_DELTA_PAGE? Presumably, that can also benefit from smooth scroll.

kovidgoyal:
What would page based smooth scrolling look like? You want to animate
the page transition? You certainly cant turn pages continuously, since
you have to give the user time to read the page.

yutotakano:
I've tested and researched a bit more and it seems that both my mouse and trackpad emit a DOM_DELTA_PIXEL, and in fact, based on some reading, it seems Chromium (which is underneath QtWebEngine) will *always* emit either a calculated DOM_DELTA_PIXEL or a DOM_DELTA_PAGE regardless of how the OS is configured to send scroll events ([Chromium source code for wheel event](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/events/wheel_event.cc) & [discussion on Mozilla about how they are the only browser using DOM_DELTA_LINE](https://bugzilla.mozilla.org/show_bug.cgi?id=1392460)). 

There is no way to detect whether a mouse or trackpad is being used at all.

We can have an option to enable smooth scroll on all possible DELTA modes (and mouse users can opt-in to this), but I feel it might be better to abandon this PR.

kovidgoyal:
OK closing

PR #1747, opened 2022-10-01
Accessibility enhancement: add an option to Prefs / Look & feel / …

cbhaley:
…Main interface to open the book list column context menu with Alt+Up arrow. Default: False


cbhaley:
Kovid, I have a version that uses actions and thus user-settable shortcuts etc.
If you would rather have that then the implementation in this PR, cancel this PR and I will resubmit.

cbhaley:
It also lets the user put the option on the tool bars, etc.

cbhaley:
Thinking about it, the new implementation really is better. I am closing this PR and will resubmit.

PR #1749, opened 2022-10-03
More accessibility changes related to the book list header context menu:

cbhaley:
1) add a dialog to set the width without using the mouse.
2) Change the name from "Book list context menu" to "Book list header menu" to avoid confusion with the context menu on book list cells.


cbhaley:
I picked QGridLayout because QFormLayout doesn't align the buttons, making them look strange. Oh well...

kovidgoyal:
Here you go: 
https://github.com/kovidgoyal/calibre/commit/cdcc20228c0c5e31ac58720e2d6351e10cd0ff56

cbhaley:
That is a clever solution. Thanks,

PR #1751, opened 2022-10-03
Changes to the column width dialog suggested in …

cbhaley:
…https://www.mobileread.com/forums/showthread.php?t=349471

- Put all the resize actions in the dialog.
- Moved menu lines around to make more sense (to me)
- Added some tooltips

BTW: your "addRow()" worked perfectly when given a button as its only argument. :)


cbhaley:
Apparently un_pogaz submitted the same files.

cbhaley:
Sorry, he submitted a different icon, a replacement for the "mouse" used for the context icon.

PR #1752, opened 2022-10-04
Bug #1991571: 'Clear' button on short longtext columns. Changed look …

cbhaley:
…to be like the other widgets.

### DIFF
--- a/src/calibre/gui2/custom_column_widgets.py
+++ b/src/calibre/gui2/custom_column_widgets.py
@@ -138,10 +138,11 @@ class Base:
 class SimpleText(Base):
 
     def setup_ui(self, parent):
-        self.editor = QLineEdit(parent)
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Merged, however, I notice custom series fields dont have a clear button
either.

kovidgoyal:
Sorry, meant to say series index

cbhaley:
That too was intentional. I don't know what it means to clear the series_index. It iis a SpinBox that sn't allowed to be empty. A clear could set it to zero I suppose.

kovidgoyal:
I dont much care, though if you do implement it, it should set to 1 I
think as that is the default series index value.

cbhaley:
I am not going to do it. Two reasons. 1) The code isn't set up to deal with toolbuttons on the second line of 2-line widgets. Significant work (and risk) is required for something I think is minor. 2) The clear button on the Series clears the series_index setting it to 1.0.

PR #1755, opened 2022-10-10
fix: book cover access without authentication in srv/content

panaC:
When browsing an opds feed, an image cannot be protected. I created a dedicated route for the cover with auth_required=False

Useful for thorium-reader on browsing the calibre opds feed

### DIFF
--- a/src/calibre/srv/content.py
+++ b/src/calibre/srv/content.py
@@ -288,6 +288,21 @@ def icon(ctx, rd, which):
         ans.seek(0)
         return ans
 
+@endpoint('/cover/{book_id}/{library_id=None}', auth_required=False, android_workaround=True)
+def getCover(ctx, rd, book_id, library_id):
+    book_id, rest = book_id.partition('_')[::2]
+    try:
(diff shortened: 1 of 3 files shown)


kovidgoyal:
That would leak information about what books are in the library to
un-authenticated users. I dont follow why the book covers cannot be
protected by authentication, if the feed itself is protected.

panaC:
How can I display the image in the webview if this image is protected by digest authentication ? 

An API authenticate the xml/json data but not its medias. If you want to protect media content it must be signed.

kovidgoyal:
Pass the authentication data to the webview. And the problem is not
protecting the integrity of the data, the problem is not exposing what
books are in your library over the network.

panaC:
Please tell me how to pass the authentication data to the webview with the webapi ? Do you have the reference in MDN ?

kovidgoyal:
I have no idea what webview you are referring to. The content server is
absolutely not going to leak book covers to unauthenticated users.

PR #1756, opened 2022-10-10
Python templates, as discussed at length in …

cbhaley:
…https://www.mobileread.com/forums/showthread.php?t=349525.

I think this is stable, usable, and very close to finished, but I will probably need to react fo feedback, especially yours. :)


cbhaley:
I pushed a fix for the failing test

cbhaley:
Pushed some changes from un_pogaz.

cbhaley:
Added a few tests for python templates

cbhaley:
un_pogaz pushed a few more changes to the syntax highlighter

cbhaley:
Added another test -- calling a stored python template from a GPM template

kovidgoyal:
I havemerged, looks OK to me. One suggestion I have (which may already
have been discussed, I have not read the full thread on MR) is make the
python function signature

def evaluate(book, ctx):
   ...

Where ctx is an object that carries the db, globals etc as attributes.
Its simpler for users and more easily extensible in the future than
using **kwargs.

PR #1765, opened 2022-10-16
Python template introspection

un-pogaz:
A introspection system that allow to use, in Python template mode, the fonctions currently loaded in the TemplateFormatter.
The functions can be retrieve by they name and they name plus a '_' at the end in case of conflict with Python language keywords.

### DIFF
--- a/manual/template_lang.rst
+++ b/manual/template_lang.rst
@@ -670,7 +670,7 @@ You can add the above text to your template using the context menu, usually acce
 
 The context object supports ``str(context)`` that returns a string of the context's contents, and ``context.attributes`` that returns a list of the attribute names in the context.
 
-The ``context.funcs`` attribute allows to use the Builtin and User functions, and also the Stored GPM/Python templates so that you can exectute them directly in your code. The functions can be retrieve by they name and they name plus a '_' at the end in case of conflict with Python language keywords. Note that all functions will return a string value, only the special functions ``arguments()``, ``globals()`` and ``set_globals()`` will return a dict based on the passed keywords arguments.
+The ``context.funcs`` attribute allows to use the Builtin and User functions, and also the Stored GPM/Python templates so that you can exectute them directly in your code. The functions can be retrieve by they name and they name plus a '_' at the end in case of conflict with Python language keywords.
 
 Here is an example of a PTM template that produces a list of all the authors for a series. The list is stored in a `Column built from other columns, behaves like tags`. It shows in :guilabel:`Book details` and has the :guilabel:`on separate lines` checked (in :guilabel:`Preferences->Look & feel->Book details`). That option requires the list to be comma-separated. To satisfy that requirement the template converts commas in author names to semicolons then builds a comma-separated list of authors. The authors are then sorted, which is why the template uses author_sort.
 
--- a/manual/template_lang.rst
+++ b/manual/template_lang.rst
@@ -661,6 +661,7 @@ A PTM template begins with:
      # db: a calibre legacy database object
      # globals: the template global variable dictionary
      # arguments: is a list of arguments if the template is called by a GPM template, otherwise None
+     # funcs: allows to use the Builtin/User functions and Stored GPM/Python templates
 
      # your Python code goes here
      return 'a string'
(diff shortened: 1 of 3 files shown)


kovidgoyal:
@cbhaley are you ok with this PR? If so I will proceed to review it.

cbhaley:
I was ok with the approach and the implementation two days ago. I haven't looked at what un_pogaz has done since then, which seems to be quite a lot of refinement. I am certainly willing to look through it before you do the review, but I doubt I will find much. Let me know.

I do need to read through comments and the documentation to correct typos and some grammar faults, but that can be done after you accept the PR. NB: I wish my French is as good as un_pogaz's English.

kovidgoyal:
On Sun, Oct 16, 2022 at 09:45:27AM -0700, Charles Haley wrote:
> I was ok with the approach and the implementation two days ago. I haven't looked at what un_pogaz has done since then, which seems to be quite a lot of refinement. I am certainly willing to look through it before you do the review, but I doubt I will find much. Let me know.

Please do. I am sufficiently disconnect from template things that I
doubt my review will be very meaningful.

> 
> I do need to read through comments and the documentation to correct typos and some grammar faults, but that can be done after you accept the PR. NB: I wish my French is as good as un_pogaz's English.

Yes, that can be done after merge.

cbhaley:
Comments on formatter.py
======================
Do the comments on lines 871 - 873 still reflect what the FormatterFuncsCaller class does?

In particular: "If the name contain a illegal character for a attribut (like .:-), use getattr()"
Can a name contain an illegal character? GPM uses "normal" identifier syntax. 
======================
Small thing: line 894 should be "elif name in formatter.funcs:"
======================
I don't think the expression on line 902 is right. It is checking that the key is not None. Shouldn't it be checking the value? Or perhaps both? Perhaps: 
            kargs = {n(k):n(v) for k,v in kargs.items()}
======================
Line 909 ...: I don't see the point of imitating the arguments() built-in function. In python you can't call it with positional variable names, as in GPM's "arguments(a, b)", because python thinks they are values, not names. You can use arguments(a='') or arguments('a'), but all that does is fill in a dict with the values in arguments. It took me many minutes and coding some exampled to work out what calling arguments() really did, and I came to the conclusion that it adds much confusion but little value. I suggest that instead of trying to mimic its use in a GPM template, you raise an exception with a message telling the user what to do, for example "Get the arguments from context.arguments instead of calling arguments()"

I would do the same thing for globals(): raise an exception telling the user "Get the globals  from context.globals instead of calling globals()".

And also for set_globals(): raise an exception telling the user "Set globals using context.globals['name'] = val instead of calling set_globals".

Line 933-940: you need "return rslt" after line 937, or wrap lines 942-947 in an else block.

Putting all this together, perhaps this:
                    if func_name == 'arguments':
                        # if the message is translatable then the code parts must be arguments to the string
                        raise_error('Get the arguments from context.arguments instead of calling arguments()')
                    elif func_name == 'globals':
                        raise_error('Get the globals from context.globals instead of calling globals()')
                    elif func_name == 'set_globals':
                        raise_error("Set globals using context.globals['name'] = val instead of calling set_globals")
                    elif func_name == 'character':
                        if _Parser.inlined_function_nodes['character'][0](args):
                            rslt = _Interpreter.characters.get(args[0], None)
                            if rslt is None:
                                raise_error(_("Invalid character name '{0}'").format(args[0]))
                        else:
                            raise_error(_('Incorrect number of arguments'))
                    else:
                        if kargs:
                            raise_error(_('Cannot support keyword arguments'))

                        # builtin/user function and Stored GPM/Python template
                        func = formatter.funcs[func_name]
                        if func.object_type == StoredObjectType.PythonFunction:
                            rslt = func.evaluate(formatter, formatter.kwargs, formatter.book, formatter.locals, *args)
                        else:
                            rslt = formatter._eval_sfm_call(func_name, args, formatter.global_vars)
======================

The rest looks good, putting aside some line spacing and expression formatting I suspect @Kovid will change.

cbhaley:
All the edits to the above comment were to fix formatting. There wasn't any content change.

un-pogaz:
> Can a name contain an illegal character? GPM uses "normal" identifier syntax.

Yes, you can save a template or a function with a illegal character. Okay it's realy borderline case, and normally you could not use these functions in the GPM, but they exist.
image

> I don't see the point of imitating the arguments() ... globals ,set_globals().

Yep. Removed. I had also thought about raising an error, but finally, I wanted to push a full implementation.

This means that we don't need kargs anymore, because it was only useful for them. I leave it in the definition of call() because it will allow us to return a cleaner error message than the Python one.

(Holly, without this, it is a much cleaner code now. Too much is the enemy of good.)

PR #1769, opened 2022-10-18
Several things related to saved searches:

cbhaley:
- Add some "manage" actions to the saved searches menu
- Make the saved search button on the search toolbar and the saved searches action use the same menu
- Change some of the icons to use the existing dedicated saved search icons.
- Display the menu under a toolbar action or a menu action if either exists. **NB: I don't know if displaying under a menu works on a Mac or Linux**
- Make manage categories and saved searches use the same code to place the menu on the screen.

### DIFF
--- a/src/calibre/gui2/actions/manage_categories.py
+++ b/src/calibre/gui2/actions/manage_categories.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 # License: GPLv3 Copyright: 2022, Charles Haley
 #
-from qt.core import QPoint
 
 from calibre.gui2.actions import InterfaceAction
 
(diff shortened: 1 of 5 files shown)


kovidgoyal:
I doubt the menu actions will work on macOS or Linux when using a global
menu bar, as Qt has no access to the geometry for the global menu bar. I
can test it later, or just wiat for someone to complain :)

cbhaley:
I think wait. :)

The fallback is to place the menu on the upper left corner of the book list. People might not realize that something else might be able to happen.

cbhaley:
Question: why do you change "except:" to "except Exception"?

cbhaley:
Problem: changing the keyboard shortcut in Saved searches to None removes the action from the shortcut list. What is the best fix? Using '' works, but perhaps there are reasons not to do that?

I can make a PR with None ==> '' if that is the right approach.

kovidgoyal:
On Tue, Oct 18, 2022 at 05:26:26AM -0700, Charles Haley wrote:
> Problem: changing the keyboard shortcut in Saved searches to None removes the action from the shortcut list. What is the best fix? Using '' works, but perhaps there are reasons not to do that?

Using '' means multiple actions with the same '' will be reported as
conflicts. The best fix would be to use an invalid shortcut something
like F99 and F88, I think that should both allow it to show up in the
shortcuts list and not conflict.

kovidgoyal:
On Tue, Oct 18, 2022 at 05:13:55AM -0700, Charles Haley wrote:
> Question: why do you change "except:" to "except Exception"?

except catches everything including things like KeyboardInterrupt and
SystemExit that should not be caught.

PR #1771, opened 2022-10-19
Make tools on the search bar customizable.

cbhaley:
Given that the saved searches action and the saved search button on the search bar now do the same thing, it seems reasonable to allow the user to choose whether the saved search button shows. I went from there to "Why not allow adding other buttons to the bar?" There is a lot of available real estate on that bar. This commit does that using a new toolbar in preferences / toolbars.

I didn't use a QToolBar, instead constructing a 'classic' toolbar in an QHBoxLayout. That gave me more control over the look and placement. And also, I had no end of trouble trying to work through how to use a QToolBar that isn't attached to a QMainWIndow.

In the process I removed the legacy Saved Search combo box, its associated buttons, and the tweak. I am sure there are some people still using the years-old interface and that they will complain. I can accept that, given that the button is better in almost every way.

### DIFF
--- a/resources/default_tweaks.py
+++ b/resources/default_tweaks.py
@@ -522,12 +522,6 @@ content_server_thumbnail_compression_quality = 75
 #    cover_drop_exclude = {'tiff', 'webp'}
 cover_drop_exclude = ()
 
-#: Show the Saved searches box in the Search bar
-# In newer versions of calibre, only a single button that allows you to add a
-# new Saved search is shown in the Search bar. If you would like to have the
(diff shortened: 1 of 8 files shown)


cbhaley:
Added separators.

kovidgoyal:
merged, one comment:

The buttons on the search bar dont behave like they do in the toolbars.
Clicking on a button with a menu pops up the menu but does not perform
the action. And right clicking on them has no effect. This will be
confusing and a source of bug report noise. So maybe if the action has a
menu, have it open on right click only, depending on popup_type in the
action definition.

PR #1777, opened 2022-10-25
Related to Enhancement Request #1994065: Details pane: Splitter betwe…

cbhaley:
…en cover and details

This is a rather invoved PR. I first looked at putting a splitter in book details, and decided to in the process fix something that has bothered me for years. On Windows in light mode I can't see the splitter handles. In dark mode I can, but I run in light mode and don't want to change. This PR adds a custom QSplitterHandle class that draws the handle depending on the palatte.

The PR includes changes for the splitters in classes I know how to test: the main UI, Pinned columns, QuickView, Edit metadata, tweak books, single download metadata, Check library, Book info, and (of course) Book details.

I am sure that changing the other splitters is a straightforward as the ones I changed -- changing the import and changing the base class of the splitter. However, I don't want to change them unless I can run them to be confident there are no glitches.

### DIFF
--- a/src/calibre/gui2/book_details.py
+++ b/src/calibre/gui2/book_details.py
@@ -8,9 +8,9 @@ from collections import namedtuple
 from functools import partial
 from qt.core import (
     QAction, QApplication, QClipboard, QColor, QDialog, QEasingCurve, QIcon,
-    QKeySequence, QLayout, QMenu, QMimeData, QPainter, QPen, QPixmap,
+    QKeySequence, QMenu, QMimeData, QPainter, QPen, QPixmap,
     QPropertyAnimation, QRect, QSize, QSizePolicy, Qt, QUrl, QWidget, pyqtProperty,
-    pyqtSignal
+    QTimer, pyqtSignal
 )
 
 from calibre import fit_image, sanitize_file_name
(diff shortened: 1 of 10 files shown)


cbhaley:
With your changes the cover doesn't resize when moving the book details splitter. Is this what you wanted? If not, do you want me to fix it? Or will you do it?

kovidgoyal:
On Wed, Oct 26, 2022 at 07:56:35AM -0700, Charles Haley wrote:
> With your changes the cover doesn't resize when moving the book details splitter. Is this what you wanted? If not, do you want me to fix it? Or will you do it?

Yes, that was intended. Since we now set the size manually, its pretty
strange to have it change automatically.

cbhaley:
OK, but just to be sure, try this (I am in Narrow Mode):
- Using the splitter in book details, set things so the cover fills its area.
- Move the splitter between book details and the books list to make book details smaller. The cover doesn't change size. Part if it is now occluded. 
- Move the splitter inside book details. The cover jumps to fit in the new size and then changes size to fit.

Thus after resizing book details the cover doesn't "properly" fill its area. I must move the slider between the cover and the text to fix it.

cbhaley:
Actually, it might not be your changes. I see I didn't handle "resizeEvent" in book details. I will submit that change.

PR #1778, opened 2022-10-26
As I discussed in the previous PR, this commit makes the cover adapt …

cbhaley:
…to the new area when the book details pane changes size. I used the same technique you used in .widgets.

### DIFF
--- a/src/calibre/gui2/book_details.py
+++ b/src/calibre/gui2/book_details.py
@@ -10,7 +10,7 @@ from qt.core import (
     QAction, QApplication, QClipboard, QColor, QDialog, QEasingCurve, QIcon,
     QKeySequence, QMenu, QMimeData, QPainter, QPen, QPixmap,
     QPropertyAnimation, QRect, QSize, QSizePolicy, Qt, QUrl, QWidget, pyqtProperty,
-    pyqtSignal
+    QTimer, pyqtSignal
 )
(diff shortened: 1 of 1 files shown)


cbhaley:
Your change to use a style doesn't work on Windows 10. Do I need a new binary?

kovidgoyal:
yes, this needs a new binary as the custom drawing now happens in C++ code.

cbhaley:
OK, I will wait for a new beta or the next release, whichever comes first.

kovidgoyal:
Yes, one can theoretically set a proxystyle via Python, but performance
is really bad and you dont get access to various low level Qt things
that are needed sometimes for custom rendering of components. So it's
implemented in C++ in calibre.

kovidgoyal:
On Wed, Oct 26, 2022 at 08:50:46AM -0700, Charles Haley wrote:
> Your change to use a style doesn't work on Windows 10. Do I need a new binary?

I uploaded a new beta build, so you can try it out if you like.

kovidgoyal:
Actually, no the build process failed, so I will update the builds later today.

kovidgoyal:
Finally got the build working again so new beta is uploaded. Incidentally do you have a way to test the smart_device driver still? There was a report that the beta is crashing with it on MR

PR #1784, opened 2022-11-05
Fix for tag browser "AttributeError: 'NoneType' object has no …

cbhaley:
…attribute 'name'"

I found this by accident:

calibre 6.8*  embedded-python: True
Windows-10-10.0.19044-SP0 Windows ('64bit', 'WindowsPE') ('Windows', '10', '10.0.19044')
Python 3.10.1
[...]
File "C:\CBH_Data\calibre.git\calibre_dev\src\calibre\gui2\tag_browser\view.py", line 981, in show_context_menu
AttributeError: 'NoneType' object has no attribute 'name'


cbhaley:
The exception happens when right-clicking on the Format root node, not on a child node.

PR #1792, opened 2022-11-16
perf: CssChecker Pool performance

polyproline:
CssChecker Pool Use QEventLoop instead processEvents. Using Qt signal better than busy check
lru_cache is better than getattr to lazy initialization share object


kovidgoyal:
Have you even run your code? It prevents the editor from even starting
up. And there is absolutely no significant performance difference between
processEvents() and running eventloop exec(). eventloop exec() just calls
processEvents itself.

As for using lru_cache, yes it's clearer but the code in question was
written before lru_cache existed and changing it to lru_cache now is not
worth the code churn.

polyproline:
I see, I was wong.

PR #1795, opened 2022-11-19
Bug(s) #1997137: Tag browser error.

cbhaley:
There are two independent problems here.

1) The exception is caused by right-clicking in the white space of the tag browser. I can't make it happen right-clicking anywhere else. It has nothing to do with grouped searches.

2) The "creation of grouped search terms" problem has been there since forever. The search preferences didn't tell the tag browser to completely rebuild the tree. Restarting calibre fixed it.

### DIFF
--- a/src/calibre/gui2/preferences/search.py
+++ b/src/calibre/gui2/preferences/search.py
@@ -229,6 +229,7 @@ class ConfigWidget(ConfigWidgetBase, Ui_Form):
         self.gst_value.blockSignals(False)
 
     def commit(self):
+        restart = ConfigWidgetBase.commit(self)
         if self.opt_case_sensitive.isChecked() and self.opt_use_primary_find_in_search.isChecked():
             error_dialog(self, _('Incompatible options'), _(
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Whya re we moving the commit to before the check for incompatible options? Isnt the idea that the incompatible options should prevent a commit?

cbhaley:
The commit must be before self.db.field_metadata.add_grouped_search_terms(self.gst) in commit() or the the reset_tag_browser() in refresh_gui() doesn't work. The tag browser doesn't see the new values. I'm not sure why. There must be a race condition somewhere.

You are right, though, it should be after the error check. Change submitted.

cbhaley:
I think I traced down why the commit must be where it is. A set_pref('grouped_search_terms') will call cache._initialize_dynamic_categories(), which removes the gst "user categories" from field_metadata(), replacing them with whatever is in preferences. The pref "grouped_search_make_user_categories" must be set with any changes before that happens, which is why the commit() must be where it is.

PR #1799, opened 2022-12-03
Preserver ASIN on mobi export

ricrogz:
I recently noticed that exporting to mobi formats always overwrites the ASIN with the book's uuid, so that the ASIN is lost in the export.

This patch should preserver the ASIN (if present), still using the uuid as 'SOURCE'


kovidgoyal:
This is by design. Amazon uses the identifier as means to sync
annotations/last read/etc. Preserving the identifier in a file that has
been potentially altered is not a good idea.

PR #1804, opened 2022-12-14
Duplified with ambito.recipe

isantolin:
ambito.recipe uses RSS feeds and its working


kovidgoyal:
yes but this recipe uses parse_index() thus potentially returns
different articles. If it isnt working, it should be fixed, not deleted.

williamj20 (approved):

eli-schwartz:
@kovidgoyal It appears that #2229 was produced by regurgitating this PR with no changes (I got suspicious when I saw the commit from the other PR was from 2022 by an entirely different author).

I wonder if your original review comment still applies. :D

kovidgoyal:
Yeah probably does but since no one has stepped up to fix the recipe, might as well delete it at this point.

PR #1808, opened 2022-12-17
Clean up the new composite splitter API to ignore case and ignore …

cbhaley:
…dups. Make Quickview use the new API.


kovidgoyal:
Not sure I follow. Why are you constructing a dictionary rather than a set and then converting to a list? If you want q list of unique values use the uniq function from db.write.py thought this will return a tuple which is arguably better than a list here. Alos you left in a couple fo debug prints.

cbhaley:
New changes committed. The commit messages explain my reasoning.

PR #1814, opened 2023-01-08
Update choose_library.ui

chocolatechipcats:
Added a bit of text that's meant to link to the network drive FAQ entry. I couldn't properly implement the link itself (a href failed, and I couldn't find an example of a working hyperlink in a dialog).

### DIFF
--- a/src/calibre/gui2/dialogs/choose_library.ui
+++ b/src/calibre/gui2/dialogs/choose_library.ui
@@ -38,6 +38,16 @@
      </property>
     </widget>
    </item>
+      <item row="3" column="0" colspan="4">
+    <widget class="QLabel" name="network_warning">
+     <property name="text">
+      <string>Putting your library on a networked drive?</string>
+     </property>
+     <property name="wordWrap">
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Here are some examples.

https://github.com/search?q=repo%3Akovidgoyal%2Fcalibre%20path%3A.ui%20https&type=code

The basic rule is that you can lightly style the text with HTML, but you have to escape it so that the Qt UI compiler treats it as part of the rich text contents of the QLabel.

eli-schwartz:
https://github.com/kovidgoyal/calibre/blob/7a829f484afda16992326f3bb7080a8cd6b1d0a9/src/calibre/gui2/convert/toc.ui#L112-L114
https://github.com/kovidgoyal/calibre/blob/7a829f484afda16992326f3bb7080a8cd6b1d0a9/src/calibre/gui2/convert/toc.py#L31-L32

For hyperlinks to the user manual specifically, there's a slightly different pattern which would be best. :)

PR #1825, opened 2023-02-16
Extend FileTypePlugin with postconvert() and postdelete() methods

siebert:


kovidgoyal:
1) The help text for postconvert should say it is run after a conversion only not after a books is added.

2) There is another place where conversions happen (srv/convert.py)

3) The help text for postdelete should note that it is not run when a book record is deleted, only when one or more formats from the book are deleted.

siebert:
Regarding 2), how do I trigger such a conversion, so I can test that the code in srv/convert.py is executed?

kovidgoyal:
Use the calibre content sever and convert a book in it. Click on a book to view its details page then click on the convery button in the top toolbar.

PR #1826, opened 2023-02-19
add zstd as a dependency

xcffl:
Since Qt has zstd enabled by default, most distributions build Qt with zstd enabled, some dynamically linked libraries with zstd enabled are called by the statically linked libraries bundled with Calibre, it's error-prone to build Calibre without zstd.

See also: https://github.com/flathub/com.calibre_ebook.calibre/issues/34#issuecomment-1435642018

---

You may squash this PR when you merge it if you want to

### DIFF
--- a/bypy/sources.json
+++ b/bypy/sources.json
@@ -996,5 +996,15 @@
             "hash": "sha256:1324b66b356051de662ba87d84f73ada062acd42b047ed1246e60a449f833e10",
             "urls": ["pypi"]
         }
+    },
+
+    {
+        "name": "zstd",
+        "os": "linux",
+        "unix": {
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This needs to be added before pyzstd and support for building it has to be added to https://github.com/kovidgoyal/bypy  as well as a check to ensure that pyzstd links against it.

kovidgoyal:
No it supports linking see https://pyzstd.readthedocs.io/en/latest/#build-pyzstd

kovidgoyal:
Oh and if you want zstd used by Qt when building, then it needs to be moved before qt as well. Best to just put it after brotli assuming it doesnt have any other dependencies itself.

PR #1840, opened 2023-03-15
The documentation is out of date in several locations on how to change the library location

iconoclasthero:
I've changed "Library" to the currently-displayed "ebooks" (see https://imgur.com/RQFOACt) and put in a new .png of the same size that says "ebooks" and **not** "Library."  

This configuration is very difficult to find and should be placed in the properties menu for "Saving books to disk" because that is what I am doing, saving the newly introduced books into Calibre and saving them to the location I specify.  If that isn't intuitive enough, then perhaps "Add books" makes more sense.  Either way, where it is is nearly impossible to find especially with out-dated documentation.


kovidgoyal:
I dont follow. The default library name is "Calibre Library" and it is
shown in the toolbar as such. You at some point in the past
renamed/moved it "ebooks". And saving to disk has nothing to do with
calibre library locations. It is how you *export* books from your calibre
library.

PR #1844, opened 2023-03-24
Added Kobo Libra 2 to output profiles

Smileyrasmus:
### Things I don't know
- if `fbase` and `fsizes` makes sense for the kobo libra 2 specifically. I just copied from the KoboReaderOutput class, thinking it would work nicely.

If someone can explain what fbase and fsizes does in practice, then I can test further on my own device to ensure correct settings.


kovidgoyal:
There's not much point adding profiles for modern devices, since they
are quite capable of scaling images well themselves. Just use one of the
tablet or generic e-ink profiles. Indeed, its on my todo list to make
calibre default to not rescaling images during conversion.

PR #1845, opened 2023-03-25
This is a large commit.

cbhaley:
DB changes:
1) Add link columns to "normalized" tables. All such in-memory tables have an attribute self.link_map.
2) Add API to set and get links for fields
3) get_metadata now includes an attribute giving the link maps for the book. Book link maps are cached.
4) ProxyMetadata can return link maps
5) Added a test for the API.

URL Scheme:
1) Added a "book-details" URL that asks calibre to open a book info window on a book in some library

Book Details:
1) You can now have multiple book info windows.
2) If an item as an associated link then that link is made available using "(item link)" link text.
3) Book info windows on books in other libraries have no links

UI:
1) the Manage Category editor presents a fourth column for links.

OPF:
1) The OPF used for backing up (metadata.opf) contains the link map for the book. Currently this isn't used, but it should be used in recover_database. I didn't do anything with OPF3.

### DIFF
--- a/manual/url_scheme.rst
+++ b/manual/url_scheme.rst
@@ -116,6 +116,16 @@ If you perform a search in calibre and want to generate a link for it you can
 do so by right clicking the search bar and choosing :guilabel:`Copy search as
 URL`.
 
+Open a book details window on a book in some library
+------------------------------------------------------
+
+The URL syntax is::
+
+    calibre://book-details/Library_Name/book_id
+
+This opens a book details window on the specified book from the specified library without changing the
+current library or the selected book.
+
 
 .. _hex_encoding:
 
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -1121,6 +1121,7 @@ class DB:
                 CREATE TABLE %s(
                     id    INTEGER PRIMARY KEY AUTOINCREMENT,
                     value %s NOT NULL %s,
+                    link TEXT NOT NULL DEFAULT "",
                     UNIQUE(value));
                 '''%(table, dt, collate),
 
--- a/src/calibre/db/cache.py
+++ b/src/calibre/db/cache.py
@@ -2405,6 +2405,9 @@ class Cache:
         table = self.fields[field].table
         if not hasattr(table, 'link_map'):
             raise ValueError(f"Lookup name {field} doesn't have a link map")
+        # Clear the links for book cache as we don't know what will be affected
+        self.link_maps_cache = {}
+
         fids = {k: self.get_item_id(field, k) for k in value_to_link_map.keys()}
         id_to_link_map = {fid:value_to_link_map[k] for k, fid in fids.items() if fid is not None}
         result_map = table.set_links(id_to_link_map, self.backend)
--- a/src/calibre/db/cache.py
+++ b/src/calibre/db/cache.py
@@ -153,6 +153,7 @@ class Cache:
         self.format_metadata_cache = defaultdict(dict)
         self.formatter_template_cache = {}
         self.dirtied_cache = {}
+        self.link_maps_cache = {}
         self.vls_for_books_cache = None
         self.vls_for_books_lib_in_process = None
         self.vls_cache_lock = Lock()
@@ -290,6 +291,7 @@ class Cache:
             self.format_metadata_cache.clear()
         if search_cache:
             self._clear_search_caches(book_ids)
+        self.link_maps_cache = {}
 
     @write_api
     def reload_from_db(self, clear_caches=True):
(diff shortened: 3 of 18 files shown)


cbhaley:
By the way, the tests sometimes get confused by the frequent schema upgrade operations

cbhaley:
Another push to fix not properly deleting links in the UI

kovidgoyal:
On Sat, Mar 25, 2023 at 09:49:53AM -0700, Charles Haley wrote:
> By the way, the tests sometimes get confused by the frequent schema upgrade operations

I dont follow, schema upgrades are deterministic, which test is failing
and with what error?

cbhaley:
I said it badly. I meant "some tests get confused ..." One of them is test_library_wide_properties (calibre.db.tests.legacy.LegacyTest). It doesn't like that the version number and timestamp changed.

cbhaley:
Pushed improved link tests and a fix for a problem the tests found

kovidgoyal:
1) How is links_map_cache supposed to work? As far as I can see it is only
cleared in clear_caches() and set_link(). But deleting or renaming an
item or changing the items for a book will also change the link map for
a book. Am I missing something?

2) Also to fix the failing test, all you need to do is run schema_upgrade
on the metadata.db file used for the tests. Something like:


from calibre.library import db
db = db("src/calibre/db/tests")


should do the trick.

3) In lazy.py probably move the check for link_maps to after SIMPLE_GET
so getting basic things are not affected performance wise. Or even move
it to the end of the method, since getting it is rare.

4) In schema_upgrades.py I suggest collecting all the ALTER statements
into on large string separated by ; and running a single execute. Faster
and more robust since we wont end up with some tables updated and some
not.

I'll review the higher level stuff later.

cbhaley:
I pushed the changes we discussed, as well as an improved test.

kovidgoyal:
merged, making a rather large number of changes to the code, hopefully I
didn't break anything. Please review.

cbhaley:
You took out the ability to open a non-slaved book details window on a book. Now it simply opens another slaved window. You also took out remembering the position of these "extra" windows. What that intentional? There is no point in having multiple slaved windows all showing the same thing.

kovidgoyal:
On Fri, Mar 31, 2023 at 04:07:06AM -0700, Charles Haley wrote:
> You took out the ability to open a non-slaved book details window on a book. Now it simply opens another slaved window. You also took out remembering the position of these "extra" windows. What that intentional? There is no point in having multiple slaved windows all showing the same thing.


Not sure I follow, opening calibre://book-details/whatever
opens a book detail window that has no links and no next/previous
buttons and does not change what is displayed when the currnt book in
the library changes. And these types of windows remeber their geometry
separately from regular book details windows.

cbhaley:
See the attached image.

My code allowed you to open any number of book detailed windows. The first such window is special, slaved to the library view. The other windows are locked on the book being displayed when the window was opened. The calibre://show-details windows are a special case where the title bar changes. It remembered the geometry of each window, not just the slave window and the calibre://show-details

With the code as you changed it, one cannot open a book details window locked on a book. Instead we get another slaved window showing exactly what the existing slaved window is showing. There is no utility in that. Why have 3 windows showing the same thing? In addition, if I open a second calibre://show-details link it places it on top of the existing one. Moving it changes the remembered geometry for all of them.

Clipboard01

cbhaley:
FWIW: the above is why display_number existed -- so multiple geometries would be stored and the behavior would be predictable.

cbhaley:
However, perhaps ProxyMetadata never sees changed data? In other words, it is "illegal" to use an existing mi instance after the database has changed?

kovidgoyal:
On Fri, Mar 31, 2023 at 06:18:52AM -0700, Charles Haley wrote:
> However, perhaps ProxyMetadata never sees changed data? In other words, it is "illegal" to use an existing mi instance after the database has changed?

Yes, ProxyMetadata will not show you up to date data, this is true for
(almost) all attributes not just link_maps

kovidgoyal:
On Fri, Mar 31, 2023 at 04:55:36AM -0700, Charles Haley wrote:
> See the attached image.
> 
> My code allowed you to open any number of book detailed windows. The first such window is special, slaved to the library view. The other windows are locked on the book being displayed when the window was opened. The calibre://show-details windows are a special case where the title bar changes. It remembered the geometry of each window, not just the slave window and the calibre://show-details

yes my code does the same, except that instead of saving geometries
keyed by number it saves a separate geometry for the window
corresponding to the current library and windows corresponding to other
libraries. I didnt really see what sense it made to have remembered
geometries based on dialog number. It means the size of things depends
on the order you open them, which is not very intuitive. But I guess it
allows you to easily tile them, something I didnt consider as I use a
tiling window manager myself.

> 
> With the code as you changed it, one cannot open a book details window locked on a book. Instead we get another slaved window showing exactly what the existing slaved window is showing. There is no utility in that. Why have 3 windows showing the same thing? In addition, if I open a second calibre://show-details link it places it on top of the existing one. Moving it changes the remembered geometry for all of them.

Yes, I forgot to handle the case of multiple windows showing the current
library book being opened. I was supposed to have just raised the
existing window in that case.

But, on further consideration, I guess having it open another copy
locked to the current book is more useful. So I am OK to restore your
code, feel free to send a PR with it.

PR #1851, opened 2023-04-01
Fix links for series and custom series columns being confounded

cbhaley:
Is reported as Bug 2014971


cbhaley:
Added a second commit for a "Copy all links" feature.

PR #1852, opened 2023-04-02
Add Category edit buttons to columns that have category editors.

cbhaley:
Authors, Series, Tags, Publisher, custom columns


cbhaley:
The buttons are in Edit Metadata (Single)

PR #1854, opened 2023-04-05
Enhancement #2015317: rename user category tag browser icons when a …

cbhaley:
…category is renamed


kovidgoyal:
Can other things refer to the icon? IOW, is it safe to rename the icon?

cbhaley:
Nothing else should refer to it. The icon is named with the lookup key of the user category in question. There cannot be more than one such lookup key. Of course I can't speak for what plugins might do.

I could leave it but that would accumulate garbage for no reason I can see.

PR #1857, opened 2023-04-07
Update README for download section

Goooler:
https://github.com/kovidgoyal/calibre/releases
https://formulae.brew.sh/cask/calibre


kovidgoyal:
The only supported, permanent location to download calibre from is https://calibre-ebook.com/download

PR #1858, opened 2023-04-07
Undo commit e11b8c8. book-details links should appear in the special …

cbhaley:
…window (no search links, etc) even if in the current library. The LibraryBroker deals with not opening the database twice.

This might be worth a point release. See https://www.mobileread.com/forums/showthread.php?p=4312453#post4312453 and later.


kovidgoyal:
As for a point release, I dont mind but lets wait a little longer in
case any more issues are reported.

PR #1877, opened 2023-04-22
Bug #2017316: Error when renaming tag in Tags Manager.

cbhaley:
Note: I rolled back commits d5dbb3c63a1fe05f85a31cf1e4cd021ec76e750a and 89391da7b7074b42aad7c440e97ac76cfd18175a.


kovidgoyal:
If I had to guess I would say ownership issues. Check the parents of the
table widget items. Nothing in Qt has changed in the last weeks.

PR #1879, opened 2023-04-23
API for the extra_files cache

cbhaley:
This PR is for the base API for the extra files cache, including clearing when files are added and with the directory is views using the calibre UI.

My assumption: you will review this code. I will build the formatter functions once the review is finished.

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -1887,11 +1887,12 @@ class DB:
 
     def iter_extra_files(self, book_id, book_path, formats_field, yield_paths=False, pattern=''):
         known_files = {COVER_FILE_NAME, METADATA_FILE_NAME}
-        for fmt in formats_field.for_book(book_id, default_value=()):
-            fname = formats_field.format_fname(book_id, fmt)
-            fpath = self.format_abspath(book_id, fmt, fname, book_path, do_file_rename=False)
-            if fpath:
-                known_files.add(os.path.basename(fpath))
+        if formats_field is not None:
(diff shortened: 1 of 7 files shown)


cbhaley:
Here is a sample python template I used to test it
```
python:
def evaluate(book, context):
	db = context.db.new_api
	ans = db.list_extra_files(book.id)
	# print(ans)
	return ', '.join([str(ans[k]) for k in ans.keys()])
```

kovidgoyal:
I was envisaging an in-memory cache. So it starts out always empty and is built up as used. If we are going to store it on disk might as well create a new table for it. But I dont really want to get in to having to track the extra files in the db either via preferences or a table. Extra files, unlike formats/covers dont really have much in the way of UI to manage them in calibre. Which means we expect users to manage them outside calibre. As such trying to track them within the db isnt really a good idea, IMO. The in-memory cache is a reasonable compromise. First query will still be slow but subsequent ones wont. And we dont have a on disk structure that will get out of sync and all the associated headache.

cbhaley:
OK, taking out storing the cache is easy enough.

I will leave the "Clear the cache" in library maintenance that someone could use if they made wholesale changes while calibre is running and doesn't want to restart for some reason.

I think that clearing the cache for individual books when opening the folder browser should stay.

cbhaley:
Requested changes pushed

kovidgoyal:
I have merged, unifying the two list_extra_files functions into one API.
It should still work fine for your use case. I also decided to remove
the action to clear the cache, see commit for details.

PR #1880, opened 2023-04-23
Two semi-dependent changes:

cbhaley:
1) Add a tweak to change the heading for standard columns
2) Add "id", "formats", and "path" as standard columns. They are by default hidden, like "languages"


kovidgoyal:
I added the Id column to a test library, now everytime calibre starts the book list is scrolled to the id column instead of its previous position.

cbhaley:
I can't make this happen. I tried using Show column and selecting id, creating an empty library and showing id, and showing id in an existing library.

What are you doing?

PR #1881, opened 2023-04-24
The formatter functions to support extra book files.

cbhaley:


cbhaley:
I pushed an unrelated commit: Prevent calling edit metadata from locked book details windows. EM edits 
  the currently selected book, which is almost certainly not the book in the locked window.

PR #1884, opened 2023-04-25
Enhancement #2017403: Add a preview tab to Markdown "long text" editor

un-pogaz:
https://bugs.launchpad.net/calibre/+bug/2017403

Add a editor widget for Markdown text. This editor has 2 tabs : one for the source editing (default) and a other to preview a static result rendering of the text.
This editor is used in "Edit Metadata of a book".


cbhaley:
It works. As far as I can see it doesn't break any of the rules for custom column widgets. 

One comment: the class contains the method smarten_punctuation() that as far as I can see cannot be called. It is in the context menu for the rich text editor, but as the markdown editor doesn't manage the context menu the option to call the method isn't there. I suggest that either it be removed or the markdown editor manage the context menu and add that option.

un-pogaz:
Okay, added.

PR #1895, opened 2023-04-29
Fix weird problem where a menu can be shown on the wrong screen if it…

cbhaley:
… is constructed before the first show event. On my test system it is 100% repeatable.

FWIW: I spent several hours trying to find a way to work around this issue. The trigger is clear -- populating the menu for the first time outside of the Qt "show()" process. This commit fixes it by not populating the 'real' menu until the show sequence has started. If something wants the menu before the show process has started then a fake action is returned with the menu.

### DIFF
--- a/src/calibre/gui2/actions/__init__.py
+++ b/src/calibre/gui2/actions/__init__.py
@@ -8,7 +8,7 @@ __docformat__ = 'restructuredtext en'
 from functools import partial
 from zipfile import ZipFile
 
-from qt.core import (QToolButton, QAction, QIcon, QObject, QMenu,
+from qt.core import (QToolButton, QAction, QIcon, QObject, QMenu, QPoint,
         QKeySequence)
 
 from calibre import prints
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Almost all menus are built outside of show() is that really the cause?
If so why would it affect only this menu?

cbhaley:
I wish I could tell you. However, it is 100% repeatable on my system that if:

- calibre is on screen 2
- I build the menu in initialization_complete() (called by ui.py)
- I click a tool button on the tool bar or the search bar
- The resulting menu overflows into 2 columns

then when the tool button is clicked the menu shows on screen 1 instead of screen 2.

The problem doesn't happen if I use an action on the menu bar. Also, once I click the action on the menu bar, the tool buttons start working.

If I don't build the menu in initialization_complete() then the menu is placed correctly on screen 2 in all cases. Unfortunately the plugins see the wrong data until the tool button is clicked because the menu hasn't been set up.

If I remove enough "sortable columns" so the menu doesn't overflow into 2 columns then the menu is always placed correctly on screen 2. 

My suspicion is that this problem will happen any time a menu is initialized by initialization_complete() and it overflows onto 2 columns.

My fix is to build a copy of the menu when 'something' (a plugin) asks for it before the tool button has been clicked.

kovidgoyal:
OK, but there might be code that expects qaction to be populated with
correct text/icon. So when constructing it copy those things from the
real qaction.

Also using a proxy qaction means that using the qaction at lines 68 and
126 will not work anymore until the action becomes the correct one.

Why not build the menu both in initialization_completed() and
then again in aboutToShow()?

cbhaley:
> Why not build the menu both in initialization_completed() and then again in aboutToShow()?

This doesn't fix it.

>Also using a proxy qaction means that using the qaction at lines 68 and
126 will not work anymore until the action becomes the correct one.

Line 68 isn't affected because initialization_complete() hasn't been called yet. However, you are right about line 126.

There is another approach that both adds functionality (a shortcut) and works around the problem. I can do what is done in actions.manage_categories.py and actions.saved_searches.py to add the possibility of a shortcut for Sort by. This opens a menu under the toolbar/menu widget, and in the process corrects the menu-on-wrong-screen problem. The workaround then becomes "if the problem occurs then press the shortcut." I would assign the default shortcut of Ctrl+F5 to go with the two existing shortcuts F5 (redo sort) and Shift+F5 (reverse sort).

The problem is discoverability. I can add the workaround to the tooltip, but that makes it rather front-and-center and probably confusing. Perhaps we do nothing about discoverability because the problem will be rare. All four triggering conditions must be true.

If you are OK with adding the shortcut then I would like to move the "controller" function into actions.__init__.py. There would be 3 actions that use it. I think the existence of the method should be more discoverable for programmers.

kovidgoyal:
On Sun, Apr 30, 2023 at 01:26:15AM -0700, Charles Haley wrote:
> > Why not build the menu both in initialization_completed() and then again in aboutToShow()?
> 
> This doesn't fix it.
> 
> >Also using a proxy qaction means that using the qaction at lines 68 and
> 126 will not work anymore until the action becomes the correct one.
> 
> Line 68 isn't affected because initialization_complete() hasn't been called yet. However, you are right about line 126.
> 
> There is another approach that both adds functionality (a shortcut) and works around the problem. I can do what is done in actions.manage_categories.py and actions.saved_searches.py to add the possibility of a shortcut for Sort by. This opens a menu under the toolbar/menu widget, and in the process corrects the menu-on-wrong-screen problem. The workaround then becomes "if the problem occurs then press the shortcut." I would assign the default shortcut of Ctrl+F5 to go with the two existing shortcuts F5 (redo sort) and Shift+F5 (reverse sort).
> 
> The problem is discoverability. I can add the workaround to the tooltip, but that makes it rather front-and-center and probably confusing. Perhaps we do nothing about discoverability because the problem will be rare. All four triggering conditions must be true.
> 
> If you are OK with adding the shortcut then I would like to move the "controller" function into actions.__init__.py. There would be 3 actions that use it. I think the existence of the method should be more discoverable for programmers.

OK. I am good with that.

cbhaley:
The commit I just pushed fixes it without the discoverability problem by never allowing Qt to show the menu. The button's action when clicked is to show the menu under the button. The tricky part was getting rid of the menu arrows on the tool buttons, as they serve no purpose.

PR #1899, opened 2023-05-01
Now three independent commits

cbhaley:
1) Make model.refresh_rows() also clear the extra files cache for those books. Rationale: if someone changes the contents of data/ outside of calibre then this provides a way to get calibre to recheck.
2) Add opening the data folder to the Open action. Allow it to have a shortcut. Create the data folder if it doesn't exist.
3) Bug #2018227: User Categories: "Hide empty categories" failing for sub-categories


cbhaley:
Add commit: Bug #2018227: User Categories: "Hide empty categories" failing for sub-categories

PR #1913, opened 2023-05-19
The manual is inaccessible because of the tracking

sebs:
I have my ad blocker disabled and still can not see the manual. 
lots of errors about google tag manager not working


kovidgoyal:
Then I suggest using a working browser. The manual works fine, and uses Google ads which are needed to pay for its hosting costs. If google ads offends you I suggest you either learn to block it correctly ordont use the manual.

sebs:
adblocker was off ;) I did check that.

sebs:
today everything works.

PR #1917, opened 2023-05-24
Minor string change for save_alternate_cover

chocolatechipcats:
There was an extraneous 'as' in the context menu option. I removed it.


chocolatechipcats:
The force-push was correcting a typo in the commit message

PR #1918, opened 2023-05-24
some icons

unkn0w7n:


kovidgoyal:
The bloomberg icons are .ico files they need to be converted to actual PNG first

PR #1924, opened 2023-05-28
User option for behavior when editing cell

un-pogaz:
Allow the user to choose which behavior to adopt between "Select all content" and "No select/Append" when editing "Texts" cell.


kovidgoyal:
That's a lot more work than that. You have to do it for all delegates throughout calibre. Your patch only affects text like delegates in the book list. There are many other places where delegates create editors.

PR #1926, opened 2023-06-03
removing defunct recipes.

unkn0w7n:
I am looking for recipes that haven't been modified for 7 years and deleting them. (You did PEP8 for all files 7 years ago)

The recipes I've removed below are the ones with feeds that no longer work.

But I want to remove all the recipes that are 7 years old without having to go through the code or checking if feeds work, they probably don't have _keep/remove_tags_ right, as those websites html architecture would've changed by now.

Is it okay with you if I remove most of them? I remember trying so many recipes and nothing working when I first came across fetch news, & a majority of recipes still don't work.


kovidgoyal:
I dont think blindly removing recipes is a good idea. 

1) Some website designs are quite static, especially niche/smaller
websites

1.5) Recipes based on RSS feeds that get the actual content from the
feed are likely to remain working for a long time

2) Even if keep_tags/remove_tags is wrong it is much easier to fix those
in an existing recipe than to start from scratch. Deleting the recipes
would mean that someone that is trying to get a recipe for a site will
have to start from scratch (the recipe would be in git history but
practically speaking no one would think to look there).

unkn0w7n:
Okay, that's also true. I did see 7 year old recipes with feeds that still worked.

Can we point people towards _mobileread recipe forum_ or _bugs.launchpad_ to report these recipes. It took a whole of 3 months for me to find mobileread & go on requesting fixes for some of these recipes. None of the Indian recipes worked right back then, except Hindu. 
Maybe at least give the launchpad link, whenever a recipe gives error. If they are desperate enough, they'll make an account & create a request ticket. If no one fixes it, we can then remove the recipe.

kovidgoyal:
On Sat, Jun 03, 2023 at 10:14:11PM -0700, unkn0w7n wrote:
> Okay, that's also true. I did see 7 year old recipes with feeds that still worked.
> 
> Can we point people towards _mobileread recipe forum_ or _bugs.launchpad_ to report these recipes. It took a whole of 3 months for me to find mobileread & go on requesting fixes for some of these recipes. None of the Indian recipes worked right back then, except Hindu. 
> Maybe at least give the launchpad link, whenever a recipe gives error. If they are desperate enough, they'll make an account & create a request ticket. If no one fixes it, we can then remove the recipe.

Yes, sure, adding a link to support forums in the recipe failure case is
a good idea.

PR #1945, opened 2023-06-23
Don't cmage fallback for formula

NogNoa:


kovidgoyal:
I know of no attribute known as altimg in the HTML spec. calibre is not
going to rewrite random custom attributes.

NogNoa:
Oh. I didn't realise I published this request separately mid-edit. Sorry about that.

I guess with the explanation it looked more reasonable. 

I didn't actually checked the details beforehand. It looks like MathML was a seperate language/XML application that is embeddable in HTML, but was also included in HTML5.

PR #1946, opened 2023-06-24
Don't complain over <math altimg= >

NogNoa:
Image files could be defined as fallback in the altimg atteribute of a <math> tag. The editor's error checker doesn't recognise this usage and report the file as unreferanced.


NogNoa:
I'm not sure how many eReaders support this, (mine doesn't). But calibre itself display the image.

PR #1953, opened 2023-07-02
hindutamil.recipe

unkn0w7n:
Does using encoded characters make it useful? Should I encode the tamil text in the recipe like this? _\u0B87\u0BA8\u0BCD\u0BA4\u0BC1_


kovidgoyal:
No, that was needed in ancient calibre versions

PR #1956, opened 2023-07-10
Cleanup of PR cleanup (Layout actions)

cbhaley:


cbhaley:
The button_names() method was there to help API users build combo boxes of pane names or use translated names in check boxes. There isn't another (easy) way to get the list of translated pane names other than reinventing what the method did.

kovidgoyal:
Ah OK, then maybe add a docstring explaining it so people know what its for.

PR #1966, opened 2023-07-14
Enhancement #2027727: Accented characters in tag browser find

cbhaley:


kovidgoyal:
It occurs to me that the option is "Unaccented characters match accented characters and punctuation is ignored". So it would probably confusing to have the tag browser follow the first part (accented) but not the second (punctuation).

cbhaley:
The problem for me is that it is much more work to add what is in effect all the search types. Currently the tag browser 'find' allows the '=' for full match. This is passed as a bool to various places. To allow ^ (character variant) (and probably '~') I would need either to change the bool to an enum or push parsing out to the use sites.  Neither will be fun.

I also don't think that ignoring punctuation is a good idea, especially in the face of identifiers and hierarchical items.

Your point is why I changed the tooltip, to say explicitly that character variant searches are used.

PR #1976, opened 2023-07-22
recipe updates

unkn0w7n:
def values(objects):
    body = '<p>'
    for content in objects['content']:
        if 'value' in content:
            body += content['value']
        elif 'content' in content:
            values(content)
    return body

In bloomberg, businessweek lines 241-260, I tried to replace with code above but it did not work the same way. So, I let it go on for 4 repetitions.


kovidgoyal:
For this kind of thing one should use a recursive function, like this: https://github.com/kovidgoyal/calibre/commit/3a573e5693ffdaff83e9e0edba664cb24faac1c1

unkn0w7n:
Thank you 😊

PR #1996, opened 2023-08-16
For some reason I don't understand, the array of mi values must be a …

cbhaley:
…list, not a tuple.

When a tuple, this line:
`w.setText(mi[r].title)`

produces the exception:
```
  File "C:\CBH_Data\calibre.git\calibre_dev\src\calibre\gui2\preferences\save_template.py", line 63, in do_open_editor
  File "C:\CBH_Data\calibre.git\calibre_dev\src\calibre\gui2\dialogs\template_dialog.py", line 433, in __init__
  File "C:\CBH_Data\calibre.git\calibre_dev\src\calibre\gui2\dialogs\template_dialog.py", line 578, in set_mi
AttributeError: 'tuple' object has no attribute 'title'
```

### DIFF
--- a/src/calibre/gui2/preferences/save_template.py
+++ b/src/calibre/gui2/preferences/save_template.py
@@ -53,7 +53,7 @@ class SaveTemplate(QWidget, Ui_Form):
         from calibre.gui2.ui import get_gui
         db = get_gui().current_db
         view = get_gui().library_view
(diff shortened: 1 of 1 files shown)


kovidgoyal:
THat will be because of line 522 in template_dialog.py

PR #1999, opened 2023-08-18
Fix book details rendering and customization

un-pogaz:
After view [this post](https://www.mobileread.com/forums/showthread.php?p=4348631) and discover the imposibilty to use the custom CSS, a investigate the rendering of the book details.

This pull is the result of my test, and fix all the issue I found.
Is now possible to propely use a custom CSS to personalize any elements of the book details HTML.

### DIFF
--- a/resources/templates/book_details.css
+++ b/resources/templates/book_details.css
@@ -8,15 +8,19 @@ a {
     text-decoration: none;
 }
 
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I decided to implement it slightly differently

PR #2000, opened 2023-08-18
Use CSS named variables standard

un-pogaz:
The CSS contain a implementation for [named variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties).

### DIFF
--- a/resources/templates/book_details.css
+++ b/resources/templates/book_details.css
@@ -31,7 +31,7 @@ table.fields td {
 table.fields td.title {
     font-weight: normal;
     font-style: italic;
-    color: palette(placeholder-text);
+    color: var(--text-color);
     text-align: right;
 }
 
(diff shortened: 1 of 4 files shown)


un-pogaz:
A have push on a parallele branche another commit that following this one.
He maybe to odd or superflue, I don't know. Should I add it to this pull request?

[un-pogaz:function-css_variables, commit d13d1e1](https://github.com/kovidgoyal/calibre/commit/d13d1e1c71eab4450996217278bf3c1986900d81)

kovidgoyal:
The reason to use palette() is that these colors are based on QPalette colors which can be set on the containing widget. Currently Qt doesnt support palette(placeholder-text) but hopefully they will fix that in the future.

PR #2016, opened 2023-09-09
Two commits:

cbhaley:
1) fix the DB handler so it doesn't update all notes when updating 1.
2) Support notes in manage categories and the tag browser.

Still to do: support notes in Manage Authors


cbhaley:
Added the edit authors dialog

PR #2017, opened 2023-09-09
fix inconsistency field type for default book

un-pogaz:
Fix a inconsistency that return a turple for custom fields with multiple value for the "default book" in the template editor, instead of a list.

mi.get('author') + mi.get('#custom_multiple') => TypeError - can only concatenate list (not "tuple") to list.


kovidgoyal:
@cbhaley I have merged this but please take a look as its your code.

cbhaley:
@kovid, this is fine. ProxyMetadata converts tuples to lists, so this manually-constructed metadata object should do the same.

kovidgoyal:
thanks

PR #2027, opened 2023-09-16
Add import and export note to the category editors.

cbhaley:


cbhaley:
Note: I am worthless at creating icons, so I used two that exist and work for me.

kovidgoyal:
I think it will be better to just have import/export in the edit note
dialog, no need for a profusion of buttons.

cbhaley:
What about putting the operations only in the context menu, leaving off the buttons?

kovidgoyal:
Context menu is okay, but I suggest waiting till I am done with the
export/import for the edit dialog, as there are likely to be some
changes in the API while I work on it. I will ping you when I am done
with it.

kovidgoyal:
I have finished the implementation. Export now exports to a single HTML
file with the images embedded in the file.

cbhaley:
You need to sanitize the item_val before passing it to the file chooser. I tried it with an item val of "\Bahh" and it gave some permission error. It would also fail if the item_val contains (for example) a colon.

kovidgoyal:
true, thanks.

PR #2035, opened 2023-09-22
Add retrieve all item_id for each field assosiated to a notes

un-pogaz:
Add a function that return a dict {field:[item_id, ], } of all notes defined in the DB.

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -975,6 +975,9 @@ class DB:
     def notes_data_for(self, field_name, item_id):
         return self.notes.get_note_data(self.conn, field_name, item_id)
 
+    def get_notes_id_map(self):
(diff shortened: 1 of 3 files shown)


kovidgoyal:
What's the use case for this function?

un-pogaz:
I prepare a plugin to support notes. In my selection dialog, I list all the field and value associated with a note, and exclude if no note present.
Actualy, I need to retrieve the list of all fields that supports notes, the full list of item_id for each of this fields, and then, get the note for each field/item_id pair to test if a note its presence or not.

This function take the problem in reverse, and retrieve the list/dict of field and value associated with a note in perform a single SQL request to the DB.

After enconter this little case in my plugin, I thing that it could really be a native function.

PR #2036, opened 2023-09-22
Use an icon for notes in book details.

cbhaley:
The icon comes from https://www.iconfinder.com/icons/8726310/notes_icon. Feel free to replace it. :)


kovidgoyal:
After adding a new note to an item that does not have an existing note the note icon doesnt show up until I switch to another book and back.

cbhaley:
Well, that opened a small can of worms. :) The same thing could happen with links in some cases.

PR #2039, opened 2023-09-24
build: pass on the user's compiler and compiler-flags settings to qmake

eli-schwartz:
PyQt-builder uses qmake under the hood, which has a frustrating habit of not respecting standard stuff like $CC / $CFLAGS unless the project authors go out of their way to manually specify to use the same value as gotten from the environment.

Implement this for the same environment variables that we currently respect when building internal extensions *without* PyQt-builder.

### DIFF
--- a/setup/build.py
+++ b/setup/build.py
@@ -209,6 +209,9 @@ class Environment(NamedTuple):
     cc: str
     cxx: str
     linker: str
+    base_cflags: List[str]
+    base_cxxflags: List[str]
+    base_ldflags: List[str]
     cflags: List[str]
     ldflags: List[str]
     make: str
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This is breaking the build on windows, see the CI

kovidgoyal:
Looks like qmake is passing option in both -option and /option form

eli-schwartz:
I made a dumb mistake. The linker on linux is the same as the "best" compiler for whatever language you are using (CXX for mixed binaries, CC for just-C binaries), but for Windows it has to be link.exe / lld-link.exe, and I forgot to take this into account. So the Windows CI attempted to link the extensions using cl.exe and that... did not work.

eli-schwartz:
Okay, thankfully the fixup has resulted in a clean bill of health under Windows CI, which I'll take as good news since I can't otherwise test it anyway. :D

PR #2043, opened 2023-09-28
fix crash in Get Books when regenerating UIC files

eli-schwartz:
Current versions of PyQt seem to generate code incompatible with our class definition:

```
  File "/home/eschwartz/git/calibre/src/calibre/gui2/actions/store.py", line 49, in do_search
    return self.search()
           ^^^^^^^^^^^^^
  File "/home/eschwartz/git/calibre/src/calibre/gui2/actions/store.py", line 55, in search
    sd = SearchDialog(self.gui, self.gui, query)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eschwartz/git/calibre/src/calibre/gui2/store/search/search.py", line 38, in __init__
    self.setupUi(self)
  File "/home/eschwartz/git/calibre/src/calibre/gui2/store/search/search_ui.py", line 84, in setupUi
    self.results_view = ResultsView(parent=self.verticalLayoutWidget)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ResultsView.__init__() got an unexpected keyword argument 'parent'
```

The issue is that arguments forwarded to the PyQt class are now being generated using keyword arguments rather than non-keyword args. As a direct wrapper over a class belonging to another project, we really should forward both types just in case.

### DIFF
--- a/src/calibre/gui2/store/search/results_view.py
+++ b/src/calibre/gui2/store/search/results_view.py
@@ -39,8 +39,8 @@ class ResultsView(QTreeView):
     download_requested = pyqtSignal(object)
     open_requested = pyqtSignal(object)
 
-    def __init__(self, *args):
-        QTreeView.__init__(self,*args)
+    def __init__(self, *args, **kwargs):
+        QTreeView.__init__(self,*args, **kwargs)
 
         self._model = Matches()
         self.setModel(self._model)


eli-schwartz:
This error is a bit odd since it does not occur on Arch, where `*_ui.py` does not get regenerated, but does occur on Gentoo, where the files do get regenerated.

The difference is that Arch packaging inherently resets timestamps for files in the tarball, but Gentoo does not. And inside the tarball, there are issues such as this:

```
$ bsdtar -tvf /var/cache/distfiles/calibre-6.27.0.tar.xz | grep search/search
-rw-r--r--  0 kovid  kovid   11132 May 27 22:04 calibre-6.27.0/src/calibre/gui2/store/search/search_ui.py
-rw-r--r--  0 kovid  kovid   11336 Sep 21 22:55 calibre-6.27.0/src/calibre/gui2/store/search/search.ui
```

It is not clear to me why the ui.py files have older timestamps than the .ui files despite not apparently getting regenerated as part of the release process.

eli-schwartz:
> It is not clear to me why the ui.py files have older timestamps than the .ui files despite not apparently getting regenerated as part of the release process.

879ebdf68edd17d9e5ccab628c5a2e79e56697b7

Oh, right, it is because you are using git archive to export most of the files, that makes sense...

Instead of zero-ing it out, it may also be possible to use `git log -1 --pretty=format:%ct HEAD` to get the mtime that git-archive embeds, and use it for the manually copied over files too. I suppose there's not much difference between the two.

kovidgoyal:
https://github.com/kovidgoyal/calibre/commit/1f296569f34e7600fd121a6e20ca51db10263d8e

PR #2055, opened 2023-10-04
Bug 2038470: Template error when trying to get Goodreads votes from Ids

cbhaley:
This is the "compatible" fix.


kovidgoyal:
As noted on the bug report I would like to fix this differently.

PR #2063, opened 2023-10-13
Update recipe for El Diplo

tdido:
This updates includes the following changes:

- Download the cover
- Add the author names below the article titles
- Group the articles into sections mimicking the publication's organisation as close as possible
- Cleanup anchors to avoid broken links


tdido:
@kovidgoyal I'd like to keep the original copyright for the previous author as well, since this is based on his work. What would the syntax be like?

PR #2069, opened 2023-10-16
calibre-server: don't call listen() on pre-allocated sockets

roshanshariff:
When using systemd socket activation, the pre-allocated socket is already
listening. There is no need to call `socket.listen()` on it again.

* `calibre/srv/loop.py`

Move `socket.listen()` (and log message) from `ServerLoop.serve()` to
`ServerLoop.initialize_socket()`, and skip it for pre-allocated sockets.

* `calibre/srv/embedded.py`

`Server.serve_forever()` should call `ServerLoop.serve()`, not
`ServerLoop.serve_forever()`. The latter function initializes the socket, which
is not needed because `Server.start()` has already called
`ServerLoop.initialize_socket()`.

I've tested that this fixes the socket activation issue in [bug 2039395](https://bugs.launchpad.net/calibre/+bug/2039395). I also
confirmed that it doesn't break starting the server without socket activation,
or the embedded server run from the Calibre GUI.

### DIFF
--- a/src/calibre/srv/embedded.py
+++ b/src/calibre/srv/embedded.py
@@ -131,7 +131,8 @@ class Server:
                 pass
         reset_caches()  # we reset the cache as the server tdir has changed
         try:
-            self.loop.serve_forever()
+            # self.loop.initialize_socket() already called in self.start()
+            self.loop.serve()
         except BaseException as e:
             self.exception = e
         if self.state_callback is not None:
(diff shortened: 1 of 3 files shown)


kovidgoyal:
THis is causing the socket activation test to fail. Note that you can run the test locally as calibre-debug -t socket_activation

roshanshariff:
@kovidgoyal, thanks! I've force-pushed an updated commit that calls `socket.listen()` at the beginning of the test.

PR #2070, opened 2023-10-17
Update wsj_free.recipe

unkn0w7n:
the whats news section is a repeat of articles from other sections, so removed it.

### DIFF
--- a/recipes/wsj.recipe
+++ b/recipes/wsj.recipe
@@ -22,6 +22,8 @@ try:
 except ImportError:
     from urllib import quote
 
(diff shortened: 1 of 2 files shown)


kovidgoyal:
IIRC wsj.recipe and wsj_free.recipe are clones of each other so they should
be changed in tandem keeping only the needs_subscription line different.

unkn0w7n:
hmm.. just made changes to the other recipe too.

idk if the login code (get browser) will create any problems with archive.is redirects. Can't we just change the needs subscription to False in wsj too.

PR #2076, opened 2023-10-23
...

unkn0w7n:
identifying tags and cleaning up html is a problem in archive.is


kovidgoyal:
why is there a difference between wsj and wsj_free in this PR

unkn0w7n:
i missed changing those lines in wsj_free.

PR #2084, opened 2023-10-28
Recipe to download a PressReader periodical

ping:


kovidgoyal:
I dont think this is suitable as a builtin recipe, since it requires
customization to use. I suggest instead converting it into a simple GUI
plugin, that asks the user for the login and periodical details, then
generates a customised recipe from the template and adds it to the
calibre customized recipes and schedules it appropriately.

kovidgoyal:
The CI error comes from calling browser() at the class level, dont do
that as the full calibre env is not guaranteed to exist at compile time.

PR #2089, opened 2023-10-31
fix NoteTableWidgetItem

un-pogaz:
Following #2088, here some fix to

1. canceling authors manage dialog not restore all notes in every case
2. restore original notes in non-authors manage dialog

### DIFF
--- a/bypy/sources.json
+++ b/bypy/sources.json
@@ -404,9 +404,9 @@
     {
         "name": "podofo",
         "unix": {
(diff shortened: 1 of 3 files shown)


kovidgoyal:
I'll hold off on merging this as @cbhaley says he is also working on these, since its his code, lets wait to hear from him.

kovidgoyal:
@cbhaley decided ot fix this differently, closing.

PR #2098, opened 2023-11-04
manage_categories: If the current cell is in a category, add editing …

cbhaley:
…notes and links for the book's individual items


cbhaley:
It uses manage catagories so one can edit all the data for the category item.

cbhaley:
Pushed a fix for publisher. Also for single-valued categories

PR #2101, opened 2023-11-05
Add a "Show items in selected books" choice to manage tags and manage…

cbhaley:
… authors.

Radio buttons are more visually intuitive than checkboxes, but it is a little strange not having a "Use all" radio button. You click the radio button again to turn it off. Adding the third button takes up more vertical space than it is worth. Switching back to checkboxes (that act like radio buttons) is trivial if you think that is a better visual appearance.


cbhaley:
I also cleaned up some text, making manage tags and manage authors use the same verbiage.

kovidgoyal:
I think I prefer checkboxes it would never have occurred to me to click a radio button again to clear it. Or if you want to save vertical space while being as clear as possible you can use a label like "Only show items from:" with a combobox to select the restriction.

cbhaley:
Fix pushed using a third radio button, using setContentsMargins to reclaim enough vertical space.

cbhaley:
Pushed a change to set the focus to the table widget whenever it is rebuilt.

PR #2104, opened 2023-11-06
Icon cache changes:

cbhaley:
1. Add an icon cache to QIcon.
2. Use the cache for every icon getter in these dialogs.

As I said in the comments, I didn't add the cache to QIcon.ic (__call__) because I don't know the rules. If 'names' are not context sensitive then the cache can be added to __call__ and cached_icon() can go away .


kovidgoyal:
There are sufficient complications with icon retrieval that I dont feel
comfortable using the cache in __call__. There are themes, plugin icons,
user overrides, etc. Using a cache is bound to change the semantics of
some of that, and I dont think icon creation is a performance issue
anywhere else anyway, if it is cached_icon() can be used in such places
as needed.

PR #2113, opened 2023-11-15
Update financial_times.recipe

unkn0w7n:
i think source srcset tag conflicts with img src on calibre ebook viewer, images wont show & opinions moved to the top so that they dont end up in other sections.


kovidgoyal:
the calibre viewer is a browser in browsers srcset takes precedence over
src so you should remve srcset or modify it in the recipe.

PR #2114, opened 2023-11-16
Remove deprecated utcnow, utcfromtimestamp

salty-horse:
[Python 3.12 has deprecated](https://docs.python.org/3.12/whatsnew/3.12.html#deprecated) `datetime.utcnow` and `datetime.utcfromtimestamp`.

In this change I updated most instances of it. I tested only on 3.10.1, but all of the changes I made should be compatible with 3.8+. (I couldn't figure out how to run Calibre with different versions of Python that I have installed. Is there a way to give `calibre-debug` a python executable to use?)

I tested the changes to `calibre/utils/date.py` thoroughly to have identical output before/after. The other recipes and scripts, I tested by separately running the changed snippets of code.

I left the uses of `datetime.utcfromtimestamp` in the following files, as I'm not sure where the output is being used, and whether it's ok for the datetimes to be timezone-aware or not.
* `src/calibre/devices/kindle/driver.py`
* `src/calibre/library/catalogs/epub_mobi_builder.py`

### DIFF
--- a/recipes/apple_daily.recipe
+++ b/recipes/apple_daily.recipe
@@ -5,7 +5,7 @@ __copyright__ = '2013-2015, Eddie Lau'
 __Date__ = ''
 
 from calibre import (__appname__, force_unicode, strftime)
-from calibre.utils.date import now as nowf
+from calibre.utils.date import now as nowf, utcnow
 import os
 import datetime
 import re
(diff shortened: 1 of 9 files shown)


eli-schwartz:
> I couldn't figure out how to run Calibre with different versions of Python that I have installed. Is there a way to give `calibre-debug` a python executable to use?

You can run without installing:
```
python3.XX ./run-local calibre-debug -g
```

Note that you'll first need to use `python3.XX setup.py build --clean && python3.XX setup.py build` in between each step, to recompile the C/C++ extensions against the python you wish to use.

salty-horse:
Thanks! Do you need me to test this more thoroughly with various versions?

BTW, I have a patch to use the built-in `datetime.fromisoformat` which in 3.11 achieved feature parity with Calibre's own C implementation. I assume it's useless since Calibre aims to support 3.8+. Should I still upload it for future use?

eli-schwartz:
I'm a humble contributor, I don't need you to test anything :P

But my personal opinion on the matter is that it seems fine to use such a builtin as part of an `if sys.version_info >= (3, 11)` check. This makes it easy to detect when its use can be dropped entirely.

PR #2115, opened 2023-11-16
Fix minimum required Python version in setup.py

salty-horse:
According to the development guide, and the usage of the `:=` operator, the minimum required Python version is 3.8.

`setup.py` still checked for 3.7, so I updated it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ __docformat__ = 'restructuredtext en'
 import sys, os
 
 
-def check_version_info(minver=(3, 7, 0)):
+def check_version_info(minver=(3, 8, 0)):
     vi = sys.version_info
     if vi < minver:
         def fmt(v):


eli-schwartz:
> and the usage of the `:=` operator

offtopic, but where is that currently used? I don't see it from a quick grep, and I think it basically stopped being used as a response to

salty-horse:
I found `:=` in [setup/wincross.py#L146](https://github.com/kovidgoyal/calibre/blob/master/setup/wincross.py#L146), but maybe it's not meant to be part of the project proper.

eli-schwartz:
I am pretty sure that's a private script of Kovid's, which he uses as part of `Install the Windows headers for cross compilation'`. The wincross.py script is only invoked using runpy, even, so it should not cause import-time errors when running `python setup.py build && python setup.py gui && python setup.py install`, on the very odd chance that someone out there is still using an EOL python to install calibre.

This may be a sign that he'd like to use `:=` more and should take the opportunity of officially bumping the minimum to start cleaning up old code in the project proper. 🤷

salty-horse:
>should take the opportunity of officially bumping the minimum to start cleaning up old code in the project proper.
I thought it's already at 3.8, since the [development guide](https://manual.calibre-ebook.com/develop.html) says so.

kovidgoyal:
Dropping 3.7 is fine since it is now EOLed by upstreamed.

PR #2119, opened 2023-11-18
Improve IPv6 support for server mode

qykth-git:
Improve IPv6 support:
* Use `::` as default server address. It also accepts IPv4.
* Display as `[host addr]:port` style for IPv6 address.

### DIFF
--- a/src/calibre/gui2/actions/device.py
+++ b/src/calibre/gui2/actions/device.py
@@ -27,7 +27,11 @@ def local_url_for_content_server():
     from calibre.srv.opts import server_config
     opts = server_config()
     interface = opts.listen_on or '0.0.0.0'
-    interface = {'0.0.0.0': '127.0.0.1', '::':'::1'}.get(interface)
+
+    addr_map = {'0.0.0.0': '127.0.0.1',
+                '::':      '::1'}
+    if interface in addr_map:
+        interface = addr_map[interface]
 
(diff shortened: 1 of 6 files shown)


kovidgoyal:
:: will also work for ipv4 only if the kernel knows about ipv6. This is not guaranteed as calibre is used on a wide variety of systems some of them quite old. So I dont feel comfortable changing the default as it could break existing installs on an update.

qykth-git:
Adds runtime check for IPv6 support.
This is done by `socket.has_dualstack_ipv6()` .

Are you feels better with this fix?

kovidgoyal:
That isnt good enough, someone could make some other change and save the setting in which case :: will be saved as the listen_on value. Then if that config file is used ona  computer without dualstack it will break. You need to check if listen_on == '::' when it is used and auto translate it to 0.0.0.0 when dual stack is not present.

qykth-git:
Use empty IP address as default `listen_on` value, and choose IPv4/v6 only when the `listen_on` address is empty.
How about this fix?

kovidgoyal:
That's better but what about zeroconf? If you are listening dualstack
then zerconf needs to publish dual stack IPs as well.

qykth-git:
zeroconf code is not touched yet.
IPv6 address might needs for zeroconf.

PR #2129, opened 2023-12-02
Changes to eliminate the black box cover_flow flash. The important …

cbhaley:
…ones are in ui.py. The changes in cover_flow.py are for timed logging.

Attached is a debug log showing the startup on my Win10 machine. You can see that resetting self.created_at has no effect. The old and new values are the same.

[debug_log.txt](https://github.com/kovidgoyal/calibre/files/13536590/debug_log.txt)

Note that all flashing goes away if the setMinimumSize on line 223 in cover_flow.py is removed. Another debug log is attached showing that run. The timings are rather different. This change (deletion) is not included in this PR.

[debug_log no setMinimumSize.txt](https://github.com/kovidgoyal/calibre/files/13536602/debug_log.no.setMinimumSize.txt)

### DIFF
--- a/src/calibre/gui2/cover_flow.py
+++ b/src/calibre/gui2/cover_flow.py
@@ -20,7 +20,7 @@ from qt.core import (
 
 from calibre.constants import islinux
 from calibre.ebooks.metadata import authors_to_string, rating_to_stars
-from calibre.gui2 import config, gprefs, rating_font
+from calibre.gui2 import config, gprefs, rating_font, timed_print
 from calibre_extensions import pictureflow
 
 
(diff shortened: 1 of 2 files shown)


kovidgoyal:
So the fix is basically starting the smart device later? Any chance starting the smart device can be moved to a thread?

cbhaley:
Changed to using a thread.

Here is the resulting startup debug log. There is still a small "flash" caused by the repaint of the tag browser & list view. This goes away if the setMinimumSize(300, 150) in cover_flow.py is removed.

[debug log.txt](https://github.com/kovidgoyal/calibre/files/13538852/debug.log.txt)

kovidgoyal:
So start_plugin() expects to be called on the GUI thread, and just calls the start_plugin() method of the dynamic plugin. I dont know if its safe to run that in a separate thread. For instance, I dont know if there is a race with dynamic_plugins being populated. Indeed as far as I can see that race exists anyway with the GUI starting this device at startup. But perhaps it's not practically a problem as the GUI thread is much slower than the device thread.

I am going to hold off on this for a bit. I will try to get rid of the whole nested splitters and debounced resize machinery which should make for much smoother resizes and properly fix startup flash.

kovidgoyal:
On Sun, Dec 03, 2023 at 08:50:11AM -0800, Charles Haley wrote:
> Changes to use a singleshot timer pushed. As noted in the commit, the changes in cover_flow.py are optional.

I got a report that QV is no longer remembering its size. It always
starts with a fixed size. Quoting the report:

Hi, Kovid.

I noticed you made some changes to avoid the black/white flashing at the startup.

Running from source, it seems to be fixed, but I got another problem now: the Quickview panel does not remember its size. It always start with a fixed size (way bigger then what I choose previously).

cbhaley:
Fixed. See PR https://github.com/kovidgoyal/calibre/pull/2133

PR #2130, opened 2023-12-03
Use canonical method to invoke sip-build tool

qykth-git:
Use canonical method to invoke sip-build tool

https://www.riverbankcomputing.com/static/Docs/sip/command_line_tools.html
> The build tools can also be invoked using the -m command line
> option of the Python interpreter. For example, the following
> two command lines are equivalent:
>   sip-build -h
>   python -m sipbuild.tools.build -h

### DIFF
--- a/setup/build.py
+++ b/setup/build.py
@@ -721,8 +721,8 @@ sip-file = {os.path.basename(sipf)!r}
             self.create_sip_build_skeleton(src_dir, ext)
             cwd = src_dir
             cmd = [
-                sys.executable, '-c',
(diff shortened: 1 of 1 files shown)


kovidgoyal:
It is deliberately done like this because there are installations out there that have python on path but not sip-build.

qykth-git:
This patch uses `python -m module` style.
If `python -c 'from sipbuild.tools.build import main; main();'`  works, Python module load path works for `sipbuild` module.
This means `python -m sipbuild.tools.build` also works.

PR #2134, opened 2023-12-08
Economist update

unkn0w7n:
facebook user_agent works.


kovidgoyal:
Removing the GDPR cookie will likely cause it to break for European
users. Does it actually need to be removed?

kovidgoyal:
Actually simply adding the accept-language header seems to get us past cloudflare.

unkn0w7n:
i tried googlebot and bingbot headers first, both did not work and then tried facebook headers. I wasn't expecting any random ua to work. 😁 Anyhow we did figure out a way to bypass that impossible cloudflare check.

I'm not sure for how long this workaround would last. Maybe we should figure out alternatives.

PR #2137, opened 2023-12-17
The tag browser notes/links icon stuff.

cbhaley:
Clicking the "buttons" works because of how mouse tracking works. The user must move the mouse over/into the current item line before clicking a button, which causes that line to be painted. When the line is painted the horizontal positions of the buttons on the line are recorded. If/when a click happens, the click X point is checked to see if it is in the ranges defined by the horizontal positions.


cbhaley:
Kovid, I found problems with the button click stuff. I need some sleep so will work more on it tomorrow.

cbhaley:
Pushed:
1) Fixes for the icon clicks. It now works when items are painted out of order. Resizing is now accounted for.  Toggling and setting item focus now works properly.
2) Added items to the tag browser config menu, including the possibility of shortcuts.

PR #2141, opened 2023-12-20
Fix calling plugin initialization_complete before it actually is …

cbhaley:
…complete.


kovidgoyal:
Hopefully the new central container will allow moving initilization mostly back into initialize() making this unnecessary. So will keep this open till I either implement tat or abandon it.

kovidgoyal:
I have merged the new layout code which obviates the necessity for delayed initialization.

PR #2145, opened 2023-12-28
Add -O3 to extension compile flags

codetheweb:
This speeds up .mobi/.azw3 conversions by quite a bit; around 30-50% on the lengthy .epub to .azw3 conversions I tested with:

<table>
<tr>
 <td>Title
 <td>previous conversion time (under -O2)
 <td>new conversion time (with -O3)
 <td>speedup
<tr>
 <td>A Tale of Two Cities
 <td>5s
 <td>3.87s
 <td>1.3x
<tr>
 <td>Les Misérables
 <td>22.8s
 <td>15.42s
 <td>1.5x
<tr>
 <td>War and Peace
 <td>21s
 <td>14.3s
 <td>1.5x
</table>

<details>
<summary>Benchmarking details</summary>

Books: [epubs.zip](https://github.com/kovidgoyal/calibre/files/13789869/epubs.zip)

Simple script written by ChatGPT:

```bash
#!/bin/bash

if [ $# -ne 1 ]; then
    echo "Usage: $0 <directory>"
    exit 1
fi

directory="$1"

if [ ! -d "$directory" ]; then
    echo "Error: $directory is not a directory."
    exit 1
fi

for file in "$directory"/*; do
    if [ -f "$file" ]; then
        echo "Benchmarking $file..."
        # Redirect stdout to /dev/null and measure the time taken
        time ./ebook-convert "$file" "~/temp/${file%.*}.azw3" > /dev/null 2>&1
        echo "----------------------------------------"
    fi
done

echo "Benchmarking complete."
```
</details>

-----------------------------------

Not sure if where I added the flag is the best place to put it, please let me know if there's a more appropriate place.

### DIFF
--- a/setup/build.py
+++ b/setup/build.py
@@ -353,6 +353,9 @@ def init_env(debug=False, sanitize=False, compiling_for='native'):
         else:
             cflags.extend('-I' + x for x in get_python_include_paths())
             ldflags.append('/LIBPATH:'+os.path.join(sysconfig.get_config_var('prefix'), 'libs'))
+
+    cflags.append('-O3')
+
     return Environment(
         platform_name=platform_name, dest_ext=dest_ext, std_prefix=std_prefix,
         base_cflags=base_cflags, base_cxxflags=base_cxxflags, base_ldflags=base_ldflags,
@@ -541,7 +544,6 @@ class Build(Command):
                 cflags = [
                     '-DCALIBRE_MODINIT_FUNC='
                     '{} __attribute__ ((visibility ("default"))) {}'.format(extern_decl, return_type)]
-                cflags.append('-O3')
             if ext.needs_cxx and ext.needs_cxx_std:
                 if env.cc_output_flag.startswith('/') and ext.needs_cxx == "11":
                     ext.needs_cxx = "14"
--- a/setup/build.py
+++ b/setup/build.py
@@ -541,6 +541,7 @@ class Build(Command):
                 cflags = [
                     '-DCALIBRE_MODINIT_FUNC='
                     '{} __attribute__ ((visibility ("default"))) {}'.format(extern_decl, return_type)]
+                cflags.append('-O3')
             if ext.needs_cxx and ext.needs_cxx_std:
                 if env.cc_output_flag.startswith('/') and ext.needs_cxx == "11":
                     ext.needs_cxx = "14"


eli-schwartz:
Currently no -O level is specified in setup/build.py (or anywhere inside the build system) did you test with a stock build or vs. setting -O2 in CFLAGS, or what?

codetheweb:
My bad, I edited the post. I don't fully understand the build system yet but it looks like `-O2` is currently used for compiling the extensions, as if I change the new line in this PR to be `-O2` I get similar timings to the `ebook-convert` binary in the latest Calibre release. The speedup is now not quite as dramatic 😅. But still pretty decent.

kovidgoyal:
IIRC and I could be wrong, O2 is the default when no optimization level is specified. This is because O3 can cause issues with some code. Whether any of those issues manifest in actual calibre code, one cannot know without trying. That is why I left it as default. The correct place to specify this is in the init_env() function. It should be done for all platforms, including windows. Since nowadays calibre has a more extensive test suite than before I suppose its worth trying if it passes all tests.

eli-schwartz:
> IIRC and I could be wrong, O2 is the default when no optimization level is specified.

Build systems often inject O2 if none is supplied, when producing Make or ninja files containing compiler command lines.

eli-schwartz:
> This is because O3 can cause issues with some code.

It enables more aggressive compiler optimizations, and correspondingly, runs the "risk" of taking code that has UB and tickling that UB even harder, optimizing harder on it to the point of producing buggy compiled programs.

If it's well tested (and confirmed to run faster) it's usually worth opting in.

codetheweb:
Moved in [49d791b](https://github.com/kovidgoyal/calibre/pull/2145/commits/49d791b1f05d1cb928e399fede3b14e74c40258a). If there's concern about other things breaking, I could try using pragmas to isolate the optimizations to `src/calibre/ebooks/compression/palmdoc.c` since it seems like that's where the majority of time is spent.

kovidgoyal:
On Thu, Dec 28, 2023 at 07:13:05PM -0800, Eli Schwartz wrote:
> > IIRC and I could be wrong, O2 is the default when no optimization level is specified.
> 
> Build systems often inject O2 if none is supplied, when producing Make or ninja files containing compiler command lines.

There is no build system involved here, calibre's build system directly
runs the compiler/linker.

kovidgoyal:
On Thu, Dec 28, 2023 at 07:16:22PM -0800, Eli Schwartz wrote:
> > This is because O3 can cause issues with some code.
> 
> It enables more aggressive compiler optimizations, and correspondingly, runs the "risk" of taking code that has UB and tickling that UB even harder, optimizing harder on it to the point of producing buggy compiled programs.

Exactly and a fair bit of the C/C++ code calibre has was inherited from
other projects rather than being written by me, so it makes me a little
nervous.

> 
> If it's well tested (and confirmed to run faster) it's usually worth opting in.

I doubt *all* of it is well tested. Still maybe worth rolling the dice.

kovidgoyal:
On Thu, Dec 28, 2023 at 07:21:11PM -0800, Max Isom wrote:
> Moved in [49d791b](https://github.com/kovidgoyal/calibre/pull/2145/commits/49d791b1f05d1cb928e399fede3b14e74c40258a). If there's concern about other things breaking, I could try using pragmas to isolate the optimizations to `src/calibre/ebooks/compression/palmdoc.c` since it seems like that's where the majority of time is spent.

Lets see if the PR passes all tests first.

eli-schwartz:
> There is no build system involved here, calibre's build system directly runs the compiler/linker.

I understand that, and I'm saying that "build systems" are why it might seep into the general public knowledge that "O2 is the default, isn't it?"

> Exactly and a fair bit of the C/C++ code calibre has was inherited from other projects rather than being written by me, so it makes me a little nervous.

The good news :D is that compilers get better and better all the time at taking advantage of UB even with O2. I'd say it's something to be aware of, but not worry too much about. Given the testsuite covers it, and the CI runs the testsuite with sanitizers, I'd feel extra reassured that it's not a big problem if I were you.

Given benchmarks show a decent speedup, I'm inclined to say the dice are worth rolling.

PR #2153, opened 2024-01-06
Add an option to choose where to show the cover browser when using …

cbhaley:
…the narrow layout.

The change is more complicated because the cover browser cannot be shown on the right while in wide mode.

Note: there is a problem with the cover browser splitter when using the CB on top in narrow mode. If you use the the splitter handle to make the CB disappear then the splitter doesn't come back when you use the layout button to show the CB. You must hide the CB then find the handle and drag it open.

Steps:
1. Set the CB to show on top.
2. Lift the splitter handle until the CB disappears.
3. Click the CB layout button to redisplay the CB.

The splitter is gone. To get the splitter back:
1. Use the layout button to hide the CB.
2. Search with the mouse until you find the well-hidden splitter above the book list.
3. Pull that splitter down until the CB reappears.

The splitter is again shown.

I looked for how to redisplay the splitter when the CB is displayed with the button but was foiled.

### DIFF
--- a/src/calibre/gui2/__init__.py
+++ b/src/calibre/gui2/__init__.py
@@ -426,6 +426,7 @@ def create_defs():
     defs['show_links_in_tag_browser'] = False
     defs['show_notes_in_tag_browser'] = False
     defs['icons_on_right_in_tag_browser'] = True
+    defs['cover_browser_on_right'] = True
 
     def migrate_tweak(tweak_name, pref_name):
         # If the tweak has been changed then leave the tweak in the file so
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Fixed by https://github.com/kovidgoyal/calibre/commit/cfc21828d1c8971b83f443a9419124971bc2f2c2

And this option needs to be a three way, auto, right, top

cbhaley:
Closed because redone in another commit for adding "auto"

PR #2161, opened 2024-01-13
Update livemint.recipe

unkn0w7n:


unkn0w7n:
I want to remove empty p tags in this recipe.
is this the right way to remove empty p tags

for p in soup.findAll('p'):
     if not p.find('img'):
        if self.tag_to_string(p).strip() == '':
           p.extract()
           
wouldn't this unnecessarily put strain on cpu to go through all p tags?
is there a better way?

kovidgoyal:
No, if you want to find empty p tags you have to go through all of them
in the document. You could checking if a tag is empty more efficient,
depending on just what you mean by empty. You can use

bool(p.string)
or
bool(p.contents)

for instance.

PR #2166, opened 2024-01-19
Fix for edit widget placement problem

cbhaley:
described in https://www.mobileread.com/forums/showthread.php?p=4390503#post4390503


cbhaley:
A better fix might be to force displaying a scroll bar then hiding it. This should make the behavior the same as it was. Unfortunately I don't see how one might do this.

kovidgoyal:
To get the same behavior as before, you should probably use view.rect()
- verticalScrollbar().width()

PR #2169, opened 2024-01-25
Fix for the cover render hang discussed in …

cbhaley:
… https://www.mobileread.com/forums/showthread.php?t=358705.

- All Qt operations are now done on the GUI thread. File operations are done on a separate thread.
- QPixmap is used instead of QImage.
- The ThumbnailCache is now versioned.

The rendering operations are done using a FunctionDispatcher instead of a Qt signal to avoid a possible long queue of render operations causing the UI to be sluggish.

Emblems etc aren't rendered if the cover cache hasn't been set for the book. This reduces some subtle flashing as the cover is repainted. However, covers are seen as empty regions until the cover is actually rendered.

I also fixed a bug where the wrong cache would be used after a switch library. The UUID used for the cache was set at thread creation time instead of set_database time.

### DIFF
--- a/src/calibre/db/tests/utils.py
+++ b/src/calibre/db/tests/utils.py
@@ -4,6 +4,7 @@
 __license__ = 'GPL v3'
 __copyright__ = '2013, Kovid Goyal <kovid at kovidgoyal.net>'
 
+import os
 import shutil
 
 from calibre import walk
@@ -71,12 +72,12 @@ class UtilsTest(BaseTest):
         c.empty()
         self.assertEqual(c.total_size, 0)
         self.assertEqual(len(c), 0)
-        self.assertEqual(tuple(walk(c.location)), ())
+        self.assertEqual(tuple(walk(c.location)), (os.path.join(c.location, 'version'),))
         c = self.init_tc()
         self.basic_fill(c)
         self.assertEqual(len(c), 5)
         c.set_thumbnail_size(200, 201)
         self.assertIsNone(c[1][0])
         self.assertEqual(len(c), 0)
-        self.assertEqual(tuple(walk(c.location)), ())
+        self.assertEqual(tuple(walk(c.location)), (os.path.join(c.location, 'version'),))
     # }}}
--- a/src/calibre/db/utils.py
+++ b/src/calibre/db/utils.py
@@ -7,6 +7,7 @@ __copyright__ = '2013, Kovid Goyal <kovid at kovidgoyal.net>'
 import errno
 import os
 import re
+import shutil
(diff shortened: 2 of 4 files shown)


kovidgoyal:
Looks good to me, and seems to perform better than the old code, in
my rather unscientific tests. One question:

Would resizing in the fetch thread be worth it? It means less
blocking work in the GUI thread. It can be done using the PIL package
already bundled with calibre avoiding Qt locking issues. So CoverTuple
would grow an additional field to hold the thumbnail as a PIL image
object. In the gui thread the 
https://pillow.readthedocs.io/en/stable/reference/ImageQt.html

can be used to directly convert the thumbnail to a QImage and then a
QPixmap.

I dont know if the additional code complexity is worth it, since you
have spent much more time in this code recently, I defer to your
judgement.

Regardless of the answer to that question, I will merge.

cbhaley:
Do you mean creating a resized PIL image in the cover thread and storing that in the cache. then using PIL to create the QImage/QPixmap from the PIL image in the GUI thread? The PIL page says that it uses a subclass of QImage so I hesitate to put that operation in the thread.

I don't think it would change performance much. Depends on the speed of PIL I suppose. It could make painting covers faster, perhaps at near cover cache speed, because there is much less work to do on the GUI thread. I am not convinced that it is worth storing the QPixmap made from the PIL image in the disk cache because it is used mostly once to be put into CoverCache. The cost of storing and fetching the pixmap might exceed the cost of converting the PIL image to a pixmap.

In any event I will look at it, if for no other reason than to learn something new (PIL).

kovidgoyal:
On Thu, Jan 25, 2024 at 08:09:23AM -0800, Charles Haley wrote:
> Do you mean creating a resized PIL image in the cover thread and storing that in the cache. then using PIL to create the QImage/QPixmap from the PIL image in the GUI thread? The PIL page says that it uses a subclass of QImage so I hesitate to put that operation in the thread.

So in the covers thread you would create a regular PIL image and store
it (PIL.Image). Additionaly you would pre-calculate the thumbnail by doing
PIL.Image.thumbnail() (or whatever the method is called) which creates a new PIL Image. 

Both the thumbnail and the image would be transmitted in CoverData to the GUI thread.
There you would do QPixmap.fromImage(PIL.ImageQt(cover_data.thumbnail))
to render the thumbnail.

Or some variation thereof.

> In any event I will look at it, if for no other reason than to learn something new (PIL).

Yeah, that's always fun. I am busy teaching myself assembly programming
in order to use SIMD instructions for another project of mine, enjoying
myself thoroughly.

PR #2171, opened 2024-01-26
Change to using PIL for image rendering. This lets us do all …

cbhaley:
…thumbnail generation on the cover thread.

I couldn't make the PIL ImageQt stuff work. It seems to use a common butter somewhere. Generating and adding an additional image to the CoverCache corrupted previously added images. This implementation avoids this problem by passing a byte string of the PIL thumbnail to re_render(), which creates the QPixmap on the GUI thread. As a side effect, all the CoverCache operations are now on the GUI thread.

This implementation is visibly faster than the last one.

### DIFF
--- a/src/calibre/gui2/library/alternate_views.py
+++ b/src/calibre/gui2/library/alternate_views.py
@@ -5,12 +5,14 @@ __license__ = 'GPL v3'
 __copyright__ = '2013, Kovid Goyal <kovid at kovidgoyal.net>'
 
 import itertools
+from io import BytesIO
 import math
 import operator
 import os
 import weakref
 from collections import namedtuple
 from functools import wraps
+from PIL import Image
 from qt.core import (
     QAbstractItemView, QApplication, QBuffer, QByteArray, QColor, QDrag, QEasingCurve,
     QEvent, QFont, QHelpEvent, QIcon, QImage, QIODevice, QItemSelection,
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Hmm looks visibly slower on my system. One tip is instead of using
BytesIO() use ReadOnlyFileBuffer() from speedups.py for Image.open()
Assuming PIL works with fake file objects that return memoryview rather
than bytes that will avoid a whole extra copy of the image data.

Also, I will write a utility function to efficiently convert a PIL image
function to a QPixmap. You are right in that the ImageQt module is
pretty poor. I'll update here once the function is written.

kovidgoyal:
I think I miscommunicated. make_thumbnail is run on a non-GUI thread,
therefore it should return the PIL Image object. The GUI thread should
convert it to a pixmap using convert_PIL_image_to_pixmap
otherwise we are back to using Qt code in a non-GUI thread.

kovidgoyal:
Or maybe make_thumbnail() is called on the GUI thread? The comment at
the start says its called on the cover thread?

cbhaley:
I did put convert_PIL_image_to_pixmap() in the cover thread, which I should have realized wasn't right. The latest commit ensures that just the PIL thumbnail Image is created on the cover thread (no Qt). That image is passed to re_render() on the GUI thread where it is converted to a QPixmap with convert_PIL_image_to_pixmap().

kovidgoyal:
Merged with some optimisations.

PR #2177, opened 2024-01-31
Plugin results are not limited to a single relevant result

leakymemory:
Background:
I recently started using Calibre to organize my TTRPG PDFs and found a plugin which connects to RPGGeek to download additional metadata for individual books. However, in many cases I found that the top search result was incorrect and in looking at the data returned, found that the correct book was further down the list. Because plugins are limited to a single relevant result, I was only seeing that top result in the metadata dialog.

Fix:
Add an additional property to the plugin configuration, "maximum_relevant_results" which allows the plugin author to override the default (1) and return show those additional results in the download metadata dialog.

RPGGeek Plugin: https://github.com/ErikLevin/calibre_rpggeek_plugin

### DIFF
--- a/src/calibre/customize/__init__.py
+++ b/src/calibre/customize/__init__.py
@@ -89,6 +89,10 @@ class Plugin:  # {{{
     #: care.
     can_be_disabled = True
 
+    #: The maximum number of results which can be returned by the plugin
+    #: Default relevant results is 1.
+    maximum_relevant_results = 1
+
     #: The type of this plugin. Used for categorizing plugins in the
     #: GUI
     type = _('Base')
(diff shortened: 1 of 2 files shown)


kovidgoyal:
I dont quite follow. Why not use the existing keep_dups setting for this? Accessible as Preferences->Metadata sources->Keep more than one result per source.

The only caveat there is results with the exact same title+authors or identifiers are still discarded. Are you saying the rpg geek plugin has multiple results with identical title+author but that are different in some other way?

leakymemory:
Sigh.  No, this works fine.  I'm embarrassed (and sorry) that I missed that.  PR rescinded. :)

PR #2180, opened 2024-02-03
Replace deprecated urlopen cafile parameter

salty-horse:
The `cafile` parameter of `urlopen` has been deprecated since python 3.6, and is [removed in the upcoming 3.13](https://docs.python.org/3.13/whatsnew/3.13.html#urllib).

It's used in the `setup.py recent_uas` command. I replaced it with an SSL context, and cleaned up the json dumping code.

Note: I don't really use this feature - I just searched for deprecated functions.

### DIFF
--- a/setup/resources.py
+++ b/setup/resources.py
@@ -166,8 +166,8 @@ class RecentUAs(Command):  # {{{
     def run(self, opts):
         from setup.browser_data import get_data
         data = get_data()
-        with open(self.UA_PATH, 'wb') as f:
-            f.write(json.dumps(data, indent=2, ensure_ascii=False, sort_keys=True).encode('utf-8'))
+        with open(self.UA_PATH, 'w', encoding='utf-8') as f:
+            json.dump(data, f, indent=2, ensure_ascii=False, sort_keys=True)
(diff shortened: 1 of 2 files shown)


setup/resources.py
```diff
@@ -166,8 +166,8 @@ class RecentUAs(Command):  # {{{
     def run(self, opts):
         from setup.browser_data import get_data
         data = get_data()
-        with open(self.UA_PATH, 'wb') as f:
-            f.write(json.dumps(data, indent=2, ensure_ascii=False, sort_keys=True).encode('utf-8'))
+        with open(self.UA_PATH, 'w', encoding='utf-8') as f:
+            json.dump(data, f, indent=2, ensure_ascii=False, sort_keys=True)
```
eli-schwartz, line 170: This change is not mentioned in the commit message and is totally unrelated. (It is also just painting the bikeshed...)

salty-horse: I know, and it's atypical of me. I apologize, and will split them up in the future.

It's not *completely* unrelated as it's part of the same "recent UAs" functionality.

PR #2188, opened 2024-02-13
update news.py

unkn0w7n:
remove svg and button tags in all recipes.

### DIFF
--- a/src/calibre/web/feeds/news.py
+++ b/src/calibre/web/feeds/news.py
@@ -1032,8 +1032,8 @@ class BasicNewsRecipe(Recipe):
         for attr in self.remove_attributes:
             for x in soup.findAll(attrs={attr:True}):
                 del x[attr]
-        for bad_tag in list(soup.findAll(['base', 'iframe', 'canvas', 'embed',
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Removing button is fine, but svg can be useful.

kovidgoyal:
There are some news sources that uses SVG for charts/graphs.

unkn0w7n:
oh.. most graphs i saw were in iframe. 
so adding it to remove tags based on recipe is better.

maybe we can add something like `if svg.findParent('a', attrs={'href':lambda x: x and ('twitter' or 'facebook' or 'wa') in x}):`

kovidgoyal:
On Mon, Feb 12, 2024 at 11:41:08PM -0800, unkn0w7n wrote:
> maybe we can add something like `if svg.findParent('a', attrs={'href':lambda x: x and ('twitter' or 'facebook' or 'wa') in x}):`

It's too dicey, for example

<a href="/article-about-facebook">...<svg>

And in future if new social websites pop up, it will stop working
anyway.

PR #2190, opened 2024-02-15
tag_to_string

unkn0w7n:
add space when joining strings.
if text is present in a sub div tag, it gets added without space between two strings.

if double spaced it'll get normalized anyhow.

also added 'button' to bad_tags.

### DIFF
--- a/src/calibre/web/feeds/news.py
+++ b/src/calibre/web/feeds/news.py
@@ -1032,7 +1032,7 @@ class BasicNewsRecipe(Recipe):
         for attr in self.remove_attributes:
             for x in soup.findAll(attrs={attr:True}):
                 del x[attr]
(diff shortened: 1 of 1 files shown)


kovidgoyal:
That will cause <div>a<span>b</span>c to be rendered as "a b c" instead of abc.

kovidgoyal:
That will cause <div>a<span>b</span>c to be rendered as "a b c" instead of abc.

unkn0w7n:
hmm.. I see your point, but I think most span tags are used for words than letters, maybe span tags with letters are a lot rarer when compared to sub tags.

I was thinking of description part in some recipes. It adds two diff groups of text without space as they are present in sub div or span tags.

kovidgoyal:
If you want to handle that, only add a space if the tag name is p or div
or section or article or main that should cover almost all cases, which
you can do in the loop.

PR #2193, opened 2024-02-18
support building from git source without internet

wrvsrx:
Currently, bootstrap of calibre need access to internet, which makes it hard to build it from git source in sandbox (Here `sandbox` means an environment without internet access) such as nix build environment.

This PR make it possible to build calibre without internet access, as long as we have downloaded needed resources and pass them to calibre bootstrap using environment variable. Now calibre can be built from git source without internet (In https://github.com/NixOS/nixpkgs/pull/289684, I build calibre from git source by patching this PR to calibre source code).

### DIFF
--- a/setup/hyphenation.py
+++ b/setup/hyphenation.py
@@ -83,7 +83,7 @@ class Hyphenation(ReVendor):
     description = 'Download the hyphenation dictionaries'
     NAME = 'hyphenation'
     TAR_NAME = 'hyphenation dictionaries'
-    VERSION = 'master'
+    VERSION = 'libreoffice-7.6.5.2'
     DOWNLOAD_URL = 'https://github.com/LibreOffice/dictionaries/archive/%s.tar.gz' % VERSION
     CAN_USE_SYSTEM_VERSION = False
 
--- a/setup/install.py
+++ b/setup/install.py
@@ -366,21 +366,25 @@ class Bootstrap(Command):
 
     def pre_sub_commands(self, opts):
         tdir = self.j(self.d(self.SRC), 'translations')
-        clone_cmd = [
-            'git', 'clone', f'https://github.com/{self.TRANSLATIONS_REPO}.git', 'translations']
-        if opts.ephemeral:
-            if os.path.exists(tdir):
-                shutil.rmtree(tdir)
-
-            st = time.time()
-            clone_cmd.insert(2, '--depth=1')
-            subprocess.check_call(clone_cmd, cwd=self.d(self.SRC))
-            print('Downloaded translations in %d seconds' % int(time.time() - st))
+        env = os.getenv('TRANSLATIONS')
+        if env:
+            subprocess.check_call(['cp', '-r', '--no-preserve=mode,ownership', env, 'translations'])
         else:
-            if os.path.exists(tdir):
-                subprocess.check_call(['git', 'pull'], cwd=tdir)
-            else:
+            clone_cmd = [
+                'git', 'clone', f'https://github.com/{self.TRANSLATIONS_REPO}.git', 'translations']
+            if opts.ephemeral:
+                if os.path.exists(tdir):
+                    shutil.rmtree(tdir)
+
+                st = time.time()
+                clone_cmd.insert(2, '--depth=1')
                 subprocess.check_call(clone_cmd, cwd=self.d(self.SRC))
+                print('Downloaded translations in %d seconds' % int(time.time() - st))
+            else:
+                if os.path.exists(tdir):
+                    subprocess.check_call(['git', 'pull'], cwd=tdir)
+                else:
+                    subprocess.check_call(clone_cmd, cwd=self.d(self.SRC))
 
     def run(self, opts):
         self.info('\n\nAll done! You should now be able to run "%s setup.py install" to install calibre' % sys.executable)
--- a/setup/resources.py
+++ b/setup/resources.py
@@ -143,9 +143,14 @@ class CACerts(Command):  # {{{
             if err.errno != errno.ENOENT:
                 raise
             raw = b''
-        nraw = download_securely('https://curl.haxx.se/ca/cacert.pem')
-        if not nraw:
-            raise RuntimeError('Failed to download CA cert bundle')
+        env = os.getenv('CACERT')
+        if env:
+            with open(env, 'rb') as f:
+                nraw = f.read()
+        else:
+            nraw = download_securely('https://curl.haxx.se/ca/cacert.pem')
+            if not nraw:
+                raise RuntimeError('Failed to download CA cert bundle')
         if nraw != raw:
             self.info('Updating Mozilla CA certificates')
             with open(self.CA_PATH, 'wb') as f:
--- a/setup/browser_data.py
+++ b/setup/browser_data.py
@@ -23,9 +23,15 @@ def filter_ans(ans):
 
 def common_user_agents():
     print('Getting recent UAs...')
-    raw = download_from_calibre_server('https://code.calibre-ebook.com/ua-popularity')
+    ua_env = os.getenv('UA_POPULARITY')
+    if ua_env:
+        with open(ua_env, 'r') as f:
+            content = f.read()
+    else:
+        raw = download_from_calibre_server('https://code.calibre-ebook.com/ua-popularity')
+        content = bz2.decompress(raw).decode('utf-8')
     ans = {}
-    for line in bz2.decompress(raw).decode('utf-8').splitlines():
+    for line in content.splitlines():
         count, ua = line.partition(':')[::2]
         count = int(count.strip())
         ua = ua.strip()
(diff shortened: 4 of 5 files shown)


kovidgoyal:
Why would you want to build from git source? calibre releases source
tarballs, with all resources included that you dont need internet to
build from.

wrvsrx:
Because the release tarball contains some generated js code and pyqt code, patching python code in tarball cannot modify the that generated code.

For example I want to apply https://github.com/kovidgoyal/calibre/commit/14caed3a6f0f57e172d1eb407703f0e4cc179e11 to fix https://www.mobileread.com/forums/showthread.php?t=359082 before next release.

kovidgoyal:
I'm afraid I am not going to commit to maintaining the ability too build
from git source in perpetuity for this. It should be trivial to diff the
generated js between the release tarball and what you get from git
checkout to make the patch against the release tarball.

wrvsrx:
> It should be trivial to diff the generated js between the release tarball and what you get from git checkout to make the patch against the release tarball.

I'm afraid not. Different versions of RapydScript and pyqt generates very different code. We have to guess which version of RapydScript and pyqt is used when you package the tarball, otherwise we'll generate a big patch containing a lot of useless diff.

kovidgoyal:
On Sun, Feb 18, 2024 at 07:35:51AM -0800, wrvsrx wrote:
> > It should be trivial to diff the generated js between the release tarball and what you get from git checkout to make the patch against the release tarball.
> 
> I'm afraid not. Different versions of RapydScript and pyqt generates very different code. We have to guess which version of RapydScript and pyqt is used when you package the tarball, otherwise we'll generate a big patch containing a lot of useless diff.

Since I am the maintainer of rapydscript I know for a fact that it does
not infact generate very different JS in different versions, and indeed
is in maintenance mode so has no different version beyond trivial bug
fixes.

As for PyQt its irrelevant to the patch in question.

setup/hyphenation.py
```diff
@@ -83,7 +83,7 @@ class Hyphenation(ReVendor):
     description = 'Download the hyphenation dictionaries'
     NAME = 'hyphenation'
     TAR_NAME = 'hyphenation dictionaries'
-    VERSION = 'master'
+    VERSION = 'libreoffice-7.6.5.2'
```
eli-schwartz, line 86: This change is bad and should not be done.

```
$ python setup.py hyphenation --help
[...]
  --path-to-hyphenation=PATH_TO_HYPHENATION
                        Path to the extracted hyphenation dictionaries source
  --hyphenation-url=HYPHENATION_URL
                        URL to hyphenation dictionaries source archive in
                        tar.gz format
```

I implemented this in https://github.com/kovidgoyal/calibre/pull/1074 for... non-network builds, as it happens. My motivation was actually to produce a recipe that built calibre from git master, not from git tags.

wrvsrx:
> Since I am the maintainer of rapydscript I know for a fact that it does not infact generate very different JS in different versions, and indeed is in maintenance mode so has no different version beyond trivial bug fixes.

It does. Run `python setup.py bootstrap` in different environments generate very different js code. I got very different results using v7.5.1 git source. Here's the output of `diff new/resources/viewer.js old/resources/viewer.js > a.diff`

[diff new/resources/viewer.js old/resources/viewer.js 
](https://gist.github.com/wrvsrx/0ed29ae9144eab097ddfeaa6de53c049)

setup/install.py
```diff
@@ -366,21 +366,25 @@ def add_options(self, parser):
...
-                shutil.rmtree(tdir)
-
-            st = time.time()
-            clone_cmd.insert(2, '--depth=1')
-            subprocess.check_call(clone_cmd, cwd=self.d(self.SRC))
-            print('Downloaded translations in %d seconds' % int(time.time() - st))
+        env = os.getenv('TRANSLATIONS')
+        if env:
+            subprocess.check_call(['cp', '-r', '--no-preserve=mode,ownership', env, 'translations'])
         else:
```
eli-schwartz, line 372: This all looks way too complicated, and I think it makes far more sense to allow `git pull` to fail here if the network is down or blocked.

It turns out that this is already possible, and I did it a long time ago: just copy the translations in on your own, and then run `for stage in build iso639 iso3166 translations gui resources cacerts recent_uas; do python setup.py $stage; done` instead of running bootstrap directly.

setup/resources.py
```diff
@@ -143,9 +143,14 @@ def run(self, opts):
...
-        if not nraw:
-            raise RuntimeError('Failed to download CA cert bundle')
+        env = os.getenv('CACERT')
+        if env:
+            with open(env, 'rb') as f:
+                nraw = f.read()
+        else:
+            nraw = download_securely('https://curl.haxx.se/ca/cacert.pem')
+            if not nraw:
+                raise RuntimeError('Failed to download CA cert bundle')
```
eli-schwartz, line 153: Just do not run this command at all, if you've copied it in place yourself...

eli-schwartz (commented):
> Why would you want to build from git source? calibre releases source tarballs, with all resources included that you dont need internet to build from.


> Because the release tarball contains some generated js code and pyqt code, patching python code in tarball cannot modify the that generated code.

This is incorrect logic. You can build from the release tarball and then patch the pyj or pyqt code, then re-run the build steps that locally generate those files. In no way does this obligate you to manually sideload your own data here.

kovidgoyal:
On Sun, Feb 18, 2024 at 08:05:06AM -0800, wrvsrx wrote:
> > Since I am the maintainer of rapydscript I know for a fact that it does not infact generate very different JS in different versions, and indeed is in maintenance mode so has no different version beyond trivial bug fixes.
> 
> It does. Run `python setup.py bootstrap` in different environments generate very different js code. I got very different results using v7.5.1 git source. Here's the output of `diff new/resources/viewer.js old/resources/viewer.js > a.diff`

That will have nothing to do with rapydscript versions, indeed the last
release of rapydscript was from three years ago, IIRC.

Regardless, all you need to do is build from source from the commit
before the fix commit, and then at the fix commit and diff the generated
js, that will give you a small patch you can apply to the released js.

Or just wait the typical 2-3 weeks between calibre releases. Imagine
that instead of fixing this bug a few hours after it was reported I
waited a couple of weeks :)

setup/install.py
wrvsrx: I just keep all original logics in `else` block, the diff is misleading. I plan to submit commits related to the git pull logic here later.

setup/resources.py
wrvsrx: I just keep all original logics in else block. I planned to do cleanup work in following commits if this PR has been accpeted.

setup/hyphenation.py
wrvsrx: You are right, this change is unnecessary.

eli-schwartz:
I have vague memories of looking at rapydscript-ng generated JS one time and seeing trivial changes in the source code result in numeric counts of internal variables get thrown off -- resulting in a relatively large diff of lines where e.g. `ρσ_Iter5` became `ρσ_Iter4` but nothing else changed.

This does not explain all the changes in @wrvsrx's diff...

Also it is *still* true that you don't need to apply a diff file when you can just package rapydscript-ng and use it. :)

I actually did have to jump through a couple hoops to package it for Gentoo: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/rapydscript-ng/rapydscript-ng-0.7.22.ebuild

This is because Gentoo does a network sandbox -- Arch doesn't -- so for Gentoo I had to have a conditionalized ebuild that can be run in live mode to produce a node_modules directory, tar it up, and release it as a custom Gentoo-specific asset downloaded by the stable ebuild.

As a result, I simply do not worry, period, about changes to pyj files and can freely backport them *if* I feel that's worthwhile. Though primarily I use it since I always try to package calibre using the system mathjax where possible, and hence have to regenerate the rapydscript as well.

It is such a tremendous not-deal. I don't get the problem here.

kovidgoyal:
@eli-schwartz IIRC you dont actually need rapydscript-ng anymore. calibre includes the rapydscript compiler and runs it via PyQtWebEngine. rapydscript-ng is faster, but otherwise it shouldnt matter.

wrvsrx:
Maybe it's caused by differences in some dependencies of rapidscript or python version... Thanks for your responsing.

eli-schwartz:
> @eli-schwartz IIRC you dont actually need rapydscript-ng anymore. calibre includes the rapydscript-ng compiler and runs it via PyQtWebEngine. rapydscript-ng is faster, but otherwise it shouldnt matter.

Somehow webengine crashes inside the portage sandbox, IIRC. Something something running an entire setuid chromium as a javascript engine. Also tried to create various files in /usr. I figured why bother trying to debug this when rapydscript-ng is faster anyways. ;)

wrvsrx:
The big diff might caused by that different versions of python give different ASTs.

PR #2202, opened 2024-02-21
Fix Content Server Web Interface unhandled error when sorting books (Launchpad Bug #2054170)

SammyD2002:
When sorting, the UserSessionData.set function was using the key 'key-||-library_id'  to search the all_settings dictionary and determine if it is set locally. However, since all_settings doesn't seem to keep track of the user's settings with the library_id in the key, the resulting value was undefined and an error was thrown upon attempting to use .local on it. To address this issue, I made sure the library_id was ignored for that check.

### DIFF
--- a/src/pyj/session.pyj
+++ b/src/pyj/session.pyj
@@ -326,7 +326,7 @@ def get_translations(newval):
 def is_setting_local(name):
     m = all_settings[name]
     return m.is_local if m else True
-
+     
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
By, https://github.com/kovidgoyal/calibre/commit/14caed3a6f0f57e172d1eb407703f0e4cc179e11

PR #2211, opened 2024-03-03
Two independent commits

cbhaley:
1) Fix save-to-disk templates containing template expressions like {:'current_library_name()'} printing an exception. The exception doesn't appear to break anything, but it shouldn't be there.

2) Use a timer to debounce book details display. This makes startup faster and scrolling smoother.

### DIFF
--- a/src/calibre/gui2/library/models.py
+++ b/src/calibre/gui2/library/models.py
@@ -16,7 +16,7 @@ from collections import defaultdict, namedtuple
 from itertools import groupby
 from qt.core import (
     QAbstractTableModel, QApplication, QColor, QFont, QFontMetrics, QIcon, QImage,
-    QModelIndex, QPainter, QPixmap, Qt, pyqtSignal,
+    QModelIndex, QPainter, QPixmap, Qt, QTimer, pyqtSignal,
 )
 
 from calibre import (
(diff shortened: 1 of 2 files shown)


kovidgoyal:
The debounce should be implemented on the book details side.
current_changed() may be used in some other components and so shouldnt
have semantics needed for book details. Instead create a new slot in the
book details that debounces and connect that to current_changed.

cbhaley:
Unfortunately that doesn't solve the entire problem. Currently, when current_change is triggered model() indirectly calls get_metadata() (not proxy_metadata). This forces the evaluation of all composite columns, which can take a lot of time for no purpose if the data won't be used.

I will look for alternatives. One that suggests itself is to debounce emitting new_bookdisplay_data. I don't think this has the same problem as debouncing current_changed.

Another companion change would be to have get_metadata() set the value of a composite to None instead of evaluating it. This makes it be evaluated on demand by __get_attribute__ in metadata.book.base(). If I do this then I need to be sure that there aren't any backdoors where a None composite isn't evaluated, for example in get_user_metadata() and get_all_user_metadata().

You can close this PR. I'll make another if I satisfy myself that it works and is worth the trouble.

kovidgoyal:
On Mon, Mar 04, 2024 at 01:13:36AM -0800, Charles Haley wrote:
> Unfortunately that doesn't solve the entire problem. Currently, when current_change is triggered model() indirectly calls get_metadata() (not proxy_metadata). This forces the evaluation of all composite columns, which can take a lot of time for no purpose if the data won't be used.
> 
> I will look for alternatives. One that suggests itself is to debounce emitting new_bookdisplay_data. I don't think this has the same problem as debouncing current_changed.

Yes debouncing just the signal should be OK. The only consumers of the
signal in calibre are book details, book info popups and Quick View. I
think all of them should be fine with a debounce.

There could be third party plugins that might break but unlikely and if
so we can provide an alternate API they can switch to.

PR #2226, opened 2024-03-13
Add custom templates for cite action

duydl:
I would like to add an option to selection-bar to copy highlight information using a template. This way user could easily define a template and paste their highlights into a markdown for example with readily available links to the source. 

The behavior is as follow: 
calibre-custom-hl

### DIFF
--- a/src/pyj/read_book/prefs/selection.pyj
+++ b/src/pyj/read_book/prefs/selection.pyj
@@ -42,6 +42,11 @@ def restore_defaults():
             control.valueAsNumber = val
         else:
             control.value = val
+
+    for ta in container.querySelectorAll('textarea[name]'):
+        val = session_defaults()[ta.getAttribute('name')]
+        ta.value = val
+
     set_actions(True)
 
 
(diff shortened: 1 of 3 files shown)


duydl:
I understand it kind of overlaps with the citation action. If a new action is not welcomed, would it possible to extend the highlight action instead? Like an option to copy info of a highlight into clipboard when it is created. 

Thank you for your consideration.

kovidgoyal:
I am travelling for a few days will looka t this when I have a moment.

kovidgoyal:
I am OK with adding an option to specify a template for the existing cite action. This is implemented around line 1150 in selection-bar.pyj

duydl:
test

Would this behavior be acceptable? Cite would behave the same as before with unhighlighted text and use the template with highlighted text.

kovidgoyal:
Couple of comments:

1) Move the setting below Quick highlights as it is not a commonly used
setting

2) Might make more sense to have just a single template that applies
always. When the selected text is not highlighted {notes}, etc. can simply
return the empty string. The default value of the template should be the
current format.

Or if you really want the template specially for highlights, then create
two templates one for selected and one for highlighted.

duydl:
I adjusted the location to follow your first comment. For the second, I would like to use two templates. I also added the vars into session for restore_default.

PR #2232, opened 2024-03-16
POLITICO.EU recipe

unkn0w7n:


unkn0w7n:
`conversion_options = {'tags': 'EU, News, Policy'}`
I've seen this being used in other recipes but why is it that 'tags' don't change?

kovidgoyal:
tags are hardcoded by the news download system. Users can add extra tags
in the schedule news dialog if they want, but recipes cannot.

PR #2233, opened 2024-03-17
Small change to commit d84b1e2 to be a little safer.

cbhaley:
It really isn't clear if the image can be non-None but not have a bounding box, or if getbbox can throw any exceptions. Just to be safe, check.


cbhaley:
I added an unrelated commit. I wish there was a way to have 2 PRs open to avoid having unrelated commits in the same PR.

kovidgoyal:
FYI: I believe you need to create a different branch with the separate commits. Then you can have a PR one from each branch.

PR #2253, opened 2024-04-01
Ensure that links removed by the new book details context menu item …

cbhaley:
…update the tag browser.


cbhaley:
I considered making model.current_changed() do the tags_view.recount(). I didn't do it because I wasn't sure what performance impact the change would have, as current_changed() is used in situations where no data update is done.

kovidgoyal:
On Mon, Apr 01, 2024 at 09:17:04AM -0700, Charles Haley wrote:
> I considered making model.current_changed() do the tags_view.recount(). I didn't do it because I wasn't sure what performance impact the change would have, as current_changed() is used in situations where no data update is done.

Yes, that way be dragons :)

PR #2266, opened 2024-04-19
Enhancement #2062434: Add option to show/edit current column …

cbhaley:
…definition to the booklist header context menu

### DIFF
--- a/src/calibre/gui2/library/views.py
+++ b/src/calibre/gui2/library/views.py
@@ -569,9 +569,6 @@ class BooksView(QTableView):  # {{{
             view.apply_state(view.get_default_state())
         elif action == 'addcustcol':
             self.add_column_signal.emit()
-        elif action == 'viewcustcol':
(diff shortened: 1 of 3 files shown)


kovidgoyal:
Why have a separate view and edit action? A view is just an edit
followed by a cancel, and that context menu is quite long already.

cbhaley:
I added it because of how the existing code works. If you open the dialog, change nothing, and press OK then it is considered edited (requiring a restart), which isn't intuitive. Changing that is more work than I want to do.

I originally had a view/edit menu action with two submenus. I could do that, saving space in the top-level.

Or if you are OK with requiring a restart when OK was pressed when nothing was changed then I can just show "Edit".

kovidgoyal:
I am fine with requiring a restart when doing OK with nothing pressed.
This is not a common operation.

cbhaley:
Changes pushed.

PR #2272, opened 2024-04-29
economist - reduce image quality.

unkn0w7n:


unkn0w7n:
How would I delete the ptemp file i created in place of url, after we made use of it in preprocess_raw_html?
i tried os.remove(url) in preprocess but it didnt work

kovidgoyal:
They get deleted automatically once the download is completed.

PR #2290, opened 2024-05-24
use small viewport unit for height to account for mobile browser nav bars

radonmiser:


kovidgoyal:
Shouldn't this be dvh? Otherwise wont it leave blank spaces in fullscreen?

radonmiser:
I've test dvh and svh they both work.
my understanding is that dvh should be avoided unless we specifically need it
https://dev.to/frehner/css-vh-dvh-lvh-svh-and-vw-units-27k4

when entering fullscreen dvh, lvh and svh should have the same value. when UI elements are present svh is the height with nav bars taken into account.

kovidgoyal:
Those considerations don't really apply for us, since we actually want
the reader div to be the full available height, width, no?

And even in other cases, the problem with dvh is it can cause content
to jump when the browser UI is resized. But we dont actually size
content using these units, only containers for content, like full screen
panels, or modals and such like. And these should be resized based on
available size, IMO.

Are all these changes actually necessary? Start with only changing
the units in where its clear they are needed to fix the bug. Probably
only read_book/view.pyj. I *think* just that is enough to fix your
issue.

The remaining ones we can change in future on case by case basis if
needed, with testing of the UI elements it affects.

PR #2297, opened 2024-05-28
Add support for passing custom configuration to recipies

felagund:
This would bring the changes discussed here: https://bugs.launchpad.net/calibre/+bug/2065877

It also includes changes to the Economist recipe to use the change.

It works on my machine with

    ebook-convert recipes/economist.recipe economist.mobi  --test --recipe-config="date=2024-05-18"
    
However, while developing it, I noticed that only a couple of last issues are accessible nowadays (as of now, 2024-04-27 woks, the issue after does not, and then the later ones work - two hours ago 2024-04-27 did not work).

For some reason, while using the GUI, the config does not seem to be passed to the recipe. I have no idea why, so I acknowledge this is not ready but I think I need a hint here.

There are several more questionable points:

I had an idea that if one passes "help", one would get a message about the format. However, I am not sure how to implement this properly, so maybe it can be discarded. My idea would be that if "help" is passed, other messages from running the recipe are suppressed and just the help message is displayed, but I have no idea how to accomplish that.

I am not sure about where the parse_config function is best called. I now call it in initialization of BasicNewsRecipe, but it only accomplishes to throw an error in case the option is used and a recipe does not implement it. Also using a global variable for edition_date might not be the best idea but I did not want to rework the whole logic of using a previous date in the Economist recipe.

Any feedback is welcome!

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -200,6 +200,19 @@ class Economist(BasicNewsRecipe):
     delay = 1
 
     needs_subscription = False
+    
+    def __init__(self, *args, **kwargs):
+        BasicNewsRecipe.__init__(self, *args, **kwargs)
+        global edition_date
+        if self.recipe_config:
+            edition_date = self.parse_config(self.recipe_config)
+
+    def parse_config(self,config):
+        if config == 'help':
+            raise SystemExit('You can pass a past date in format "date=yyyy-mm-dd". Note that only several past issues can be downloaded at the moment.')
+       	else:
+            edition_date = config.split('=')[1]
+            return edition_date
 
     def get_browser(self, *args, **kwargs):
         # Needed to bypass cloudflare
--- a/src/calibre/ebooks/conversion/plugins/recipe_input.py
+++ b/src/calibre/ebooks/conversion/plugins/recipe_input.py
@@ -52,6 +52,8 @@ class RecipeInput(InputFormatPlugin):
             help=_('Do not download latest version of builtin recipes from the calibre server')),
         OptionRecommendation(name='lrf', recommended_value=False,
             help='Optimize fetching for subsequent conversion to LRF.'),
+        OptionRecommendation(name='recipe_config', recommended_value=None,
+            help='Pass arbitrary configuration data to a recipe.'),
         }
 
     def convert(self, recipe_or_file, opts, file_ext, log,
--- a/src/calibre/gui2/dialogs/scheduler.py
+++ b/src/calibre/gui2/dialogs/scheduler.py
@@ -359,7 +359,12 @@ class SchedulerDialog(QDialog):
         self.custom_tags = ct = QLineEdit(self)
         la.setBuddy(ct)
         g.addWidget(la), g.addWidget(ct, 1, 1)
-        t2.la2 = la = QLabel(_("&Keep at most:"))
+        t2.la2 = la = QLabel(_("&Extra configurationi:"))
+        la.setToolTip(_("Extra configuration to be passed to the recipe, if the recipe implements it."))
+        self.recipe_config = cc = QLineEdit(self)
+        la.setBuddy(cc)
+        g.addWidget(la), g.addWidget(cc, 2, 1)
+        t2.la3 = la = QLabel(_("&Keep at most:"))
         la.setToolTip(_("Maximum number of copies (issues) of this recipe to keep.  Set to 0 to keep all (disable)."))
         self.keep_issues = ki = QSpinBox(t2)
         tt.toggled['bool'].connect(self.keep_issues.setEnabled)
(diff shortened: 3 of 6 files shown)


kovidgoyal:
You need to add metadata to the recipe class describing the available options. A simple dict of option names to short text description and long text description should do the trick. As an attribute on the recipe class. Then, you need to store this information in recipes/collection.py as XML.

This XML has to be read by the GUI to build up the GUI controls. You will also need to read it directly from custom recipe classes in addition to the XML used for builtin recipes.

The CLI should read this too and options of the form --recipe-option-name with help text. To do that you will need to add a method to InputFormatPlugin to get this dict, implement it for the recipe plugin and use it in conversion/cli.py to add more options. 

Passing the options from the GUI to ebook-convert will require some care, we can discuss that after you implement this much.

felagund:
Ok, this is more complicated than I thought it would be, let's see how far I can get:-).

Just to make sure I get your idea correctly:

1) There would be no default options unless a recipe implements this, so for the current recipes, nothing would change.
2) On the GUI side, options would be generated based on the definition in the recipe in the advanced tab of Schedule News Download, so for each option, there would be a separare field for it. 
3) On the CLI side, how would the user learn about the available options? There would be a switch like --get-options in RECIPE INPUT OPTIONS that would print the options for a given recipe (so something like `ebook-convert economist.recipe --get-recipe-options` would print something like
> --recipe-option-date  
    Specifies a past date to download instead of the current one in format yyyy-mm-dd
--recipe-option-something 
Some other option does something else
etc.
4) Recipe authors would not need to implement any parse data method but would instead just declare something like 

`options = {date: str, some_other_option: int}`

and they would then be able to use it with `self.date` or `self.some_other_option` just like they now can use `self.username` etc.?

kovidgoyal:
On Wed, May 29, 2024 at 01:43:10AM -0700, Tomáš Hnyk wrote:
> Ok, this is more complicated than I thought it would be, let's see how far I can get:-).

It usually is :) There is a reason I have never bothered to implement
this myself.

>
> Just to make sure I get your idea correctly:
>
> 1) There would be no default options unless a recipe implements this, so for the current recipes, nothing would change.

Yes.

> 2) On the GUI side, options would be generated based on the definition in the recipe in the advanced tab of Schedule News Download, so for each option, there would be a separare field for it.

Yes.

> 3) On the CLI side, how would the user learn about the available options? There would be a switch like --get-options in RECIPE INPUT OPTIONS that would print the options for a given recipe (so something like `ebook-convert economist.recipe --get-recipe-options` would print something like

ebook-convert something.recipe something.epub --help will output those options automatically in the INPUT section if you
follow what I suggested doing in conversion/cli.py, i.e. add the options
to the option parser.

> 4) Recipe authors would not need to implement any parse data method but would instead just declare something like
>
> `options = {date: str, some_other_option: int}`

class MyRecipe:
    recipe_options = {
        'date': {
            'description': 'Some short description used in GUI for label',
            'help': 'Longer description used as tooltip in GUI and in
            CLI --help output',
            'default': '',  # can be omitted
        }
    }

These options will be available on the options object that is passed to
the contructor (__init__) of BasicNewsRecipe. You will need to add some code to
that constructor that sets up the options. Something like this:

self.resolved_recipe_options = {k: getattr(options, 'recipe_option_' + k.replace('-', '_')) for k in self.recipe_options}

This is assuming that in conversion/cli.py you have mapped the keys
in recipe_options to --recipe-option- + name.replace('_', '-')

Only thing remaining is passing the options from the GUI to the backend
when running the recipe. Unfortunately I dont remember enough of that
code to be able to help you off the top of my head. We can discuss that
in more detail after all this part is done.

kovidgoyal:
Note that after adding options to a builtin recipe, it will show up in the GUI only after the next calirbe release. Also I havent implemented/tested it for custom recipes yet.

unkn0w7n:
```
recipe_specific_options = {
    'edition_date': {
        'short': 'The issue date to download',
        'long': 'Specify a date in the format YYYY-mm-dd to download the issue corresponding to that date',
        'default': 'current',
    }
}
```
what is short here, is it any different from long date? 

```
      if self.recipe_specific_options:
            log('Recipe specific options:')
            for k, v in self.recipe_specific_options.items():
                log(' ', f'{k} = {v}')
```

In the recipe, I think I should specify both the above `recipe_specific_options` dict and parse it with the `if self.recipe_specific_options:`  for k, v .items and get date from user inputs?

kovidgoyal:
See the comments above recipe_specific_options in the BasicNewsRecipe
class for how to use it.

felagund:
Hello, thanks, I was intending to dedicate a wekeend to this in the fall but I imagine what would have taken me days took you hours, so thanks again!

kovidgoyal:
You are welcome. Took me about a half a day so wasnt too bad :)

PR #2303, opened 2024-06-10
Improve Windows Build Instructions and Fix Some Compatibility Issues

ConstrictM:
Was trying to build a Windows binary and had some difficulties. I figured out how to get it to build, but I think the instructions could be updated to make it easier for others. Done in conjunction with https://github.com/kovidgoyal/bypy/pull/12

Bump ZLIB to 1.3.1
- Version 1.3 is no longer available at the link in sources.json

Support WIX Version 5
- Virtual Symbol "WixUISupportPerUser" needs to be overridden in WIX V5 https://wixtoolset.org/docs/fivefour/

Improve Windows VM Setup Instructions
- Add Specific VS Packages Required for Building Deps
- Add instruction to include cl.exe, nmake.exe and vswhere.exe in path
- Add html5lib as a Python Package for Qt-WebEngine Build

Add --dont-notarize to Window Build Steps
- Without disabling notarize, the script still attempts to sign the build

### DIFF
--- a/bypy/sources.json
+++ b/bypy/sources.json
@@ -67,8 +67,8 @@
     {
         "name": "zlib",
         "unix": {
-            "filename": "zlib-1.3.tar.xz",
-            "hash": "sha256:8a9ba2898e1d0d774eca6ba5b4627a11e5588ba85c8851336eb38de4683050a7",
+            "filename": "zlib-1.3.1.tar.xz",
+            "hash": "sha256:38ef96b8dfe510d42707d9c781877914792541133e1870841463bfa73f883e32",
             "urls": ["https://zlib.net/{filename}"]
         }
     },
--- a/bypy/windows.conf
+++ b/bypy/windows.conf
@@ -1,7 +1,26 @@
 # Requires installation of Visual Studio 2022 Community Edition, .NET SDK, WiX Toolset, Git, Ruby (without devkit), NodeJS, Python and Perl
+#
+# Visual Studio 2022 Components:
+# .NET SDK
+# C++ ATL for latest vXXX build tools (x86 & x64)
+# C++ Clang Compiler for Windows
+# C++ CMake tools for Windows
+# C++/CLI support for vXXX build tools
+# Git for Windows (can also be installed from web)
+# MSBuild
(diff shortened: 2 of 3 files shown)


kovidgoyal:
The other changes are OK but why would it need dont-notarize. If you look at __main__.py in bypy/windows signing is conditional only on sign_installers not on notarize.

ConstrictM:
> The other changes are OK but why would it need dont-notarize. If you look at **main**.py in bypy/windows signing is conditional only on sign_installers not on notarize.

Turns out that was an issue on my end. I reverted the change.

ConstrictM:
> The other changes are OK but why would it need dont-notarize. If you look at **main**.py in bypy/windows signing is conditional only on sign_installers not on notarize.

Took me a bit to understand why it was trying to sign event though I was setting dont-sign.

The command is ignoring the options because its calling BuildInstallers instead of BuildInstaller. BuildInstallers doesn't check the options which is why `./setup.py win --dont-sign` doesn't work. There are a couple answers to this. I'm not sure which is preferred.

1. Change Win(BuildInstallers) to Win(BuildInstaller)
2. Add the options to BuildInstallers
    - build_single(self.OS, bitness, shutdown)
    - build_single(self.OS, bitness, shutdown,
        not opts.dont_sign, not opts.dont_notarize, 
        compression_level=opts.compression_level, dont_strip=opts.dont_strip)

```
class Win(BuildInstallers):
    OS = 'windows'
```

```
class BuildInstallers(BuildInstaller):

    OS = ''
    ALL_ARCHES = '64',

    def run(self, opts):
        for bitness in self.ALL_ARCHES:
            shutdown = bitness is self.ALL_ARCHES[-1] and not opts.dont_shutdown
            build_single(self.OS, bitness, shutdown)
```

```
class BuildInstaller(Command):

    OS = ''
    BITNESS = ''
    ...
    def run(self, opts):
        build_single(
            self.OS, self.BITNESS, not opts.dont_shutdown,
            not opts.dont_sign, not opts.dont_notarize,
            compression_level=opts.compression_level, dont_strip=opts.dont_strip
        )
```

PR #2309, opened 2024-06-16
Allow setting more local resource files from cli

wrvsrx:
Previous discussions: https://bugs.launchpad.net/calibre/+bug/2054284

I add some code to allow setting more local resource files from cli when bootstraping.

### DIFF
--- a/setup/commands.py
+++ b/setup/commands.py
@@ -6,7 +6,7 @@ __copyright__ = '2009, Kovid Goyal <kovid@kovidgoyal.net>'
 __docformat__ = 'restructuredtext en'
 
 __all__ = [
-    'pot', 'translations', 'get_translations', 'iso639', 'iso3166',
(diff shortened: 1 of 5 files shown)


kovidgoyal:
This is too disruptive. Dont change the order of the sub commands of the bootstrap command, that's almost certain to break something. Instead, add command line options to control the sourcing of those resources in the commands in which they are sourced.

wrvsrx:
Subcommands reordering is avoided now.

Explaination to previous reordering:

I added `iso_data` as a subcommand of `resources` before since adding it as a subcommand of both `iso639` and `iso3166` causes that `iso_data.add_options()` be called twice, which would cause confliction in argparse. So `resources ` should be called before `iso639` and `iso3166`.

Now I add a check to ensure only first call to `iso_data.add_options()` take effect, so I can add `iso_data` as a subcommand of both `iso639` and `iso3166`.

PR #2315, opened 2024-06-23
improvements for device.py:

kyxap:
- multiple author match support
- debug print
- NO_AUTHOR_MATCH tag for Author mismatch
- '&' > 'and' support in book title

Before:
before

After:
after
Note: Author `C` - with `NO_AUTHOR_MATCH` when Author changed in calibre after book was pushed to the device

Also tried to report this as a bug but failed :| https://bugs.launchpad.net/calibre/+bug/2070045

Windows 11
Kobo Clara BW

### DIFF
--- a/src/calibre/devices/usbms/driver.py
+++ b/src/calibre/devices/usbms/driver.py
@@ -24,11 +24,37 @@ from polyglot.builtins import itervalues, string_or_bytes
 
 
 def debug_print(*args, **kw):
+    '''
+    Prints debug information to the console if debugging is enabled.
+
+    This function prints a message prefixed with a timestamp showing the elapsed time
+    since the first call to this function. The message is printed only if debugging is enabled.
+
+    Parameters:
+    *args : tuple
+        Variable length argument list to be printed.
+    **kw : dict
+        Arbitrary keyword arguments to be passed to the `print` function.
+
+    Attributes:
+    base_time : float
+        The timestamp of the first call to this function. Stored as an attribute of the function.
+
+    Behavior:
+    - On the first call, initializes `base_time` to the current time using `time.monotonic()`.
+    - If `is_debugging()` returns True, prints the elapsed time since `base_time` along with the provided arguments.
+    '''
+
+    # Get the base_time attribute, initializing it on the first call
     base_time = getattr(debug_print, 'base_time', None)
     if base_time is None:
+        # Set base_time to the current monotonic time if it hasn't been set
         debug_print.base_time = base_time = time.monotonic()
+
+    # Check if debugging is enabled
     if is_debugging():
-        prints('DEBUG: %6.1f'%(time.monotonic()-base_time), *args, **kw)
+        # Print the elapsed time and the provided arguments if debugging is enabled
+        prints('DEBUG: %6.1f' % (time.monotonic() - base_time), *args, **kw)
 
 
 def safe_walk(top, topdown=True, onerror=None, followlinks=False, maxdepth=128):
--- a/src/calibre/gui2/device.py
+++ b/src/calibre/gui2/device.py
@@ -1859,8 +1859,6 @@ class DeviceMixin:  # {{{
 
         def clean_string(x):
             try:
-                # Replace '&' with 'and' kobo automatically doing this before adding data into db
-                x = x.replace('&', 'and')
                 # Convert to lowercase if x is not None or empty
                 x = x.lower() if x else ''
             except Exception:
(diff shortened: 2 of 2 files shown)


cbhaley:
Ahhh, I see why I couldn't replicate your problem. You are dealing with books sideloaded on the Kobo some way other than by calibre. This means the UUID is wrong/non-existent and the book metadata might have nothing to do with the metadata in calibre. The "Match book in library" option in the device view context menu exists for these cases.

@Kovid: I am nervous about the change to clean_string changing '&' to 'and'. It isn't clear how this change will affect matching on non-Kobo devices. I am also a bit nervous about using "NO_AUTHOR_MATCH" value to indicate that the book didn't match. A book having a value in that column means the book *did* match, specifying how the book matched. Finally, it isn't clear how this might affect "Match book in library" and "Show matched book in library"  in the device view context menu.

kovidgoyal:
Replacing & with and is definitely not correct. device.py cannot have device specific fixes. If you want to make such a fix, the place to do it is the kobo driver. And I agree NO_AUTHOR_MATCH is not appropriate. 

Matching if only one author matches but all authors dont match is probably OK. 

Finally, if you want a debug_print() function use the existing one from usbms/driver.py

kyxap:
> You are dealing with books sideloaded on the Kobo some way other than by calibre.

nope, all books were send to kobo from calibre, I though there a problem of how fast it pull the data from db vs checking root metadata (it saves matched info there, right?). Anyway... after another round of testing I've found that if I delete root metadata calibre going to create new metadata but seems some peace's of information will be missing and that why ending out with Author, tittle logic, lol... now it make more sense to me :D But if I send any new book without device disconnect it will add missing data and list will be looking like:
image
So seems to me another bug to me. 

> The "Match book in library" option in the device view context menu exists for these cases.

I don this `Match book in library` in UI or in src, do you mean `Read metadata from files on device`? If so it seems if I turn it off it does not do anything differently 

>  I am nervous about the change to clean_string changing '&' to 'and'.

Yeah... I actually was not 100% sure about it and just were was the issue. Going to remove. 

> Finally, if you want a debug_print() function use the existing one from usbms/driver.py

what function? I see there same `if debug print` use

cbhaley:
The standard kobo driver doesn't use metadata.calibre for anything. It builds the booklist from the db. New books must be scanned which happens after a disconnect. Existing books aren't rescanned; they must be deleted before being resent.

The match books function is in the context menu for the device view.

kovidgoyal:
The debug_print() function at line 26 of usbms/driver.py

kyxap:
> The debug_print() function at line 26 of usbms/driver.py

my bad, was looking in wrong place, updated

kyxap:
also is there documentation of use of `metadata.calibre` and `.metadata.calibre` (wireless metadata) ?

cbhaley:
As I said before, the wireless driver doesn't use any form of metadata.calibre. The remote device supplies the metadata. 

As for documentation, metadata.calibre is built from the book list supplied by the device driver. Various drivers build the booklist in different ways. It is the driver-built booklist that device.py users.

kovidgoyal:
On Sun, Jun 23, 2024 at 08:01:47AM -0700, kyxap wrote:
> also is there documentation of use of `metadata.calibre` and `.metadata.calibre` (wireless metadata) ?

No, these are calibre internal files, not to be used by anything other
than calibre.

cbhaley:
> > The "Match book in library" option in the device view context menu exists for these cases.
> 
> I dont see `Match book in library` in UI or in src, do you mean `Read metadata from files on device`? If so it seems if I turn it off it does not do anything differently

The device view context menu is defined here: https://github.com/kovidgoyal/calibre/blob/da197459ec1277221558f0e442e615b7f98b21fc/src/calibre/gui2/__init__.py#L374

The device view context menu is set up here: https://github.com/kovidgoyal/calibre/blob/da197459ec1277221558f0e442e615b7f98b21fc/src/calibre/gui2/init.py#L100

The plugin used in the above context men for the 'Match Books' action is here: https://github.com/kovidgoyal/calibre/blob/master/src/calibre/gui2/actions/match_books.py

It uses this dialog to do the work: https://github.com/kovidgoyal/calibre/blob/master/src/calibre/gui2/dialogs/match_books.py

Note that you must choose 'Update cached metadata on device' in the device view toolbar to cause the changes to be given to the actual device driver. The usual method is calling "sync_booklists()" in the driver, which stores the changes however it does. This call is done by a signal that (eventually) calls this method in device.py: https://github.com/kovidgoyal/calibre/blob/da197459ec1277221558f0e442e615b7f98b21fc/src/calibre/gui2/device.py#L1765

PR #2323, opened 2024-06-28
Create wsj_mag.png

unkn0w7n:


unkn0w7n:
We can get audio files from WSJ and The Economist. How hard is it going to be to create a template for an audio book so that we can write audio book recipes?

I think its mostly going to be a zip of folders inside folders, containing audio file for each article. TTS of epub to audio book could also be created.

kovidgoyal:
On Thu, Jun 27, 2024 at 11:37:39PM -0700, unkn0w7n wrote:
> We can get audio files from WSJ and The Economist. How hard is it going to be to create a template for an audio book so that we can write audio book recipes?
>
> I think its mostly going to be a zip of folders inside folders, containing audio file for each article. TTS of epub to audio book could also be created.

calibre doesn't really support audio books (although the viewer can play
SMIL audio files on some platforms, depending on codec availability).

The conversion pipeline definitely doesnt support audio in nay shape or
form. You would need to create SMIL audio markup in your recipe and then
add some code to the conversion to at least pass it through if it doesnt
already.

PR #2337, opened 2024-07-13
Update wsj.recipe

unkn0w7n:
idk if you understood my question in the other PR.. just maybe update this recipe, so i know how to use the recipe_specific_options.

I would want users to know the date/input format and also add that only the past six editions are available. 
Maybe also add something to ask them to check log for available past editions.
```
# Past 6 editions are available for download.
# For available past editions see log and set date to, for example, '20240513'.
past_edition = None
```
so at the end we update `past_edition = ` whatever we got from user input.

I just want to know what's possible here.

### DIFF
--- a/recipes/wsj.recipe
+++ b/recipes/wsj.recipe
@@ -13,7 +13,7 @@ past_edition = None
 
 def media_bucket(x):
     if x.get('type', '') == 'image':
-        if x.get('subtype', '') == 'graphic':
+        if x.get('subtype', '') == 'graphic' or 'images.wsj.net' not in x['manifest-url']:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Add something like:
```py
recipe_specific_options = {
'date': {
'short': 'The date of the edition to download (YYYY-DD-MM format)',
'long': 'Tell the recipe to download the edition corresponding to the specified date which must be in the format bla bla. Note that only about 6 past editions are available from the WSJ servers.'
}}
```

Then in parse_index() of your recipe you can access this value as

```py
edition = self.recipe_specific_options.get('date')
if edition:
    date = parse_date(edition)
    ....
```

PR #2349, opened 2024-07-20
hindu and bloomberg

unkn0w7n:


unkn0w7n:
i think in news.py we can add recipe specific options if self. oldest article in recipe
```
recipe_specific_options = {
        'days': {
            'short': 'Oldest article to download from this news source. In days ',
            'long': 'For example, 0.5, gives you articles for the past 12 hours'
        }
    }
```
in parse_index
```
d = self.recipe_specific_options.get('days')
        if d:
            self.oldest_article = d
```

kovidgoyal:
news.py cant have knowledge of recipe specific options since by
definition they are specific to the recipe. You can simply implement
__init__ in your recipe, call the super class one first then set
self.oldest_article based on self.recipe_specific_options.

Or do the same with parse_feeds instead of __init__.

PR #2358, opened 2024-07-23
Support for new (2024) tolino devices

beedaddy:
This should add support for tolino shine (5th Gen.), shine color and vision color, but has only been tested with the shine 5.

Please note that the new tolino firmware 5.x uses 0 and 1 instead of "false" and "true" for boolean values. For now, I adapted the queries only if db version is >=188 _and_ device is tolino.

There's only one cover image extension used, and no folder tree. I tweaked the code to match.


kovidgoyal:
Note I made some changes to your code to make things more DRY

PR #2365, opened 2024-07-28
Fix import of manga epubs

lidavidm:
Sorry for the drive-by PR. Let me know if I should file a bug first, and/or if I should figure out how to add unit tests.

Some manga epubs sold online have two issues:

- They contain multiple OPF entries in META-INF/container.xml.  But only one exists.  Solution: check if the entry actually exists, and ignore it otherwise.
- The cover is not actually tagged as a cover.  But since this is a manga, the "first spine entry" is already a JPG/PNG.  Instead of trying to render it (which for me gives a black rectangle), try to parse it as an image and return it directly if it is actually an image.

### DIFF
--- a/src/calibre/ebooks/metadata/epub.py
+++ b/src/calibre/ebooks/metadata/epub.py
@@ -17,13 +17,10 @@ from calibre.ebooks.metadata.opf import get_metadata as get_metadata_from_opf
 from calibre.ebooks.metadata.opf import set_metadata as set_metadata_opf
 from calibre.ebooks.metadata.opf2 import OPF
 from calibre.ptempfile import TemporaryDirectory
+from calibre.utils.imghdr import what as what_image_type
 from calibre.utils.localunzip import LocalZipFile
 from calibre.utils.xml_parse import safe_xml_fromstring
 from calibre.utils.zipfile import BadZipfile, ZipFile, safe_replace
-
-import PIL
-from PIL import Image as PILImage
-
 
 class EPubException(Exception):
     pass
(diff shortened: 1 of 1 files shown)


kovidgoyal:
1) OCFReader cant know about self.archive, that's an artifact of
OCFZipReader. Instead of passing self.archive add an exists() method
to OCFReader, have it work by calling self.open() and catching OSError
and returning false. Override it in OCFZipReader. Pass self.exists to
Container rather than self.archive

2) I am not keen to have PIL try to load an image just to deal with this
out-of-spec case. Instead use the calibre imghdr module. Also do these
epubs specify a mime type for the first spine item that is correct? If
so first check that.

lidavidm:
Thank you for the feedback, I will try to update this week.

Re (2), if you feel like it's very much out of spec, I'm OK with just dropping the workaround. From what I see there is no MIME type specified in the OPF. (Alternatively, we can move the detection into the render_html_svg_workaround as one of the cases?)

kovidgoyal:
Moving it to render_html will not make any material difference, and is
anyway incorrect since that function is explicitly for rendering HTML,
its upto its callers to pass in correct HTML.

Using imghdr.what() is fine as far as I am concerned, to mitigate the
performance impact.

lidavidm:
Sorry for the delay. I've updated the PR. Since I see get_metadata assumes raster cover data is JPEG format, I've limited the hack to working for that format only (it happens to work for the books I have on hand) instead of trying to convert the data.

PR #2316, opened 2024-06-23
amazon parser: store all product details

milahu:
i needed a generic amazon parser to get all product details
so these are now stored in `mi._details`

also fixed some other stuff, probably not all will be merged

<details>
<summary>
example use
</summary>

```py
#!/usr/bin/env python3

# based on /nix/store/q220vq57bgpijr7qbymhcv8b26jb861n-calibre-7.10.0/bin/.calibre-wrapped

import sys, os



# set calibre paths

calibre_source = os.path.dirname(__file__) + "/calibre"
calibre_prefix = "/nix/store/q220vq57bgpijr7qbymhcv8b26jb861n-calibre-7.10.0"

#path = os.environ.get('CALIBRE_PYTHON_PATH', calibre_prefix + '/lib/calibre')
path = os.environ.get('CALIBRE_PYTHON_PATH', calibre_source + "/src")
if path not in sys.path:
    sys.path.insert(0, path)

#sys.resources_location = os.environ.get('CALIBRE_RESOURCES_PATH', calibre_prefix + "/share/calibre")
sys.resources_location = os.environ.get('CALIBRE_RESOURCES_PATH', calibre_source + "/resources")

sys.extensions_location = os.environ.get('CALIBRE_EXTENSIONS_PATH', calibre_prefix + '/lib/calibre/calibre/plugins')

sys.executables_location = os.environ.get('CALIBRE_EXECUTABLES_PATH', calibre_prefix + '/bin')

sys.system_plugins_location = None



#from calibre.gui_launch import calibre
#sys.exit(calibre())

url = sys.argv[1]
cache_path = sys.argv[2]

import calibre.ebooks.metadata.sources.amazon

import calibre.utils.browser

import calibre.utils.logging

log = calibre.utils.logging.Log()

#import traceback

def log_exception(self, *args, **kwargs):
    limit = kwargs.pop('limit', None)
    print(*args, **kwargs)
    #print(traceback.format_exc(limit))
    raise

log.exception = log_exception

import queue
result_queue = queue.Queue()

# Get book details from amazons book page
worker = calibre.ebooks.metadata.sources.amazon.Worker(
    url,
    result_queue,
    browser=calibre.utils.browser.Browser(),
    log=log,
    relevance=None,
    domain=None,
    plugin=None,
    #timeout=20,
    #testing=False,
    #preparsed_root=None,
    #cover_url_processor=None,
    #filter_result=None
    cache_path=cache_path,
)

worker.run()
print("result_queue", result_queue)

[...]

PR #2370, opened 2024-08-02
fix: dateutil.parser._parser.ParserError: Unknown string format

milahu:
im still "abusing" `calibre.ebooks.metadata.sources.amazon`
to parse metadata from amazon, like in #2316

currently it fails to parse the german date string `26. Juli 2017`

im calling `calibre.ebooks.metadata.sources.amazon.Worker`
with `url = ""` so it cannot get amazon domain (de) from url

[dateparser](https://github.com/scrapinghub/dateparser) seems to "just work" to parse local date strings

### DIFF
--- a/src/calibre/utils/date.py
+++ b/src/calibre/utils/date.py
@@ -111,7 +111,11 @@ def parse_date(date_string, assume_utc=False, as_utc=True, default=None):
     if iso_pat().match(date_string) is not None:
         dt = parse(date_string, default=default)
     else:
-        dt = parse(date_string, default=default, dayfirst=parse_date_day_first)
+        #dt = parse(date_string, default=default, dayfirst=parse_date_day_first)
+        # fix: dateutil.parser._parser.ParserError: Unknown string format: 26. Juli 2017
+        # https://github.com/scrapinghub/dateparser
+        import dateparser
+        dt = dateparser.parse(date_string)
     if dt.tzinfo is None:
         dt = dt.replace(tzinfo=_utc_tz if assume_utc else _local_tz)
     return dt.astimezone(_utc_tz if as_utc else _local_tz)


eli-schwartz:
> [dateparser](https://github.com/scrapinghub/dateparser) seems to "just work" to parse local date strings

Adding a new dependency that calibre doesn't need, for a self-declared case of "abuse", seems unusual.

Is there anywhere in calibre itself that has this issue too?

milahu:
> currently it fails to parse the german date string `26. Juli 2017`

the audiobook url is
https://www.amazon.de/dp/B0748L6YSH

> Erscheinungsdatum: 26. Juli 2017

> Audible.de Erscheinungsdatum: 26 Juli 2017

(Erscheinungsdatum = publication date)

i expanded `calibre.ebooks.metadata.sources.amazon`
so it also can parse audiobook metadata
so probably this code path is not reachable with calibre

kovidgoyal:
Use self.delocalize_datestr for this. Definitely not adding another
dependency for this.

milahu:
`self.delocalize_datestr` is already called ...

<details>

```py
    def parse_detail_cells(self, mi, c1, c2):
        # ...
        elif name in self.publisher_names:
            pub = val.partition(';')[0].partition('(')[0].strip()
            if pub:
                mi.publisher = pub
            if "(" in val and ")" in val:
                date = val.rpartition('(')[-1].replace(')', '').strip()
                try:
                    from calibre.utils.date import parse_only_date
                    date = self.delocalize_datestr(date)
                    # FIXME dateutil.parser._parser.ParserError: Unknown string format: 26. Juli 2017
                    mi.pubdate = parse_only_date(date, assume_utc=True)
                except:
                    self.log.exception('Failed to parse pubdate: %s' % val)
```

</details>

... but `self.delocalize_datestr` fails
because `self.months` depends on `self.domain`

```py
self.months = months.get(self.domain, {})
```

but in my case, `self.domain` is empty
the only input is `self.cache_path`

so the actual fix for my situation is
to parse `self.domain` from the html in `cache_path`

<details>

```py
class Worker(Thread):  # Get details {{{

    # ...

    def __init__(

        # ...

        if not self.domain and self.cache_path:
            # parse domain from javascript:
            # line 34: ue_sn = 'www.amazon.de',
            # line 284: 'Application': 'Retail:Prod:www.amazon.de',
            with open(self.cache_path) as f:
                html = f.read()
            match = re.search(r"ue_sn = 'www\.amazon\.([a-z.]+)',", html)
            if match:
                self.domain = match.group(1)
            del html

        # ...

        self.months = months.get(self.domain, {})
```

</details>

---

the general question is: how generic should calibre's amazon parser be?
(currently this is the best amazon parser i know)

another example: i need ISBN-10 but the parser only returns ISBN-13
(because calibre only needs ISBN-13...?)

kovidgoyal:
On Sat, Aug 03, 2024 at 02:00:55AM -0700, milahu wrote:
> the general question is: how generic should calibre's amazon parser be?
> (currently this is the best amazon parser i know)

As generic as it needs to be for calibre. I am not committing to
maintain things that calibre does not need.

>
> another example: i need ISBN-10 but the parser only returns ISBN-13
> (because calibre only needs ISBN-13...?)

You can extract ISBN10 from ISBN13 if a book has. If a book is assigned
an ISBN 10 then its ISBN13 is auto derived from it typically.

milahu:
> You can extract ISBN10 from ISBN13

ah, thanks

```py
# based on https://github.com/xlcnd/isbnlib/raw/dev/isbnlib/_core.py

def to_isbn10(isbn13):
    """Transform isbn-13 to isbn-10."""
    if not isbn13:
        return
    isbn10 = isbn13[3:]
    #check = check_digit10(isbn10[:-1])
    import calibre.ebooks.metadata
    check = calibre.ebooks.metadata.check_digit_for_isbn10(isbn10[:-1])
    # Change check digit
    return isbn10[:-1] + check if check else ''

# ...

product_id = (
    mi.identifiers.get("amazon") or
    to_isbn10(mi.identifiers.get("isbn")) or
    "no_product_id"
)
```

PR #2378, opened 2024-08-08
use ebook title (name) as the email attachment name

abbypan:
seems that kindle backend doesn't identify the non-utf8/non-ascii attachement file name.

we can use the utf-8 encoding "name" directly,  it can be identified by kindle backend, and more readable than the random english characters.

### DIFF
--- a/src/calibre/gui2/email.py
+++ b/src/calibre/gui2/email.py
@@ -154,7 +154,7 @@ def send_mails(jobnames, callback, attachments, to_s, subjects,
     for name, attachment, to, subject, text, aname in zip(jobnames,
             attachments, to_s, subjects, texts, attachment_names):
         description = _('Email %(name)s to %(to)s') % dict(name=name, to=to)
-        if isinstance(to, str) and (is_for_kindle(to) or '@pbsync.com' in to):
+#        if isinstance(to, str) and (is_for_kindle(to) or '@pbsync.com' in to):
             # The PocketBook service is a total joke. It cant handle
             # non-ascii, filenames that are long enough to be split up, commas, and
             # the good lord alone knows what else. So use a random filename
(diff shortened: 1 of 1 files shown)


kovidgoyal:
ascii letters are used because amazon's email service fails with some
non ascii filenames.

abbypan:
I think the problem is on the email's MIME filename part.

Amazon's email server will success if the filename is utf-8 encoding, even when it is non-ascii.

MIME example: 
`Content-Disposition: attachment;
 filename*=utf-8''%E6%B5%8B%E8%AF%95%20-%20%E6%9C%AA%E7%9F%A5.epub`

kovidgoyal:
As far as I know, it *does not succeed* with utf-8. It could be Amazon
finally fixed their servers, but given this is Amazon, I highly doubt
it.

If you think it does then please post the results of trying it with
attachment filenames that are Cyrillic, Chinese, Vietnamese and
Hindi. If all those are delivered correctly to the Kindle then I will
consider not ascii-izing filenames for kindles.

abbypan:
I think the problem is because: amazon may only read the utf8/ascii encoding ebook filename from the MIME part, but not parse the ebook metadata from the attachment.

The test files of  Cyrillic, Chinese, Vietnamese and Hindi,  the screen shot of calibre and kindle  are attached.

calibre_2024-08-08_14-59-54
kindle_1912652436
[测试.txt](https://github.com/user-attachments/files/16539102/default.txt)
[आधुनिक मानक हिन्दी.txt](https://github.com/user-attachments/files/16539098/default.txt)
[АА̀А̂А̄ӒБВГҐДЂЃЕЀЕ̄Е̂ӐА̊А̃Ӓ̄ӔӘӘ́Ә̃ӚВ̌ԜГ̑Г̇Г̣Г̌Г̂.txt](https://github.com/user-attachments/files/16539099/default.txt)
[tiếng Việt.txt](https://github.com/user-attachments/files/16539100/ti.ng.Vi.t.txt)

PR #2379, opened 2024-08-11
Books from the Tolino Cloud are EPUBs (not KEPUBs).

beedaddy:
Add MimeType 'application/epub+zip' so that these books are also being listed.

### DIFF
--- a/src/calibre/devices/kobo/driver.py
+++ b/src/calibre/devices/kobo/driver.py
@@ -716,7 +716,7 @@ class KOBO(USBMS):
                 # for calibre's reference
                 path = self._main_prefix + path + '.kobo'
                 # print "Path: " + path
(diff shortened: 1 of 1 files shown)


kovidgoyal:
I dont follow your change. Non KEPUB mimetypes are handled in the else branch already, why are you changing handling of EPUB mimetypes to follow that for KEPUB files? That will almost certainly break on non Tolino devices. 

If the TOlino requires some special handling of paths then it should be done under the else branch with a check for if the device is a Tolino one.

beedaddy:
For the Tolinos, it's the same logic as for the Kobo, but the purchuased books are EPUB: If the ContentID doesn't start with "file://mnt/onboard", it is a book from the Cloud (in .kobo/kepub). This is why I put it there. I added a tolino device-check so that it leavs the behaviour for Kobos as is. Does it look better now? I could move it to the else branch but should perhaps do the same checks.

PR #2382, opened 2024-08-13
fix(bypy/macos/__main__): Use -Os

Artoria2e5:
The launchers were getting compiled with no optimization at all. If one runs `objdump --disassemble calibre-launcher.dylib --no-show-raw-insn`, the result is a lot of loads and stores from the stack. Probably not critical at all for performance, but ugly and silly nevertheless.

```as
calibre-launcher.dylib:
(__TEXT,__text) section
_run:
    2f30:       pushq   %rbp
    2f31:       movq    %rsp, %rbp
    2f34:       subq    $48, %rsp
    2f38:       movb    %cl, %al
    2f3a:       movq    16(%rbp), %rcx
    2f3e:       movq    %rdi, -8(%rbp)
    2f42:       movq    %rsi, -16(%rbp)
    2f46:       movq    %rdx, -24(%rbp)
    2f4a:       andb    $1, %al
    2f4c:       movb    %al, -25(%rbp)
    2f4f:       movl    %r8d, -32(%rbp)
    2f53:       movq    %r9, -40(%rbp)
    2f57:       movl    -32(%rbp), %eax
    2f5a:       movl    %eax, 380680(%rip)
    2f60:       movq    -40(%rbp), %rax
    2f64:       movq    %rax, 413469(%rip)
    2f6b:       movq    -8(%rbp), %rax
    2f6f:       movq    %rax, 413434(%rip)
    2f76:       movq    -16(%rbp), %rax
    2f7a:       movq    %rax, 413431(%rip)
```

### DIFF
--- a/bypy/macos/__main__.py
+++ b/bypy/macos/__main__.py
@@ -59,7 +59,7 @@ def compile_launcher_lib(contents_dir, gcc, base, pyver, inc_dir):
 
     dest = join(contents_dir, 'Frameworks', 'calibre-launcher.dylib')
     src = join(base, 'util.c')
-    cmd = [gcc] + ARCH_FLAGS + '-Wall -dynamiclib -std=gnu99'.split() + [src] + \
+    cmd = [gcc] + ARCH_FLAGS + CFLAGS + '-Wall -dynamiclib -std=gnu99'.split() + [src] + \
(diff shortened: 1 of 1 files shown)


Artoria2e5:
Checked the linux thing and it uses `-O2` there. I use `-Os` out of habit: AFAIK, Apple likes to use `-Os`.

PR #2393, opened 2024-08-23
Update Science Journal

unkn0w7n:
sometimes index_to_soup fails to fetch url in this recipe.
with the qt browser_type, I get this error
```urllib.error.URLError: <urlopen error NetworkError.ContentAccessDenied: Error transferring https://www.science.org/toc/science/current - server replied: >
Exception ignored in: <function Browser.__del__ at 0x0000014608E01E40>
  File "calibre\scraper\qt.py", line 263, in __del__
  File "calibre\scraper\qt.py", line 241, in shutdown
ImportError: sys.meta_path is None, Python is likely shutting down
```
But using read_url from scraper.simple, with [old recipe](https://www.mobileread.com/forums/showpost.php?p=4441265&postcount=5), seems to still works. I see that its been updated to use `from calibre.scraper.qt import WebEngineBrowser`

I tried using `browser_type = 'webengine'`
and got this error.
```File "<string>", line 81, in parse_index
AttributeError: 'NoneType' object has no attribute 'findAll'
Exception ignored in: <function Browser.__del__ at 0x0000019D4CDE1E40>
  File "calibre\scraper\qt.py", line 263, in __del__
  File "calibre\scraper\qt.py", line 241, in shutdown
ImportError: sys.meta_path is None, Python is likely shutting down```


kovidgoyal:
Works for me with both browser_types. In any case I have a pushed a fix for the importerror on shutdown.

PR #2395, opened 2024-08-23
Add an option to prevent a composite column having its value stored …

cbhaley:
…in an OPF. This is useful if the composite contains a large amount of data, for example images, and the value of the composite isn't needed for a client. The option is in the column definition, stored in "display" for the column.

The option removes one of the penalties of creating composites for the booklist that show notes when the user hovers over the column.

Risks: near zero. Nothing changes unless the option is changed from its default. If the option is changed then there is a risk that in some context the option will be ignored, resulting in the same behavior as before.


kovidgoyal:
Actually I unmerged it as it causes tests to fail and peering at the if condition more closely shouldnt it be `composite and not store_in_opf` rather than `not (composite and store_in_opf)`?

cbhaley:
You are right. I had managed to corrupt my test data so I didn't see that the values for all non-composite columns were being emptied. I made another PR.

PR #2398, opened 2024-08-25
Update Science Journals

unkn0w7n:
reduced image size, option to change img-res

also added economist search recipe.

### DIFF
--- a/recipes/economist_search.recipe
+++ b/recipes/economist_search.recipe
@@ -1,0 +1,291 @@
+#!/usr/bin/env  python
+# License: GPLv3 Copyright: 2008, Kovid Goyal <kovid at kovidgoyal.net>
+
+import json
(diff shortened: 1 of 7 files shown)


unkn0w7n:
does `br.addheaders` in `def get_browser` work for `browser_type = 'webengine'`?
I feel like its not working.

kovidgoyal:
It should as far as I know. You can add a test for it in
test_fetch_backend.py if you like to verify.

kovidgoyal:
In fact ther eis already a test for it. addheaders is set by the headers
paramater passed to __init__ which sets the th: 1 header which is sent
by both qt and webengine backends.

unkn0w7n:
hmm.. I'll check again.. maybe the issue is something else.

PR #2413, opened 2024-09-04
add language code

ilkermeliksitki:
Changes to be committed:
	modified:   computer_weekly.recipe


ilkermeliksitki:
@kovidgoyal may I ask why do some jobs fail? Or did you just cancel them because it was an straightforward commit?

kovidgoyal:
Because one of the dependencies calibre needs during the build is
currently offline.

PR #2427, opened 2024-09-13
tts: gracefully handle the lack of a speechd module

eli-schwartz:
We already gracefully handle the lack of speechd having the `speech-dispatcher` command utility available, and respond by disabling that backend. Go one step further, and disable it if the module is missing too.

It's not even the default engine. It seems fair for piper users to not require having this installed. The build tests will assert its presence either way so it's not like it will go unnoticed.


eli-schwartz:
After this patch, uninstalling speech-dispatcher on my system results in

> There are no available TTS engines. Install a TTS engine before trying to use Read Aloud, such as flite or speech-dispatcher

Which is actually a net win over the previous state, which was "detecting" a speechd configured with no engines, and then failing to spawn it with `Server refused to autospawn, stating this reason: b''`. :D

PR #2429, opened 2024-09-14
New function get_item_id_using_db() in cache.py

cbhaley:
1) Add a new function get_item_id_using_db() to cache.py to improve performance. This new function is 20 times faster than get_item_id() when tested on a large library.

2) Use the new function in the built-in template notes and link functions where the exact match criterion is satisfied.


kovidgoyal:
I have added a case_sensitive parameter to get_item_id which uses the
database for the query. Implementation is also moved to db/tables.py

Change your PR to use the new API for the formatter functions.

PR #2435, opened 2024-09-16
tests: refactor out speech-related Qt tests to distinct testcase

eli-schwartz:
TTS support can be gated on e.g. Gentoo's USE=speech, so we don't want to run the basic selfcheck of speech support if this is disabled. We do want to run the general Qt checks, though (and also do a selfcheck of speech support if the USE=speech is enabled!) -- this is much more conveniently done if there is a distinct testcase for it that can be passed into --exclude-test-name.

### DIFF
--- a/src/calibre/test_build.py
+++ b/src/calibre/test_build.py
@@ -319,10 +319,21 @@ class BuildTest(unittest.TestCase):
         conn.close()
 
     @unittest.skipIf('SKIP_QT_BUILD_TEST' in os.environ, 'Skipping Qt build test as it causes crashes in the macOS VM')
+    def test_qtspeech(self):
+        if is_sanitized:
+            raise unittest.SkipTest('Skipping Qt build test as sanitizer is enabled')
+
+        from qt.core import QMediaDevices, QTextToSpeech
+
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Cant do this because there is no guarantee that the tts test will run after the main Qt test, which means a qapplication may not have been created. And we cant create a qapplication in any other test as the qt test needs some part of the test run before a qapplication is constructed. 

Just use an env var to skip that part of the test. If you *really* want to make it a separate test then you would need to run it in a separate process. This can be done but is a lot of overhead .

eli-schwartz:
Are you sure it needs a QApplication? I tested this by running just the qtspeech test in isolation and the test case passed for me. It also emitted a warning that I should create a QApplication but it wasn't clear whether that was a requirement or a "we assume if you touch this module you probably need one anyway".

kovidgoyal:
I'm not sure obviously, but, I dont want to spend time refactoring this
test again in the future if it does become necessary or if it is
necessary on a platform other than Linux. My policy is generally that if
the Qt docs dont explicitly say something doesnt require QApplication or
QCoreApplication then I assume it does.

eli-schwartz:
Fair enough. I agree that it sounds more hairy than I want to explore. The environment variable approach is done in #2436

PR #2436, opened 2024-09-17
tests: add variable to suppress TTS tests

eli-schwartz:
TTS support can be gated on e.g. Gentoo's USE=speech, so we don't want to run the basic selfcheck of speech support if this is disabled. We do want to run the general Qt checks, though (and also do a selfcheck of speech support if the USE=speech is enabled!) -- so add a sneaky environment variable that we can use to conditionally skip running those parts.

### DIFF
--- a/src/calibre/test_build.py
+++ b/src/calibre/test_build.py
@@ -123,6 +123,7 @@ class BuildTest(unittest.TestCase):
         from speechd.client import SSIPClient
         del SSIPClient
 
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Dont you want to skip the piper test as well?

eli-schwartz:
... right. I had originally added that to --exclude-test-name so when I retested it was being skipped anyway and then plain forgot about it.

PR #2437, opened 2024-09-17
Update nytfeeds.recipe

unkn0w7n:
does recursion limit apply when recursion is used on multiple links.
I mean should i limit the simultaneous downloads to 1 to avoid it?

### DIFF
--- a/recipes/nytfeeds.recipe
+++ b/recipes/nytfeeds.recipe
@@ -12,17 +12,13 @@ def extract_json(raw):
     return js['initialData']['data']['article']['sprinkledBody']['content']
 
 def parse_image(i):
-    if i['__typename'] == 'Image':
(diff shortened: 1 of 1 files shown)


unkn0w7n:
one article failed to fetch because of limit.. but individually in test it worked.

kovidgoyal:
recursion limit applies to following links on any page. It control the
depth to which links are followed.

kovidgoyal:
simultaneoud_downloads has no connection to recursion_limit

kovidgoyal:
failed with what error?

unkn0w7n:
> failed with what error?
it said something about recursion as the reason for article fetch fail. I should've copied the log.
Anyhow the recipe works fine.

PR #2439, opened 2024-09-18
Enhancement #2079784: Save the main window geometry when saving a cf…

cbhaley:
…layout.


kovidgoyal:
Apologies there was a stray rebase that caused github to not realise the commit is merged.

PR #2444, opened 2024-09-22
Update NYT

unkn0w7n:
today's paper recipe


unkn0w7n:
is nytimes.recipe different from nytimes_sub.recipe?

kovidgoyal:
On Sat, Sep 21, 2024 at 10:35:28PM -0700, unkn0wn wrote:
> is nytimes.recipe different from nytimes_sub.recipe?

These days I dont think so, it used to be in the past. Use diff to
check.

kovidgoyal:
Or actually, no, I think one parses todays page and the other uses
sections.

kovidgoyal:
Basically the two recipes files should be identical, except that is_web_edition is true in one and false in the other. If the nyt requires new parsing code, you should put it in site_parsers/nytimes.py then it can be re-used ina ll nyt recipes without needing to duplicate it in each recipe.

unkn0w7n:
oh.. i think i should just update your code.. its actually pretty similar.

PR #2445, opened 2024-09-22
Fix bug in new get_item_id() db lookup. Names in the authors table …

cbhaley:
…have commas changed to vertical bars.


kovidgoyal:
Shouldnt this also apply when len(item_names) > 1?

cbhaley:
Yes, it should, and I fixed it.

I also somehow got a spurious commit of your changes into this pull request. Sorry...

cbhaley:
Another fix: unserialize() isn't defined for some fields, for example tags.

PR #2446, opened 2024-09-22
Update nytimes.py

unkn0w7n:
i dont know how to test this.. but i've updated nytimes_sub.recipe and nytimes.py in site parsers.

I think this will work but can only be sure after testing.
For users, does the nytimes.py update remotely from github like recipes?


kovidgoyal:
Yes, it updates automatically. If you run it locally after bumping the
module_version in nytimes.py you can test it, as it should use your
local version.

PR #2449, opened 2024-09-24
Update NYT Book Review

unkn0w7n:


unkn0w7n:
I would like to make these changes to news.py
one is to use small font size for figcaption tags
and another add color to resovled links.. be able to recognize resolved_internal_links from normal href links that take you out of reader into a browser.  
Screenshot 2024-09-24 175950
Screenshot 2024-09-24 180018

kovidgoyal:
I dont think changing the default appearance of HTML tags is a good
idea, the recipe system should generally be neutral to such things and
use styles from the website wherever possible. That said, I have added a
commit to add a class to such tags based on their tag name so you can
easily style them via extra_css in a recipe where needed.

As for changing link color what happens if the hardcoded color blends into the
background and on b/w readers?

kovidgoyal:
On Tue, Sep 24, 2024 at 08:04:00AM -0700, unkn0wn wrote:
> > I have added a commit to add a class to such tags based on their tag name so you can easily style them via extra_css in a recipe where needed.
>
> f797b89ca8fb5e5fb32a93384190e8b693acd408 okay this is great.. I could just use `figcaption { font-size:small; }` but wouldn't it be better for those that use auto_cleanup or just use keep tags without any CSS.
> I think figcaption normally shows up as small text in html5 anyhow.

No, it isnt. Default user agent stylesheet has body font size for
figcaption.

>
> > As for changing link color what happens if the hardcoded color blends into the background and on b/w readers?
>
> Yes, the #202020 would be a problem when people use dark mode on their kindle. Idk what to suggest, but it would be great to differentiate resolved links from href.
>
> I actually use CSS `p a[href*="article_"] { color:#202020;}` personally.. to identify these links.
> I tried `border: solid; ` it looked ugly.. but it has a button type box.. I am thinking invert color for background but im not sure it would work on kindle.

You can use a unicode symbol such as middot or similar after the link
text to indicate its a local link. Probably search around and find a
better symbol than middot.

unkn0w7n:
https://www.w3schools.com/charsets

&raquo;   » I think this is okay.. very small.. not too distracting.

\u2B26   ⮞ 
\u2B26   ⬦

kovidgoyal:
raquo is used as a quotation mark in some languages. Other alternatives are ↓ and 🔗 and 📩 and 🖛 and 🗸

but we need to worry about whether common ereaders such as the kindle have fonts with these glyphs as well.

unkn0w7n:
https://www.w3schools.com/charsets/ref_html_entities_4.asp
i think we can only use something from here so that it shows up on any device.
 can we use a `&deg;` &deg; sign. Normally they use power `&sup2;`  &sup2; `&sup3;`  &sup3; for references.. we use degree for resolved links
even if websites have math.. i dont think they'd use a href with it.
 maybe color it red?

kovidgoyal:
Use • no need to color it and it is quite common (the bullet symbol) so
hopefully is present in most devices.

unkn0w7n:
tested with css content
with dot
image
Using - dash here
image

used ` here. Its not an apostrophe but looks good.
image

PR #2453, opened 2024-09-25
For column icons and grid view emblems, support icons for both light …

cbhaley:
…and dark themes.


kovidgoyal:
The suffix should be -for-dark-theme to match how dark theme icons work
with the calibre icon themes.

cbhaley:
Changes pushed.

PR #2458, opened 2024-09-29
Make the attributes id and has_cover 'official' calibre metadata …

cbhaley:
…attributes. This allows them to be copied by mi.deepcopy_metadata().

The problem was exposed by a kiwidude plugin "Search the Internet". After a deepcopy_metadata() a template using '{id}' failed with 'unknown field'.

I don't think this change will break anything. If you are worried then reject the PR. The problem has been there since forever, so it clearly is not a burning problem.


kovidgoyal:
I dont think this is a good idea, it's *very* widely used and predicting all the effects of the change is very difficult. There is already db_id for the book_id. Maybe the template language can be changed to fallback to that if id is missing? Also your commit seems to add only id not has_cover.

PR #2467, opened 2024-10-04
Update nautilus.recipe

unkn0w7n:
remove tags

### DIFF
--- a/recipes/nautilus.recipe
+++ b/recipes/nautilus.recipe
@@ -1,40 +1,42 @@
 #!/usr/bin/env  python
-'''
+"""
 nautil.us
-'''
+"""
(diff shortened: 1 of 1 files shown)


unkn0w7n:
image
can we get a aspect ratio option along with this. Instead of figuring out px, it would be better to include an option for ratio.
**0.75** is the aspect ratio that fits kindle devices nicely.

kovidgoyal:
Not going to put device specific options there, as for specifying an aspect ratio rather than a size, the problem is it will just trim a rectangle starting at top left which is not ususally what you want. There is no good way to convert an aspect ratio into a location AND size for a rectangle.

PR #2468, opened 2024-10-04
[add][recipe] Adding RSS source for reflex.cz

roolo:
- Printed magazine
- Magazine's web: https://www.reflex.cz


roolo:
Noticed there are no tests on a pull request. Shall try to setup something?

kovidgoyal:
I dont keep tests for network resources, they fail too frequently.

roolo:
@kovidgoyal : It could be something just like loading a recipe or something like that. Doesn't have to be exactly test of a function in the code.

kovidgoyal:
What would be the point of such a test?

roolo:
Something like the arguments in the recipe are valid, it contains all the required data/method calls. Automated (pre-)verification of a recipe 🤷🏾 If it fails you know you don't have to even click on the MR.

PR #2481, opened 2024-10-19
Fix error check in field_list_count(). Also speeds it up a bit.

cbhaley:


kovidgoyal:
I have merged the change to the formatter function but I am not
comfortable with changing has_key(). Your patch would change its
semantics from returning the list of keys defined on the individual
metadata object to that + all standard metadata keys. This is different
from the semantics of other methods in that object such as __iter__ and
might break some code that relies on the old semantics. You can instead
add a docstring documenting the semantics.

cbhaley:
Problem: proxy_metadata.has_key() does include the custom columns, so the Metadata.has_key() should as well.

I'll submit another PR that makes the behavior of the two versions the same.

cbhaley:
In fact, proxy_metadata.has_key() does what my original push does.

The two should do the same thing. Choices:
1) Metadata.has_key() same as proxy_metadata.has_key(). This is what I pushed before.
2) proxy_metadata.has_key() same as previous metadata.has_key(). Custom column keys are ignored.
3) Change both of them to check actual keys, as in
```
      def has_key(self, key):
        d = object.__getattribute__(self, '_data')
        return key in d or key in d['user_metadata']
```

I prefer the third option. The first is also OK but makes you uncomfortable. The second seems wrong to me. Which do you prefer?

NB: has_key() isn't used anywhere in base calibre.

cbhaley:
Looking at the code, option 1 is the best fix to make the two behave the same way.

cbhaley:
Submitted a PR where has_key() in Metadata and ProxyMetadata do the same thing.

No problem if you don't want to accept it, in which case the implementations won't change.

PR #2482, opened 2024-10-20
Here is an implementation of has_key() where both Metadata and …

cbhaley:
…ProxyMetadata do the same thing.


cbhaley:
As I said in the other PR, no problem if you don't want to accept it in which case the implementations won't change.

PR #2483, opened 2024-10-21
Use stable ordering when writing shell completions

keszybz:
Set ordering is unstable (uses the hash for ordering), so the output file would be different every time. Sort the sets into lists before writing. In various places there were unnecessary "conversions", e.g. sorted() always returns a list, so 'list(sorted(…))' is not useful.

In one place output is fixed to say: "Choices are: doc, ereader, ztxt" instead of "['doc', 'ereader', 'ztxt']".

For module constants, just create the object directly. The code is easier to read/edit when the list is constructed directly with items that are ordered in the same way that is shown in various outputs (instead of being sorted dynamically).

There is some more syntax confusion in src/calibre/linux.py. For example, '"%s" % (x)' is the same as '"%s" % x', because just putting something in parentheses does not create a tuple. It'd be nice to clean up this whole file to use f-strings and not the error-prone %-formatting, but that's a separate issue.

In a test build of the Fedora package, the generated file is now stable.

### DIFF
--- a/src/calibre/ebooks/conversion/plugins/pdb_output.py
+++ b/src/calibre/ebooks/conversion/plugins/pdb_output.py
@@ -14,13 +14,13 @@ class PDBOutput(OutputFormatPlugin):
     author = 'John Schember'
     file_type = 'pdb'
     commit_name = 'pdb_output'
-    ui_data = {'formats': tuple(ALL_FORMAT_WRITERS)}
+    ui_data = {'formats': ALL_FORMAT_WRITERS}
 
     options = {
         OptionRecommendation(name='format', recommended_value='doc',
             level=OptionRecommendation.LOW,
-            short_switch='f', choices=list(ALL_FORMAT_WRITERS),
-            help=(_('Format to use inside the PDB container. Choices are:') + ' %s' % sorted(ALL_FORMAT_WRITERS))),
+            short_switch='f', choices=ALL_FORMAT_WRITERS,
+            help=(_('Format to use inside the PDB container. Choices are:') + ', '.join(ALL_FORMAT_WRITERS))),
         OptionRecommendation(name='pdb_output_encoding', recommended_value='cp1252',
             level=OptionRecommendation.LOW,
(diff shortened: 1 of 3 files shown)


src/calibre/ebooks/conversion/plugins/pdb_output.py
```diff
@@ -14,13 +14,13 @@ class PDBOutput(OutputFormatPlugin):
     author = 'John Schember'
     file_type = 'pdb'
     commit_name = 'pdb_output'
-    ui_data = {'formats': tuple(ALL_FORMAT_WRITERS)}
+    ui_data = {'formats': ALL_FORMAT_WRITERS}
```
kovidgoyal, line 17: This is converted into a tuple for a reason, sets are not JSON serializable

src/calibre/ebooks/conversion/plugins/pdb_output.py
```diff
@@ -14,13 +14,13 @@ class PDBOutput(OutputFormatPlugin):
...
 
     options = {
         OptionRecommendation(name='format', recommended_value='doc',
             level=OptionRecommendation.LOW,
-            short_switch='f', choices=list(ALL_FORMAT_WRITERS),
-            help=(_('Format to use inside the PDB container. Choices are:') + ' %s' % sorted(ALL_FORMAT_WRITERS))),
+            short_switch='f', choices=ALL_FORMAT_WRITERS,
+            help=(_('Format to use inside the PDB container. Choices are:') + ', '.join(ALL_FORMAT_WRITERS))),
         OptionRecommendation(name='pdb_output_encoding', recommended_value='cp1252',
```
kovidgoyal, line 24: You have removed the space after the colon

src/calibre/ebooks/conversion/plugins/pdb_output.py
kovidgoyal: Although looking further down your PR I see you changed ALL_FORMAT_WRITERS to a tuple

PR #2486, opened 2024-10-21
Some independent commits

cbhaley:
1) New template operator 'field_inlist'. It operates directly on data in a multi-valued field instead of converting the field to a string then splitting it.
2) Add a count column to manage authors.
3) Change the name of format_date_field() to field_format_date()


cbhaley:
Kovid, there are 2 naming conventions I could use for the field variants of functions and operators: field_whatever and whatever_field. The first lets one find the field functions more easily because they group together in the documentation. The second lets one more easily find variants of the same function, for example format_date() and format_date_field(). I've used the first (field_whatever), but I am thinking that the second would be better.

If I used the second then the new function and operator would be list_count_field(), inlist_field. This is an easy change to make now because these two haven't ever been released. The existing format_date_field() would remain unchanged.

Do you have an opinion?

kovidgoyal:
I prefer the suffix as well.

PR #2487, opened 2024-10-22
Changes to use a suffix instead of a prefix for 'field' functions and…

cbhaley:
… operators.


cbhaley:
Added a commit to improve error handling in the format_date_field() function

PR #2504, opened 2024-11-05
Viewer: add a delay/pause when scrolling through a boundary in Flow mode

zybexXL:
This change makes scrolling with the mouse wheel in Flow mode stop at a boundary/section, preventing abrupt section changes. This has been requested multiple times:
https://bugs.launchpad.net/calibre/+bug/1861630
https://bugs.launchpad.net/calibre/+bug/1853634
https://bugs.launchpad.net/calibre/+bug/1800248

**Implementation:**

A new preference 'scroll_mouse_boundary_delay' is added to specify the delay before crossing a boundary. It defaults to zero, which disables this feature. A value of 0.5 seconds works well.

When scrolling through a boundary, the page change request is ignored/cancelled at first, and only allowed if the user scrolls again after a delay. The net effect is that moving the mouse wheel multiple steps in one go will now stop at the boundary instead of unexpectedly moving to the next/prev chapter. To actually change the chapter, the user just has to scroll again after a short delay.

### DIFF
--- a/src/pyj/read_book/flow_mode.pyj
+++ b/src/pyj/read_book/flow_mode.pyj
@@ -87,12 +87,12 @@ def _check_for_scroll_end(func, obj, args, report):
         if last_change_spine_item_request.name is csi.name and now - last_change_spine_item_request.at < 2000:
             return False
         ## delay change on boundaries
-        if opts.scroll_stop_boundaries :
-            if last_change_spine_item_request.ignoreUntil is None :
-                last_change_spine_item_request.ignoreUntil = now + 500
-            if now < last_change_spine_item_request.ignoreUntil:
+        if opts.scroll_mouse_boundary_delay > 0 :
+            if last_change_spine_item_request.denied is None :
+                last_change_spine_item_request.denied = now
+            if now - last_change_spine_item_request.denied < opts.scroll_mouse_boundary_delay * 1000 :
                 return False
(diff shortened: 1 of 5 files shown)


kovidgoyal:
It should use a single preference that sets the time in seconds as floating point. Defaulting to zero which means no pause. Also your PR seems to be missing the code to create the actual option only to use it,

zybexXL:
The current code is using an existing option called "scroll_stop_boundaries".
I can add a new option just for this, but the existing option and new behaviour seem to fit well. Would you care to test it as is before I add the new option?

zybexXL:
I also tested code to ignore N scroll wheel stops (ie, you need to scroll N+1 stops to actually cross the boundary), but I liked the delay behavior better.

kovidgoyal:
You cant re-use that option it is documented as applying only to smooth
scrolling, you will need to create a new option named something like
"Seconds to pause before mouse wheel scrolls past internal file boundaries"

A N wheel event approach wont work as touchpads typically generate pixel
scrolls not discrete scroll events. Delay will work for both scrolling
modes.

zybexXL:
Should the default delay be zero (to preserve current behavior) or 0.5 ?

kovidgoyal:
zero

zybexXL:
Added the new Preference, defaulting to zero. Recommended value to use this feature is 0.5 to 1 second.

PR #2506, opened 2024-11-06
Enhancement 2086748] Efficient Column Positioning in Column Manager

cbhaley:
Added keyboard modifiers to the up/down buttons on the dialog. Click moves the row one place (as before). Shift-click moves the row 5 places. Ctrl-click moves the row 10 places. Ctrl-Shift-click moves the row to the top or bottom as appropriate.


cbhaley:
Added commit for Enhancement #2086801

PR #2507, opened 2024-11-07
Reduce the verbosity of format_date() documentation

un-pogaz:
Reduce the verbosity of format_date() documentation by adding new lines and make the various formatting codes available more readable.


kovidgoyal:
@cbhaley please review

cbhaley:
Interesting that this came up now. It is directly related to the email I sent you today about converting the doc strings in formatter functions to markdown and using that markdown in all the places where functions are documents. If we do this then this PR isn't needed.

I have no objection to accepting this PR, but as I said the changes will be replaced with what is in template_language.rst if we move forward with my proposal.

FWIW: I don't see any significant reduction in "verbosity". What I do see is addition of newlines.

kovidgoyal:
Let's shelve it for now, we can revisit after we settle on a final documentation approach.

un-pogaz:
It very difficult to read the different formatting codes, as they are all compacted into a single line in the small text area of the template formatter. I'm not fluent in english, so "verbosity" was the better word I have at the moment fot that. Maybe "readability"?

But if a documentation format overhaul of user function is planned, I agree that this commit is not relevant.

cbhaley:
Changes are made and stable now.

PR #2518, opened 2024-11-13
The documentation editing tool with changes to support it.

cbhaley:
It is an action so it can be opened easily, should a document writer want to use it. It is also available in the Stored templates / User functions dialog.

It is possible that no one other than me will use it. :)

### DIFF
--- a/src/calibre/gui2/dialogs/ff_doc_editor.py
+++ b/src/calibre/gui2/dialogs/ff_doc_editor.py
@@ -185,3 +185,15 @@ class FFDocEditor(Dialog):
     def functions_box_index_changed(self, idx):
         self.show_original_cb.setChecked(True)
         self.fill_in_top_row()
+
+if __name__ == '__main__':
+    from tempfile import TemporaryDirectory
+    from calibre.db.legacy import LibraryDatabase
+    from calibre.gui2 import Application
+
+    with TemporaryDirectory() as tdir:
+        app = Application([])
+        db = LibraryDatabase(tdir) # needed to load formatter_funcs
+        d = FFDocEditor(None)
+        d.exec()
+        del app
--- a/src/calibre/gui2/dialogs/ff_doc_editor.py
+++ b/src/calibre/gui2/dialogs/ff_doc_editor.py
@@ -1,0 +1,187 @@
+#!/usr/bin/env python
+
+
+__license__   = 'GPL v3'
+__copyright__ = '2010, Kovid Goyal <kovid@kovidgoyal.net>'
+__docformat__ = 'restructuredtext en'
+
+'''
+Created on 12 Nov 2024
+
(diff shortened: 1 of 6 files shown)


cbhaley:
1. How do I run it with calibre-debug -c? I assume there must be some sort of entry point somewhere.
2. What template dialog? The template tester isn't really a place for it because nothing is edited there. The Stored templates dialog uses it to edit the documentation for the user template/function being edited, which isn't the same thing.

cbhaley:
OK, I solved the calibre-debug -c by using if __name__ == '__main__'. The only problem is that it doesn't remember geometry, but that isn't a huge problem.

cbhaley:
FWIW: with this code:
```
if __name__ == '__main__':
    from tempfile import TemporaryDirectory
    from calibre.db.legacy import LibraryDatabase
    from calibre.gui2 import Application

    with TemporaryDirectory() as tdir:
        app = Application([])
        db = LibraryDatabase(tdir) # needed to load formatter_funcs
        d = FFDocEditor(None)
        d.exec()
        del app
```
I get this exception when exiting
```
    with TemporaryDirectory() as tdir:
  File "tempfile.py", line 904, in __exit__
  File "tempfile.py", line 908, in cleanup
  File "tempfile.py", line 890, in _rmtree
  File "shutil.py", line 759, in rmtree
  File "shutil.py", line 617, in _rmtree_unsafe
  File "shutil.py", line 622, in _rmtree_unsafe
  File "tempfile.py", line 881, in onerror
  File "tempfile.py", line 890, in _rmtree
  File "shutil.py", line 759, in rmtree
  File "shutil.py", line 603, in _rmtree_unsafe
  File "shutil.py", line 600, in _rmtree_unsafe
NotADirectoryError: [WinError 267] The directory name is invalid: 'C:\\Users\\Charles\\AppData\\Local\\Temp\\tmpj1_vj_j6\\.calnotes\\notes.db'
```
It leaves the temp directory undeleted.

cbhaley:
Do you object to a non-action keyboard shortcut defined in gui2.ui, similar to 'Minimize the calibre window' and 'focus book list'? This allows someone to easily get to the editor without cluttering up the list of actions. I'd make it discoverable via a comment in formatter_functions.py.

kovidgoyal:
On Wed, Nov 13, 2024 at 06:52:46AM -0800, Charles Haley wrote:
> Do you object to a non-action keyboard shortcut defined in gui2.ui, similar to 'Minimize the calibre window' and 'focus book list'? This allows someone to easily get to the editor without cluttering up the list of actions. I'd make it discoverable via a comment in formatter_functions.py.

That's fine as long as it either defaults to no shortcut or a very
obscure shortcut. Use gui.keyboard.register_shortcut() for it.

PR #2520, opened 2024-11-14
Bug #2088135: Enumerated dropdown breaks after using 'Clear' button

cbhaley:


src/calibre/gui2/custom_column_widgets.py
```diff
@@ -786,11 +786,9 @@ def setup_ui(self, parent):
         w = MultipleWidget(parent, only_manage_items=True, widget=QComboBox, name=self.col_metadata['name'])
         w.get_editor_button().clicked.connect(self.edit)
         w.setSizePolicy(QSizePolicy.Policy.Minimum, QSizePolicy.Policy.Fixed)
-        self.set_to_undefined = w.clear
```
BibhuPrasad1998: w = MultipleWidget(parent, only_manage_items=True, widget=QComboBox, name=self.col_metadata['name'])

PR #2524, opened 2024-11-17
Various small fix

un-pogaz:
Three various small fix

1. Add `&` operator and `$` prefix to their doc function
2. Remove space in some functions arguments
3. Fix various unquoted arguments


kovidgoyal:
@cbhaley please review

cbhaley:
These are all good changes with one exception. That exception: in the commit "add & operator and $ prefix doc" the addition in the second hunk, "The ``$`` prefix can be used instead of the function, as in ``$tags``." is incorrect. The function returns the count of items in a multi-valued field. $field returns the field itself.

un-pogaz:
Hum... crap. I miss the good function to that, I push the fix.

cbhaley:
All good now.

PR #2532, opened 2024-11-20
Several commits related to template function documenation

cbhaley:
* Added the summary tag.
* Used :ffsum:
* Internal links now work in HTML and summaries
* Many bug fixes


cbhaley:
I pushed a commit to make the "back" button in the HTML document viewer position the viewport to where it was when a URL was clicked.

PR #2546, opened 2024-11-27
Update economist_world_ahead.recipe

unkn0w7n:


unkn0w7n:
was waiting for graphql.. but found right headers for web pages.

PR #2556, opened 2024-12-03
Increase toolbar small icon size

cm-fy:


kovidgoyal:
why?

cm-fy:
Hi! Well, the setting has been 24px since version 0.44, now we have larger screens, I believe. I think 24px has become too small an icon, nowadays!

Alternately, I had suggested a similar finer adjustment as the one in the Editor toolbars, not sure if you saw this thread:
https://www.mobileread.com/forums/showthread.php?t=364513
(which is way beyond my commit skills)

kovidgoyal:
Well I am not going to change the size of that setting. If you really
need a different size you can add another option, say medium-small or
something similar.

PR #2562, opened 2024-12-05
Re commit 6b98fac: for some reason the \1 must be \\1 to get the "…

cbhaley:
…real" text.

Tested only with HTML output.

### DIFF
--- a/src/calibre/utils/formatter_functions.py
+++ b/src/calibre/utils/formatter_functions.py
@@ -1847,9 +1847,10 @@ followed by the first letter of the genre, i.e. the genre "Fiction" becomes
 {}''').format('''\
 [CODE]
 program:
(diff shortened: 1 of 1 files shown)


cbhaley:
Won't we have the same problem in finish_formatting()?
[CODE]
{series:re(([^\s])[^\s]+(\s|$),\1)}{series_index:0>2s| - | - }{title}
[/CODE]

kovidgoyal:
On Thu, Dec 05, 2024 at 09:07:47AM -0800, Charles Haley wrote:
> Won't we have the same problem in finish_formatting()?
> [CODE]
> {series:re(([^\s])[^\s]+(\s|$),\1)}{series_index:0>2s| - | - }{title}
> [/CODE]
>


We might, hasn't happened yet. list_join() caused issues with both french
and spanish.

PR #2563, opened 2024-12-07
Enhancement  #2091139: Reordering Columns in Custom Column Editor

cbhaley:


cbhaley:
Another unrelated commit added.

PR #2585, opened 2024-12-24
Allow import of KFX files from MTP Kindle devices

j-howell:
The new 2024 Kindle models use MTP for file access and no longer allow the use of "Download and Transfer via USB" to obtain books in AZW3 format from Amazon's website. This makes it difficult to import Kindle books into calibre, especially since Amazon has blocked most other methods for obtaining Kindle books.

These changes allow better display of metadata in Device view when an MTP Kindle is connected and allows the import of KFX format books and notebooks from the device. KFX is not the best format for use within calibre, but it is better than nothing.

### DIFF
--- a/src/calibre/devices/mtp/driver.py
+++ b/src/calibre/devices/mtp/driver.py
@@ -96,8 +96,6 @@ class MTP_DEVICE(BASE):
         if storage_id in ignored_folders:
             # Use the users ignored folders settings
             return '/'.join(lpath) in {icu_lower(x) for x in ignored_folders[storage_id]}
-        if self.is_kindle and lpath and lpath[-1].endswith('.sdr'):
-            return True
 
         # Implement the default ignore policy
 
@@ -107,8 +105,9 @@ class MTP_DEVICE(BASE):
             'pictures', 'ringtones', 'samsung', 'sony', 'htc', 'bluetooth', 'fonts',
             'games', 'lost.dir', 'video', 'whatsapp', 'image', 'com.zinio.mobile.android.reader'}:
             return True
-        if lpath[0].startswith('.') and lpath[0] != '.tolino':
+        if lpath[0].startswith('.') and lpath[0] not in {'.tolino', '.notebooks'}:
             # apparently the Tolino for some reason uses a hidden folder for its library, sigh.
+            # Kindle Scribe stores user notebooks in subfolders of '.notebooks'
             return True
         if lpath[0] == 'system' and not self.is_kindle:
             # on Kindles we need the system folder for the amazon cover bug workaround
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Does no longer ignoring the .sdr folder have other side effects? I dont
have an MTP kindle so I cant test for myself. In particular I am
concerned about whether it causes excessive file scanning slowing down
performance and if there are spurious files detected as ebooks.

j-howell:
I have two Kindles that use MTP, a 2022 Scribe and a 2024 Colorsoft. Scanning the SDR folders slows down the initial scan of the device significantly. But I do not see how to avoid that given that the metadata.kfx file needed to show the proper title and author for books is located there.

I have not seen any spurious books caused by SDR scanning. The code change specifically blocks KFX files located there (always located under an “assets” subfolder) from being treated as books.

However some extra entries do appear due to TXT and HTML files located in the /system folder on Scribe devices. That is not changed by this commit.

kovidgoyal:
Hmm MTP is already dog slow, so I am reluctant to make it slower. Can we filter out subfolders of sdr other than assets?

j-howell:
Yes, it is slow and this pull request makes it quite a bit slower. I will test a change to use same technique that you originally did to filter out any other subfolders, but I don't expect it to improve things very much.

I am fine with dropping this idea if you feel that the increased slowness outweighs the benefits. I can always run the patched code from source for my personal use.

kovidgoyal:
I guess the proper solution is really to refactor the mtp drivers to
allow querying some folders without scanning them but that's a lot of
work. See if filtering out the other subfolders makes a difference, if
not probably pass for now.

j-howell:
I has tested the patch previously with only a small number of books present on my MTP Kindle devices. I have now performed a new test with 252 books loaded onto a Kindle Colorsoft on a Windows 11 PC. I looked at both the time needed to load filesystem metadata and for the job to get the list of books on the device. I ran the test several times and averaged the results.

Load filesystem metadata:
Unpatched original MTP driver: 2.7 sec
Patched driver with this commit: 24 sec
Patched driver also skipping SDR contents other than 'assets': 24 sec

Get the list of books on the device:
Unpatched original MTP driver: 45 sec
Patched driver with this commit: 40 sec
Patched driver also skipping SDR contents other than 'assets': 40 sec

So reading the filesystem is about ten times slower if the SDR content is included. Oddly it is slightly faster to obtain the book metadata once the filesystem has been read, possibly because it mainly relies on the small metadata.kfx files located in the SDR folder. Skipping SDR content other than 'assets' had no noticeable effect, likely because there is typically very little content of that type.

I did find the almost 30 second delay between the time the Kindle was attached via USB and the reaction to that in the calibre UI to be annoying so I understand if you think that this proposed change does more harm than good.

kovidgoyal:
Yeah, I think a 10x performance drop is not acceptable for this usecase. I guess I will need to add code to the mtp driver's native layers to be able to get a file given a path and a parent folder, without going through the filesystem cache. That should be enough for the scanning use case. For the adding use case I guess we also need a way to get a folder given a path and list the files in the folder. Closing this as in the present form it isnt mergeable.

j-howell:
Makes sense to me.

kovidgoyal:
I have added the native code. After the next preview or real release you can use the self.get_file_by_name() and self.list_folder_by_name() methods to download files and list folders using names relative to a parent folder from filesystem_cache.

j-howell:
Thank you for that. I will generate a new pull request once it is available.

PR #2587, opened 2024-12-27
add support for downloading older LWN articles

smipi1:
This adds support to download older LWN articles by the article identifier


kovidgoyal:
The way to do this is to use the facility for passing settings into the
recipe, see the recipe_specific_options field in the BasicNewsRecipe API
docs.

smipi1:
Hi @kovidgoyal ,

Thanks for the feedback. I should have dug a bit deeper for better examples. Is this better?

PR #2590, opened 2024-12-29
The fix for https://bugs.launchpad.net/calibre/+bug/2092643 isn't …

cbhaley:
…enough. Sometimes Qt opens the "wrong" editor on a column with a dialog for an editor, which remains open because closeEditor doesn't close them properly. You can see this by pressing Tab twice rapidly, which can nullify the effects of the singleShot timer.

This fix prevents these "wrong" dialogs from opening. It also fixes the original issue in the bug because the currentIndex() isn't incorrectly changed.

### DIFF
--- a/src/calibre/gui2/library/delegates.py
+++ b/src/calibre/gui2/library/delegates.py
@@ -218,12 +218,15 @@ class StyledItemDelegate(QStyledItemDelegate):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.table_widget = args[0]
+        # Set this to True here. It is up the the subclasses to set it to False if needed.
+        self.is_editable_with_tab = True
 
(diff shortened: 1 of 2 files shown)


cbhaley:
I pushed a more elegant implementation of the changes to delegates.py, using inheritance instead of adding the function in all the createEditor() instances.

kovidgoyal:
With this change tabbing past a column that does not have an editor such
as the size column causes subsequent columns to also not have editors.

cbhaley:
New commit pushed.

This problem is a nightmare. If this push isn't acceptable then as far as I can see we must accept either the random dialogs being opened or the loss of "editing" when tabbing over a readonly field. Sigh...

kovidgoyal:
Looks good to me, well done!

PR #2591, opened 2024-12-31
Fix invalid escape sequence error

ganeshhalthota:
I observed this invalid sequence error is visible during installation of calibre on Ubuntu 22.04. It appears to be visible for other users as well as seen in [invalid escape sequence when installing package](https://forums.linuxmint.com/viewtopic.php?t=425483&sid=a36fe93170b18bc7493173dc567acfd5).

### DIFF
--- a/src/templite/__init__.py
+++ b/src/templite/__init__.py
@@ -32,7 +32,7 @@ from polyglot.builtins import unicode_type
 
 
 class Templite:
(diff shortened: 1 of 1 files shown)


ganeshhalthota:
This resolves the bug - https://bugs.launchpad.net/calibre/+bug/2092772

kovidgoyal:
The second backslash is needed

PR #2592, opened 2024-12-31
Automatically inserting page-progression-direction="rtl" in rtl languages ​​(Arabic and Hebrew)

zevisvei:
I would also like to add that the text direction will automatically be right to left, but I didn't find where exactly this is defined in the code

Another thing I would like to do - when converting a pdf to text in rtl languages, the text is completely reversed, meaning that the order of the letters is also left to right, but I would need some guidance on where this is defined in the code

Thank you very much and Happy New Year

### DIFF
--- a/src/calibre/ebooks/conversion/plugins/epub_output.py
+++ b/src/calibre/ebooks/conversion/plugins/epub_output.py
@@ -197,6 +197,14 @@ class EPUBOutput(OutputFormatPlugin):
             opts.mobi_passthrough = False
             opts.no_inline_toc = False
             TOCAdder(oeb, opts, replace_previous_inline_toc=True, ignore_existing_toc=True)
+        
+        try:
+            lang = self.oeb.metadata.language[0]
+            lang = lang.value
+            if lang[:2].lower() in {'he', 'ar'} and self.oeb.spine.page_progression_direction is None:
+               self.oeb.spine.page_progression_direction = "rtl"
+        except IndexError:
+            pass
 
         if self.opts.epub_flatten:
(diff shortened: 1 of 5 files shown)


kovidgoyal:
You cant do this in opf2/oeb classes are they are used by all formats not just EPUB. Instead you do it in the plugins/epub_output.py file. There check if page progression direction is unset and set it based on book language instead.

For PDF output you would need to modify the PDF output plugin to set the dir attribute based on detected language for each parapgraph/span/whatever.

zevisvei:
Is [this](https://github.com/kovidgoyal/calibre/pull/2592/files/d5d71259a9b71aa83f984272e61f27843d4a34f9..5f33d46c1f0b6033ae0fa599b2f82ae5b69624a1#diff-2f0fe63867973a64b1c6f2a0f0beb7cdc59022fe7974e215119b912ed77488be) good?
If I put it in the opf2/oeb classes, could it cause problems?, I'm asking because the definition of the scroll direction is common to several formats, so it might be worth putting it  within these classes.
Regarding the pdf, the problem is that when the input is pdf, the text itself is reversed, meaning that instead of `abcd` it appears as `dcba`.

kovidgoyal:
Yes that should be fine. I dont want it done automatically in opf2/oeb as these are used in a *lot* of places and doing it automatically could break something somewhere. 

As for your PDF issue do you mean the text output by pdftohtml (which calibre uses) is reversed? If so then you should report it to the pdftohtml project (aka poppler).

zevisvei:
I removed the changes from metadata and oeb and added the scroll direction to azw3 and mobi as well.

PR #2595, opened 2025-01-03
Make backtabbing (Shift-Tab) during editing ignore keyboard modifers …

cbhaley:
…so dialogs won't be opened.

BTW: thanks for the kind words on my last PR. :)


kovidgoyal:
You saved me a major headache diving into Qt internals, so well deserved
:)

PR #2599, opened 2025-01-09
All GUI options stuff:

cbhaley:
1. Add the option to show the "GUI actions" button in the status bar. The code to add the Preferences button is commented out.
2. Add a shortcut to All GUI Actions: Ctrl+Alt+Shift+G
3. Fix some existing problems in the All GUI Actions implementation.
4. Add the infrastructure for status line buttons (all in init.py)
5. Make using the shortcut for All GUI Actions place the menu over the button in the status bar, if it exists and is the only button/menu item. (actions.__init__.py)


cbhaley:
Pushed an improvement to menu partitioning in "all actions"

PR #2601, opened 2025-01-10
Add an option to adjust the size of the link and note icons in book …

cbhaley:
…details.

In the tag browser and booklist the icons resize automatically with the font. That doesn't happen in book details because there isn't any notion of a "row". With this option the user can make the size as desired. For more background see the first few posts of https://www.mobileread.com/forums/showthread.php?t=365537

### DIFF
--- a/src/calibre/ebooks/metadata/book/render.py
+++ b/src/calibre/ebooks/metadata/book/render.py
@@ -107,13 +107,15 @@ def mi_to_html(
         mi,
         field_list=None, default_author_link=None, use_roman_numbers=True,
         rating_font='Liberation Serif', rtl=False, comments_heading_pos='hide',
-        for_qt=False, vertical_fields=(), show_links=True, item_id_if_has_note=None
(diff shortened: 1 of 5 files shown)


kovidgoyal:
Rather than another setting why not scale it with application font size. The size could be derived by using something like size = QFontMetrics(self.font()).horizontalAdvance('m')

cbhaley:
I thought of that but ended up choosing the current approach because the current scaling isn't sufficient for people (like me) without sufficient fine motor control to right-click inside the icon. The icons in the tag browser and booklist don't have their own context menus so you can right-click anywhere. The icons in book details do have a context menu so you must "hit" them. The "fix" is to allow the user to increase the size of the icon without increasing the size of the font, making them easier to hit.

kovidgoyal:
Then make it a factor defaulting to 1.0 rather than an absolute size.

cbhaley:
Scale factor changes pushed.

PR #2605, opened 2025-01-13
pep8 strict formating

un-pogaz:
Following the discussion on [mobileread](https://www.mobileread.com/forums/showthread.php?t=365121), here the pull request for a global and strict pep8 formating.

If you have any concern on a commit, I can easly edit them.

(forced push for fix some test fail)


kovidgoyal:
Please resolve the conflicts and rebase on master

un-pogaz:
done

kovidgoyal:
Also this is rather too much code for a human being to review. It means I cant be sure you didnt sneak in some malicious code in there (Not that I think you would just that I cant review it to check). SO maybe instead split it up into one PR that adds the changes to check.py and the strict.tmol file. Then let me know what command I need to run to generate the automatic changes. I will run and commit them myself. Then if there are some manual changes left in this PR I can review and merge them without all the noise form the automatic changes.

un-pogaz:
Oh, understandable. I will make this light PR. Do you prefer discute of how decompose and reproduct it here, on the light PR or on mobileread?

eli-schwartz:
What I usually do in such a case is include in the commit message, something like:

> produced by running:
> ```
> ruff check --select XXXXX --fix
> ```
> and committing the results

Given a per-commit reproducer command it is quite easy with only light bisection to verify the results are as expected.

un-pogaz:
The Ruff code/rules are include in each commit. Nota that it will not allway fully match with the reference commit because I slightly twearked the result to better match the calibre format, or even manualy add somme edit to extend the formating because Ruff don't catch them.
(this was especially visible for the blank-line because the system of region used in calibre really mixup the result)

Some commit have none comment because is was perform entierly manualy.

un-pogaz:
No there are not doable by `setup.py check --strict` and... it's a good idea.

It's will be a bit a tidious to do to separate the "machine auto-fix" from "manual edit", but if that can facilite the review, I start that tomorow.

kovidgoyal:
On Mon, Jan 13, 2025 at 11:41:41AM -0800, un_pogaz wrote:
> It's will be a bit a tidious to do to separate the "machine auto-fix" from "manual edit", but if that can facilite the review, I start that tomorow.

Yes, please do, thanks.

un-pogaz:
Okay, here a new version of the pull request, following the idea of separate the change into several commits depending of the type of change. The type of commit are:
- auto-fix: change was perfom with the `--fix-only` command of Ruff
- manual: error was fixed manualy, following the errors reported by Ruff
- extra-edit: additional edit, generaly to extend some thing undetected by Ruff

Inside the body of the commit, there is the rule code targeted by Ruff, plus aditional command if so. `!partial` keyword indicated rule used as rereference for this edit, but not fully included into `ruff-strict-pep8.toml` because a strict application will go againt "calibre standard".

Note: "add/remove blank-line" commit was marqued 'extra-edit' because is a complet mess of intricate auto-fix, manual and extra-edit, and separet them will be a even more horrible mess to review. For a programmatic review of this commit, I suggest to compare the files before/after line by line, minus empty lines, which should theoretically result at the same array since this commit only edit empty lines.

PR #2609, opened 2025-01-15
Add ability to tag browser specify icons for individual values, as is…

cbhaley:
… already done automatically for 'formats'.

This scheme could be used to replace the existing tag browser category icon selection. I didn't do that because the risk seemed high. The category icon dict is used in several places and in plugins. It also has meaning for search and user categories, where the new value icon stuff doesn't.

If the new facilities are not used then performance risk is near zero. Performance shouldn't be an issue if used because the icon dictionaries scale well. The exception might be templates. We have no control over the complexity or performance of user-written templates.

This PR fixes a few bugs I found while implementing the new feature.

### DIFF
--- a/src/calibre/gui2/__init__.py
+++ b/src/calibre/gui2/__init__.py
@@ -420,6 +420,7 @@ def create_defs():
     defs['bd_show_cover'] = True
     defs['bd_overlay_cover_size'] = False
     defs['tags_browser_category_icons'] = {}
+    defs['tags_browser_value_icons'] = {}
     defs['cover_browser_reflections'] = True
     defs['book_list_extra_row_spacing'] = 0
     defs['refresh_book_list_on_bulk_edit'] = True
--- a/src/calibre/gui2/dialogs/template_dialog.py
+++ b/src/calibre/gui2/dialogs/template_dialog.py
@@ -488,11 +488,16 @@ class TemplateDialog(QDialog, Ui_TemplateDialog):
                  icon_field_key=None, icon_rule_kind=None, doing_emblem=False,
                  text_is_placeholder=False, dialog_is_st_editor=False,
                  global_vars=None, all_functions=None, builtin_functions=None,
-                 python_context_object=None, dialog_number=None):
+                 python_context_object=None, dialog_number=None,
+                 formatter=SafeFormat, icon_dir='cc_icons'):
         # If dialog_number isn't None then we want separate non-modal windows
         # that don't stay on top of the main dialog. This lets Alt-Tab work to
         # switch between them. dialog_number must be set only by the template
         # tester, not the rules dialogs etc that depend on modality.
+
+        # doing_emblem is also used for tag browser value icon rules in order to
+        # show the icon selection widgets.
+
         if dialog_number is None:
             QDialog.__init__(self, parent, flags=Qt.WindowType.Dialog)
         else:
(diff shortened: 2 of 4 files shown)


cbhaley:
BTW: You might think of using a rule-based system like emblems and icon rules. I looked into that. The problem is that the rules must operate on tag browser nodes (item values) that are constructed at the moment using various options. A rules editor would need to construct the same node values, which wouldn't be straight forward.

It would be straight forward to use a rules display to visualize the info, but I'm not convinced this adds much value.

kovidgoyal:
Not sure I follow why the rules editor would need to construct node values? It can just have a free form text input field to specify match expressions, it doesnt need to actually evaluate nodes while editing rules. That said I dont really think this needs rules, right clicking to specify icons per item is fine IMO. Rules really shine when there are large, unfiltered data sets to match against.

The main question I have is about the Content server, what happens to the tag browser there?

cbhaley:
The issue with nodes is that their text changes depending on hierarchy and partitioning. For example a node might be 'AA' or 'AA.BB' depending on whether it is hierarchical. For rules the user would need to "know" the structure. The context menu avoids all of that. But as you said, this isn't the problem ...

For clarity, below when I say CS I mean the server side of the content server. When I say CSUI I mean the client side of the content server.

I don't know how the new CS and CSUI work other than to say there is a lot of generated javascript behind the scenes controlling what data is visible to the CSUI. It seems clear that the existing category icons are sent to the CSUI and are used in the CSUI tag browser (which is really the search interface, isn't it?). BTW: I note that the CSUI doesn't respect the tag browser's "don't subcategorized" setting. If I go to Prefs / L&F / Tag browser / Partitioning and collapsing then

1) Choose "by first letter"
2) Uncheck "Tags"

the CSUI still shows tags partitioned by first letter. In the CSUI you must go to the controls and enter "tags" into the "Categories not to partition" box. I also note that the CSUI Formats category doesn't use the mimetype-based icons while the tag browser does.

The new "item-value-based" icons could (I think) be sent and used in the same way as the category icons. The gpref 'tags_browser_value_icons' is a dict
`{category: {item_value: (icon_file_name, for_children)}`
This dict could be sent by the CS to the CSUI user interface when a library is opened, probably with the icon_file_names translated to icons.

The CSUI would compare the node item-value to the dict and use the icon if it matches. The "for_children" part might be more complicated. What the tag browser does is walk up the node tree checking if a node with an icon has "for_children" set. If it does then that node's icon is used. I expect that the CSUI could walk up the tree in a similar fashion but I don't know what the performance impact might be.

I don't think that the template option would move to the CSUI because it requires running the template language interpreter for nodes that don't have some other icon, which would happen in the CS because it has the underlying infrastructure. You can't run the template once because it can produce different icon names depending on its internal computation. It would require a network exchange per node between the CS and the CSUI, which would be very slow.

I think the choices are:
- Don't support per-node icons in either place, rejecting my changes.
- Support per-node icons in the tag browser but don't support per-value icons in the CSUI. This is what I submitted. I don't think this is unreasonable given the existing differences between the CS/CSUI and the tag browser.
- Support the dict where the icon to use is specified by the item value, while not supporting templates.  I would be very uncomfortable attempting this work.

cbhaley:
I'm looking at srv/metadata.py. It seems I need to add the icon file name to the dict returned by category_item_as_json(). Do you also want me to add the file names as a new key in icon_map()?

I'm also going to add some documentation to the manual.

cbhaley:
I was wrong. It seems that the info must go into the tree built by process_category_node().

kovidgoyal:
On Thu, Jan 16, 2025 at 02:00:45AM -0800, Charles Haley wrote:
> I was wrong. It seems that the info must go into the tree built by process_category_node().

Yes, correct.

kovidgoyal:
No I dont think filenames in icon_map are needed. icon_map() is used for custom category icons only.

PR #2611, opened 2025-01-16
Stuff for value icons:

cbhaley:
- Documentation
- bug fixes in icon template evaluation
- Better enabling/disabling of menu lines in the tag browser context menu


kovidgoyal:
The manage icon for value sub-menu is available even when clicking on
the grouping entry int he tag browser. Is that intended?

cbhaley:
Yes. It is there and on the category icon to permit setting defaults, However, I see now that I messed up the enabling of menu lines, which I will fix in another PR.

PR #2613, opened 2025-01-16
Escape ampersands in tag browser icon value strings.

cbhaley:


cbhaley:
Also added a commit fixing a regression in setting category icons.

PR #2615, opened 2025-01-17
Add a simple icon value rule browser.

cbhaley:
Use case: rules can be "lost" if the value disappears, for example a tag. It is also useful to see the existing rules to know which are hierarchical etc.

This dialog permits deleting rules (cleaning up). It doesn't permit adding rules.

### DIFF
--- a/src/calibre/gui2/dialogs/tag_browser_rules_display.py
+++ b/src/calibre/gui2/dialogs/tag_browser_rules_display.py
@@ -1,0 +1,180 @@
+'''
+Created on 17 Jan 2025
+
+@author: Charles Haley
+'''
+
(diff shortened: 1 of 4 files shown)


kovidgoyal:
Maybe move the action for it into Preferences->Look & feel->Tag browser
I dont think its something that one needs quick access too.

cbhaley:
I didn't put it in L&F because I think that would create lots of demand that it be an editor, not a viewer. I don't want to build an editor. However, you are correct in that it isn't something one needs to see on a frequent basis.

I need to think about this. You can close this PR. I'll resubmit later after thought about where it should go and what it can do.

PR #2616, opened 2025-01-17
Improve reset icon tag browser

un-pogaz:
Improve the reset to default icons for category and values in tag browser.

Fix bug when 'Restore {} category default icon' action reset all icons except template.
Fix bug when 'Use the category icon for the default value icon' actions reset also the category icon to default (instead of just the template).

### DIFF
--- a/src/calibre/gui2/tag_browser/model.py
+++ b/src/calibre/gui2/tag_browser/model.py
@@ -424,10 +424,10 @@ class TagsModel(QAbstractItemModel):  # {{{
         '''
         old_icon = self.prefs['tags_browser_category_icons'].get(old_key, None)
         if old_icon is not None:
-            old_path = os.path.join(config_dir, 'tb_icons', old_icon)
+            old_path = os.path.join(self.icon_config_dir, old_icon)
             _, ext = os.path.splitext(old_path)
             new_icon = new_key + ext
-            new_path = os.path.join(config_dir, 'tb_icons', new_icon)
+            new_path = os.path.join(self.icon_config_dir, new_icon)
             os.replace(old_path, new_path)
             self.set_custom_category_icon(new_key, new_icon)
             self.set_custom_category_icon(old_key, None)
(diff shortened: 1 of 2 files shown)


kovidgoyal:
@cbhaley please review

cbhaley:
The icon choice hierarchy is:
1) An icon for the value, if it exists.
2) An icon for a parent node, walking up the tree, if one exists.
3) An icon from a template, if a template exists and it returns a non-empty string.
4) The default category icon, which must exist.

This hierarchy means that restoring the category icon ('Restore {} category default icon') shouldn't remove all the icon rules for the category. It should do what it says, restore the category icon to the default. 

If we want a way to remove all value-specific rules for a category then a new menu action 'Remove all value icon rules for this category' (or some other wording) should be added that is enabled only when working with a category. This action should not remove a template rule because that specifies a default icon.

I'll add the hierarchy specified above above to the documentation.

cbhaley:
@Kovid: this looks fine. I'll check again once I can run it.

un-pogaz:
Okay, I added a specific action to reset all value icons.
Inded, it more coherent that all value icons action are in the same sub menu and that 'Restore {} category default icon' only affect the category itself and not their values.

cbhaley:
Questions:
1) should there be an "Are you sure" warning dialog on the reset all value icons action? Removing all value rules is a big operation with no undo.
2) Should that action be disabled when looking at a specific value / enabled when looking at a category?

I'm not sure for either of these.

un-pogaz:
1. The 'Restore {} category default icon' have no confirm dialog, so I haved follow. I think this will be the opportunity to added it too.
2. With a confirmation dialog, I thing we can keed enable it for all values.

cbhaley:
And for what its worth, I think the new menu action should be named 'Reset all value icons to the default icon'. There might be a template that supplies an icon for a value once the value rules are deleted.

un-pogaz:
And added the confirm dialog.

un-pogaz:
Move 'reset all value icons' into the values section. The section 'All values' wasn't very relevant with just one action.

Pull ok to review and merge.

cbhaley:
@Kovid: I'm fine with so the changes.

PR #2620, opened 2025-01-19
Two changes:

cbhaley:
1) add a new tab to Pref / L&F / TB to show the rules viewer. This includes new options to show/hide all category and value icons. They work in the GUI. I don't think they need to be used in the content server because the screen layout there is so different.
2) change the architecture of preferences to support subclasses as tabs.

Doing the first change, which we discussed earlier, provoked me doing the second. Editing look_feel.ui and look_feel.py are very difficult because they are so big. Keyboard shortcuts are all mixed together, as are variables. It is hard to tell what does what. I decided to provide a mechanism to separate tabs into separate .ui and class files. Most of the changes in this PR are to support that.

Testing: as well as doing the new icon browser tab as a separate class, I converted two other tabs (Cover View and TB Hierarchies) to be sure that it works.

To put a preferences tab/window into its own pane:
- The new class inherits from the new class ConfigTabWidget in preferences.__init__.
- Create a .ui file if desired.
- Implement the tab class. It inherits from ConfigWidgetBase so the interface methods must be implemented.
- Modify the "real" preference widget .ui file, in my case look_feel.ui, to include the new class. The class can be a main tab (e.g., cover browser preferences) or a sub-tab (tag browser hierarchy and icon rules browser).
  - If the class is for a main tab then the content margins for the containing layout should be 0.
  - If the class is a sub-tab then this isn't necessary.
- In the main tab (in this case look_feel.py) create an __init__ method if needed then call `self.register_child_tab()` for each of the tabs/sub-tabs implemented by classes. I couldn't find a way to avoid this manual step.

The interface methods in the subclasses are automatically called at the appropriate time.

### DIFF
--- a/src/calibre/gui2/__init__.py
+++ b/src/calibre/gui2/__init__.py
@@ -487,6 +487,8 @@ def create_defs():
     defs['light_palettes'] = {}
     defs['saved_layouts'] = {}
     defs['book_details_note_link_icon_width'] = 1.0
+    defs['tag_browser_show_category_icons'] = True
+    defs['tag_browser_show_value_icons'] = True
 
     def migrate_tweak(tweak_name, pref_name):
         # If the tweak has been changed then leave the tweak in the file so
--- a/src/calibre/gui2/dialogs/book_info.py
+++ b/src/calibre/gui2/dialogs/book_info.py
@@ -88,7 +88,7 @@ class Configure(Dialog):
         Dialog.__init__(self, _('Configure the Book details window'), 'book-details-popup-conf', parent)
 
     def setup_ui(self):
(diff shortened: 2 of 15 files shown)


cbhaley:
I found a way to automatically register the child tab class instances, simplifying setup.

un-pogaz:
A little suggestion: If that not raise any import error, move the `calibre/gui2/preferences/look_feel_displayed_fields.py` to `calibre/gui2/preferences/look_feel_tabs/__init.py`

cbhaley:
@un-pogaz: good idea. Done and pushed.

kovidgoyal:
Merged, but in the rules view please allow deleting rule with delete
key and add a delete button to the side, The context menu is not very
discoverable.

PR #2621, opened 2025-01-20
As requested, add buttons and the delete key to the icon rules browser.

cbhaley:


kovidgoyal:
Pressing delete does not enable the Apply button. And right clicking in
an empty part of the rules list gives the follwoing traceback:

  File "/home/kovid/work/calibre/src/calibre/gui2/preferences/look_feel_tabs/tb_icon_rules.py", line 103, in show_context_menu
    item = self.rules_table.item(clicked_item.row(), CATEGORY_COLUMN)
                                 ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'row'

kovidgoyal:
Have commited fixed for both those issues.

cbhaley:
Thanks.

PR #2623, opened 2025-01-21
Three things related to templates in the content server.

cbhaley:
- Fix composites that use database methods not working in the content server.
- Fix template searches using the database API not working in the content server.
- Update the manual.


cbhaley:
I don't think I changed the behavior of templates, icons, or composites in the GUI, but I might be wrong. It might be a good idea to make another preview at the end of the week.

kovidgoyal:
yes, I plan to do a preview this week.

PR #2624, opened 2025-01-22
Updates to the manual. NB: I used references like :ref:`ff_func` …

cbhaley:
…in gui.rst. but don't know if they work.


kovidgoyal:
They work in gui.rst as well.

PR #2626, opened 2025-01-23
Implement lazy operations for "Look & feel"

un-pogaz:
Implement a lazy operations feature for "Look & feel".

This was create to fix a user case when many Icons value are defined, the opening/loading of "Look & feel" dialog can take several time to show (test: 15000 Icons value result of ~10 second of lag).
Fix this by populate the table only when the subtab "Value icon rules viewer" is selected, rather than during the `genesis()` phase of "Look & feel" widget.
Implemented in a generic way than can be used for subtab of others sections.


un-pogaz:
Closed, will work one use it in more thing in "Look & feel".

PR #2628, opened 2025-01-23
Make Look & feel / Tag browser / tb_hierarchy respect the new "lazy" …

cbhaley:
…tab process.

I wonder if the lazy tab processing will eventually need to move into a superclass. There are several preference sections that use tabs.


cbhaley:
Pushed a commit to fix a regression caused by first commit.

cbhaley:
Pushed a commit to turn off multiple selection.

PR #2631, opened 2025-01-25
Preferences / Look & feel changes

cbhaley:
- Add infrastructure for automatic handling of lazy operations (now lazy_initialize). There is a new class LazyConfigWidgetBase that supports the lazy operations. It is a drop-in replacement for ConfigWidgetBase
- Convert tab widgets: tb display, tb_partitioning, quickview

Over time I will convert other look & feel tabs, if for no other reason than to speed up opening that set of preferences.


cbhaley:
Don't do anything with this. Some of my changes weren't pushed. I think this was caused by un-pogaz's commit.

cbhaley:
Pushed a commit with all the changes that were lost. There were a lot of them.

PR #2633, opened 2025-01-26
Perform auto-fix before build the files list

un-pogaz:
Perform auto-fix (safely) before build the hash for the files list.

### DIFF
--- a/setup/check.py
+++ b/setup/check.py
@@ -11,7 +11,7 @@ import json
 import os
 import subprocess
 
-from setup import Command, build_cache_dir, dump_json, edit_file, require_clean_git, require_git_master
+from setup import Command, build_cache_dir, dump_json, edit_file
 
 
 class Message:
--- a/setup/check.py
+++ b/setup/check.py
@@ -82,10 +82,7 @@ class Check(Command):
     def file_has_errors(self, f):
         ext = os.path.splitext(f)[1]
         if ext in {'.py', '.recipe'}:
-            if self.auto_fix:
-                p = subprocess.Popen(['ruff', 'check', '--fix', f])
-            else:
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Why? The purpose of the check command is to point out files where there
are lint issues allowing the user to fix them in an editor. --fix just
is a convenience to avoid running the editor for files where all the
problems are autofixable.

If you need to autofix files in bulk you can just run

ruff check --fix

there is no real need to have a way to run it via the check command.

un-pogaz:
Arf, sorry, I should have been clearer.

It's a point of detail because of the current behavior of setup.py check:
The check run Ruff on each file indivilualy if their hash is different from the one in the cache. But if the auto fix is actived and the file only contain autofixable errors, the command `ruff check --fix` edit the file silently and the hash will be desync from the one that will be stored in the cache, and so the file will be re-inspect during the next checking even it was never edited betwen.
By performing a bulk auto-fix before `run_check_files()`, the cache of the files builded will contain the hash of the auto-fixed version, and so no useless duplicate check later.

And this edit *don't* interup the normal file-by-file checking, because the variante command `ruff check --fix-only` used here only fix what Ruff can auto-fix but don't raise others errors if it can't, and the more several errors will be raise normaly during the file-by-file checking.

un-pogaz:
I'm made a new commit (that don't show): `run_auto_fix()` is now only informative and will don't interup the checking if the bulk auto-fix fail.

kovidgoyal:
Ah, I see what you are saying, but the common use case is to have just a few files requiring fixes. So re-checking those few files on next run is less expensive than checking every file on every fix run in advance.

PR #2634, opened 2025-01-28
Conversion of % format using flint

un-pogaz:
I found a command line tool to automatically convert Python code from old % format and .format() strings to f-strings: [ikamensh/flynt](https://github.com/ikamensh/flynt).

flynt is able to convert a good number of % format that Ruff cannot, even if it can be a little to aggressive during the conversion, so is not 1:1 of the result that flynt would have produced.

A note for a later use: flynt have also a agresive mode (not performed here), will which catch and convert almost any % format to f-strings.

### DIFF
--- a/manual/plugin_examples/editor_demo/main.py
+++ b/manual/plugin_examples/editor_demo/main.py
@@ -116,7 +116,7 @@ class DemoTool(Tool):
         num = re.search(r'[0-9.]+', val)
         if num is not None:
             num = num.group()
-            val = val.replace(num, '%f' % (float(num) * factor))
+            val = val.replace(num, f'{float(num)*factor:f}')
             style.setProperty('font-size', val)
         # We should also be dealing with the font shorthand property and
         # font sizes specified as non numbers, but those are left as exercises
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,5 +76,20 @@ docstring-quotes = 'single'
 inline-quotes = 'single'
 multiline-quotes = 'single'
 
+[tool.flynt]
+line-length = 400  # over value to catch every case
+transform-format = false  # don't transform already existing format call
+exclude = [
+    "bypy/",
+    "setup/polib.py",
+    "setup/linux-installer.py",
+    "src/calibre/ebooks/metadata/sources/",
+    "src/calibre/gui2/store/stores/",
+    "src/css_selectors/",
+    "src/polyglot/",
+    "src/templite/",
+    "src/tinycss/",
+]
+
 [tool.pylsp-mypy]
 enabled = false
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -1432,7 +1432,7 @@ class DB:
 
     @user_version.setter
     def user_version(self, val):
-        self.execute('PRAGMA user_version=%d'%int(val))
+        self.execute(f'PRAGMA user_version={int(val)}')
 
     def initialize_database(self):
         metadata_sqlite = P('metadata_sqlite.sql', data=True,
--- a/src/calibre/db/cli/cmd_restore_database.py
+++ b/src/calibre/db/cli/cmd_restore_database.py
@@ -54,7 +54,7 @@ class Progress:
         if msg is None:
             self.total = float(step)
         else:
-            prints(msg, '...', '%d%%' % int(100 * (step / self.total)))
+            prints(msg, '...', f'{int(100*(step/self.total))}%')
 
 
 def main(opts, args, dbctx):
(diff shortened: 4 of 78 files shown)


eli-schwartz:
https://x.com/raymondh/status/1205969258800275456
https://bugs.python.org/issue28307

kovidgoyal:
@eschwarts: Looking at the diff its almost all % formatting to
f-strings. The only cases where it foes to .format() is when the
arguments themselves use string literals. I dont think those few cases
ar ea performance concern. I have made a followup commit to fix a few of
them in the one place where the performance might matter slightly.

eli-schwartz:
```diff
-' id="%s"' % (aid + '-' + aid_anchor_suffix)
+f' id="{aid + "-" + aid_anchor_suffix}"'
```

It's a very personal stylistic thing but I actually don't like this anyway, since my text editor syntax highlighting treats the latter as a simple string and doesn't syntax highlight as an expression.

```diff
-'search_%s'%('author_title' if x == 'book' else x)
+'search_{}'.format('author_title' if x == 'book' else x)
```

Was changed in the PR (this is what is meant by "can be a little too aggressive" I think), and not changed again by you to become
```diff
-'search_%s'%('author_title' if x == 'book' else x)
+f'search_{"author_title" if x == 'book' else x}'
```

Same for

PR:
```diff
-'content/book_%d.html' % int(self.books_by_description[0]['id']))
+'content/book_{}.html'.format(int(self.books_by_description[0]['id'])
```

which could become:

```diff
-'content/book_%d.html' % int(self.books_by_description[0]['id']))
+f'content/book_{int(self.books_by_description[0]['id'])}.html'
```

but I question whether any of that is better and would be inclined to often prefer percent formatting :D

un-pogaz:
> this is what is meant by "can be a little too aggressive" I think

Yes, that was this. Also, not making recursive-string was the behavior of Ruff for the same conversion.

@eli-schwartz I share your reticence about automatic tools, but I find that flynt was extremely efficient and faithful. After looked the code, this is explain by it uses of the Python internal compliator and the [Abstract Syntax Trees](https://docs.python.org/3/library/ast.html) to interpret the original code and the value passed, before rewrite the same Code Node into a f-string. The code write its own code.
This doesn't mean you can trust him blindly, especially on aggressive mode, but it can catch a lot of % format and provide a good hint of the good edit.

Else @kovidgoyal, a thing I want to share about convert more string in the future:
The enormous majority of % format that need the use of the aggressive mode is because the original string contain a `%d`, the only format specificator that have none .format()/f-string equivalent. And because of a subtile behavior of it, is a implicte int convertor, flynt avoid to convert them since it cannot guess if the value pased is compatible or not.
In aggressive mode, flynt solve that by adding a explicit int conversion, but after looked a test run, the vaste majority of this explicit int convertion are useless because the value is already confirmed to be a int.

I initialy started to convert them too, but after see the numbers of files grown out of control like the pep8-strict project, I conclude that such masive change need to be perform by you when you want rather than reviewing a monstruous PR, and so resume to only the safer convesion.

kovidgoyal:
@eli-schwartz Need to upgrade the syntax highlighting in your editor then :) Here's what it looks like in my editor:
screenshot

@un-pogaz yes I know the issue with %d, the vast majority of the time it is safe to convert as just {} without int() but sadly this cant be done automatically reliably, which means tons of manual work. Someday when I have nothing better to do perhaps.

eli-schwartz:
There's a vim feature request for it but it appears to be taking a while... IIRC it doesn't help remotely, that reliably determining the end quote char for an f-string requires fully parsing the semantics of python, since any valid python expression is now valid inside of one, including nested f-strings as well as the same quote char you used to open the original string itself. It's a documented requirement that you need a PEG parser running the cpython grammar to figure out where the start and end bounds of a string is, now.

Would probably be easier if they focused on implementing pre-PEG versions of f-string syntax highlighting but it is what it is.

Mostly the increased power of them and the ability to "aggressively" rewrite things as f-strings seems to be geared towards people who want to write a 12-line program surrounded by two bytes of string literals, as a string containing a 12-line program. That's why I tend to get nervous whenever I see an f-string expression more complicated than `{func(varname)}`

kovidgoyal:
That screenshot is from nvim, FYI. Probably using tree-sitter, IIRC.

un-pogaz:
A idea that I just got to make this task easier:
Edit the `formatted_value()` in `flynt/transform/percent_transformer.py` so that it not cast a int() during the conversion of %d (Since in your case, it much easier to add manualy a int() if your not sure than remove them from almost all of the occurence).

remove this
```
val = ast.Call(
    func=ast.Name(id="int", ctx=ast.Load()),
    args=[val],
    keywords={},
)
```

PR #2638, opened 2025-01-29
Option to show only rules for category in current library

un-pogaz:
Add a option to show only rules for category available in the current library.

Also add a note in the ui about that the icon value rules are defined per-user, not per-library.


un-pogaz:
Okay.
Now the option filter also the values for the current library, and keep the last sorting after rebuild the content of the table.

cbhaley:
Looking at your changes I know we will step on each other. I'll wait until you fix the filtering problem.

PR #2640, opened 2025-01-30
Fix display all the parent values of hierarchical categories

un-pogaz:
Here the fix to display all the parent values of hierarchical categories.

Also include fix for a exception when trying to edit a rule recently created.

@cbhaley


cbhaley:
This works. And thanks for fixing that "new item" problem.

PR #2652, opened 2025-02-05
Add the east Asian language transliteration tweak we discussed.

cbhaley:
I tested it with the Gutenberg book "苦悶の欄" by Earl Derr Biggers (https://www.gutenberg.org/ebooks/39287). With the tweak empty the title was transliterated to "Ku Men noLan". With the tweak set to 'ja' it was transliterated to "Kumon no Ran". Is this transliteration correct? I don't know.


kovidgoyal:
No idea if its correct or not, I dont understand Eat Asian languages, at
all.

PR #2654, opened 2025-02-06
Add shutdown log message

un-pogaz:
Adding log message durring the shutdown phase.

It always bothered me a bit that the log never output a message of when calibre is shutdown, make it indistinguishable if calibre was shutdown normaly or if the process was kill.


kovidgoyal:
Messages should only be printed in debug mode and there is no need for
content server message, it is shown in the main gui anyway.

PR #2655, opened 2025-02-07
Commits to implement using a template for a web URL in custom columns

cbhaley:
The first two commits are straightforward. The third, render.py, was much more difficult. I'll be away until Tuesday so I'm pushing it today so you can see what I'm doing.

I tested all the column types but I'm not confident there isn't some edge case I missed.

### DIFF
--- a/src/calibre/ebooks/metadata/book/base.py
+++ b/src/calibre/ebooks/metadata/book/base.py
@@ -753,6 +753,9 @@ class Metadata:
                 res = human_readable(res)
             return (name, str(res), orig_res, fmeta)
 
+        if self.get(key, None):
+            return (key, str(self.get(key)), self.get(key), None)
+
         return (None, None, None, None)
 
     def __unicode__representation__(self):
--- a/src/calibre/ebooks/metadata/book/render.py
+++ b/src/calibre/ebooks/metadata/book/render.py
@@ -12,6 +12,7 @@ from calibre import force_unicode, prepare_string_for_xml
 from calibre.constants import filesystem_encoding
 from calibre.db.constants import DATA_DIR_NAME
 from calibre.ebooks.metadata import fmt_sidx, rating_to_stars
+from calibre.ebooks.metadata.book.formatter import SafeFormat
 from calibre.ebooks.metadata.search_internet import DEFAULT_AUTHOR_SOURCE, name_for, qquote, url_for_author_search, url_for_book_search
 from calibre.ebooks.metadata.sources.identify import urls_from_identifiers
 from calibre.library.comments import comments_to_html, markdown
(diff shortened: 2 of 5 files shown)


src/calibre/ebooks/metadata/book/base.py
```diff
@@ -753,6 +753,9 @@ def format_field_extended(self, key, series_with_index=True):
                 res = human_readable(res)
             return (name, str(res), orig_res, fmeta)
 
+        if self.get(key, None):
```
kovidgoyal, line 756: Use kv := self.get(key, None) to avoid multiple calls.

src/calibre/ebooks/metadata/book/render.py
```diff
@@ -66,6 +67,19 @@ def search_action_with_data(search_term, value, book_id, field=None, **k):
     return search_action(search_term, value, field=field, book_id=book_id, **k)
 
 
+def cc_search_action_with_data(search_term, value, book_id, fm, mi, field=None,  **k):
+    if mi is not None and fm is not None:
+        template = fm.get('display', {}).get('web_search_template')
+        if template:
+            formatter = SafeFormat()
+            mi.set('item_value', value)
+            u = formatter.safe_format(template, mi, "BOOK DETAILS WEB LINK", mi)
```
kovidgoyal, line 76: Maybe add a try/except to ignore exceptions during template evaluation.

kovidgoyal:
Looks mostly OK to me. If there are issues with individual column types we can deal with them when users report them.  I will merge and see what to do about the server.

kovidgoyal:
One thing that's left to do is implement URL quoting of the values inserted into the template. I am not sure what's the best way to do that. When using a template for authors this is done automatically by using an EvalFormatter with quoted metadata see the render_author_link() function. 

In the current implementation for cusotm columns we are using a full metadata object, with unquoted values. I have implemented quoting for item_value but for the remaining fields I think we need to have a template lang function to do quoting that can be used as needed by the user in the template.

cbhaley:
Thanks for the review and changes. :)

Regarding urlencoding item_value, I don't think this is a good idea. It means that the user cannot reliably test it to make decisions based on it. The user can quote it using one of the methods below (to be added). I will make those methods deal with the _no_plus variant.

As for quoting in general, what I thought of and propose is two formatter functions:
1) url_encode(val, use_plus) -- does what you expect, using qquote().
2) url_construct(path, (query_name, query_value)+, use_plus) -- constructs a URL by concatenating path with a query string constructed from query_name/query_value pairs. The query_values would be urlencoded. If the user needs to urlencode part of the path (the method, host, and path all concantenated together) then they do that themselves, which means use_plus doesn't apply to the path.

I'll get to this on Tuesday after I get back.

kovidgoyal:
I am fine with adding URL quoting functions to the template language,
however, making the user use them by default is not a good idea, IMO.
The common case is just simple templates like:

https://whatever.com&q={item_value}

This common case should be the easiest to use.

So I suggest adding item_value_unquoted as a field in addition.

Also regarding the url quoting functions it might be better to have a
more orthogonal create_url_query((name, value)+, use_plus)
which will return ?name=value&name=value with name and value
appropriately quoted. Then to build a URL the user can do something like

https://whatever.com/{url_quote(field1)}{create_url_query(...)}

PR #2664, opened 2025-02-12
Add a note to the L&F / Book Details custom author url tooltip that …

cbhaley:
…the author is already URL-encoded.


cbhaley:
Also includes [Bug #2098133: Setting custom icon for category doesn't work]

PR #2671, opened 2025-02-14
Much better fix for the invalid column problem in the tb icon rules …

cbhaley:
…editor. Also fixed a problem with the "Only in current library" box.


cbhaley:
@Kovid, wait on this. I found more problems left over from the "not in library" changes a while back.

PR #2672, opened 2025-02-14
Several tag browser rule editor fixes, including:

cbhaley:
* Handle non-existent lookup keys and field values.
* Allow choosing a new lookup key if the one in the pref doesn't exist.
* Disable changing the "Only in library" checkbox to avoid throwing away edits.


kovidgoyal:
Yeah I merged and took out the print myself before I saw your later
commit.

PR #2677, opened 2025-02-15
Add option in book details to suppress author search links.

cbhaley:
Question: should the constant be in calibre.constants instead of calibre.db.constants?


kovidgoyal:
db.constants is fine by me

PR #2680, opened 2025-02-20
More refactoring. Make Cancel work for newly added rules. Improvement…

cbhaley:
… in icon handling.


kovidgoyal:
merged. apologies had to rebase because of a mess up I made in the git history.

PR #2681, opened 2025-02-22
clarify wording

rhoio:
Phrases like "xxx challenged" should be avoided when speaking about disabled people.

see https://dragonscave.space/@jscholes/114044327869174785 for blind people noticing this case and https://www.cdrnys.org/blog/disability-dialogue/the-disability-dialogue-4-disability-euphemisms-that-need-to-bite-the-dust for an explanation why it should be avoided.

### DIFF
--- a/src/calibre/ebooks/conversion/plumber.py
+++ b/src/calibre/ebooks/conversion/plumber.py
@@ -422,7 +422,7 @@ OptionRecommendation(name='remove_fake_margins',
 OptionRecommendation(name='add_alt_text_to_img',
     recommended_value=False, level=OptionRecommendation.LOW,
     help=_('When an <img> tag has no alt attribute, check the associated image file for metadata that specifies alternate text, and'
-            ' use it to fill in the alt attribute. The alt attribute is used by screen readers for assisting the visually challenged.')
+            ' use it to fill in the alt attribute. The alt attribute is used by screen readers for assisting blind people.')
 ),
 
 OptionRecommendation(name='margin_top',


kovidgoyal:
Nope. alt attributes are for more than blind people. They can help
anyone that cannot see the screen from reasons ranging from
malfunctioning screens, poor vision, migraines, etc etc. Blind
people are not the only class of people to be helped by alt attributes.
Indeed, while having functional vision myself, I often use the alt text for
images to get a better understanding of images I am viewing.

If for some reason you are triggered by the phrase "visually
challenged", either come up with a phrase that is equally clear and
means the same or better just leave it be instead of wasting time that
could be far more productively employed.

jscholes:
Just to make some things clear: I wasn't expecting to prompt change or be pulled into any sort of discussion around this.  I don't find the current language "triggering", but just mildly amusing and awkward in how it does apply to blindness.

As it stands, "visually challenged" is more inclusive than the proposed change here, for the reasons already stated about who can benefit from alt attributes.  So while the PR was undoubtedly well intended, I agree with its rejection and didn't/don't endorse it.

In terms of some alternative wording, here are some suggestions should you want to use them in the future:
* "The alt attribute improves accessibility by providing text descriptions for users who cannot see or fully interpret visual content."
* "The alt attribute helps users who cannot view, clearly see or interpret images, and is used by assistive technology like screen readers."
* "The alt attribute provides text descriptions for users who cannot see or fully perceive the images for various reasons such as blindness, low vision, bandwidth constraints, etc."

As ever, thank you for this project.

PR #2683, opened 2025-02-24
Remove git credentials after checkout

Fdawgs:
This PR removes Git credentials/SSH keys after checkout as a security precaution by setting `persist-credentials` to `false`. They are not used after the initial checkout, and this stops them from accidentally leaking through a script; see [`actions/checkout` documentation](https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,7 @@ jobs:
               uses: actions/checkout@v3
               with:
                 fetch-depth: 10
+                persist-credentials: false
 
             - name: Set up Python
               uses: actions/setup-python@v3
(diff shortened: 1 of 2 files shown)


Fdawgs:
> This already uses:
> 
> ```yaml
> permissions:
>     contents: read # to fetch code (actions/checkout)
> ```
> 
> So I don't see any point in avoiding those credentials from being in the git config too.

Whilst that stops a workflow from writing to the repository it doesn't stop credentials from being written to the workflow output, i.e. sneaking in a command into `setup/unix-ci.py` to output to the shell.

eli-schwartz:
What exactly is supposed to be the consequence of sneaking a command into the CI scripts? How does hiding a credential that possesses no power affect this?

PR #2685, opened 2025-02-24
Closing in SIGTERM/SIGINT.

sergiotarxz:
I do not know if it can be done in this way. (I do not understand the implications and why you catched this signals to start with)

But I think complying with SIGINT and SIGTERM is minimum politeness in Linux with the terminal users.

Currently the situation is very bad, Calibre acts almost like a malicious program that refuses to be killed under any circumstance unless you stop it in the GUI.

If there is an important reason why Calibre cannot die that easily then precisely for that reason you should gracesfully catch SIGINT and SIGTERM, because you can handle them.

The user is going to send SIGKILL that you cannot handle if you refuse to die and that's is going to be a lot more damaging.

References https://bugs.launchpad.net/calibre/+bug/2099777

### DIFF
--- a/src/calibre/gui2/viewer/main.py
+++ b/src/calibre/gui2/viewer/main.py
@@ -196,6 +196,10 @@ def main(args=sys.argv):
                 pass
     args = processed_args
     app = Application(args, override_program_name=override, windows_app_uid=VIEWER_APP_UID)
+    import signal
+    def interrupt(a, b):
+        app.quit()
+    signal.signal(signal.SIGINT, interrupt)
     from calibre.utils.webengine import setup_default_profile
     setup_default_profile()
 


src/calibre/gui2/__init__.py
```diff
@@ -1150,7 +1150,7 @@ def setup_unix_signals(self):
 
     original_handlers = {}
     for sig in (signal.SIGINT, signal.SIGTERM):
-        original_handlers[sig] = signal.signal(sig, lambda x, y: None)
+        original_handlers[sig] = signal.signal(sig, lambda x, y: QApplication.instance().quit())
```
eli-schwartz, line 1153: > I do not know if it can be done in this way. (I do not understand the implications and why you catched this signals to start with)

Well, the current PR seems to be blatantly wrong at least. This function implements a mechanism used for example here:

https://github.com/kovidgoyal/calibre/blob/9bf4fcfd412ca2c00870975c33c7636706e61b10/src/calibre/ebooks/pdf/html_writer.py#L388-L401

which temporarily blocks all signals and then restores them after running a certain code block.

After your change, this function no longer *blocks* signals at all, which is what signal.signal per definition does. Instead, it overrides the signals to call `instance().quit()` immediately on the spot.

...

If your goal was to stop overriding and ignoring signals, I would expect a more straightforward approach to be, stop using this setup_unix_signals() function at all.

eli-schwartz (changes requested):

eli-schwartz:
Aside: it appears that at least for me, CTRL+C with ebook-edit or with the main library GUI is respected, and it quickly shuts down. But indeed, the viewer on its own, does not shut down. There must be something more nuanced than the changes in this PR, but I haven't dug deeper.

sergiotarxz:
Thank you, I will take a look to see what I can discover.

sergiotarxz:
Even if I remove calls to signal.signal still the app won't close in ctrl+c, but the funny thing is that if I I do what I do in the current patch it would work.

Currently I handle signal.signal only for ebook-viewer. Can ebook-viewer do something sensitive enough that this is not suitable?

src/calibre/gui2/__init__.py
sergiotarxz: I no longer mess with this.

eli-schwartz:
@sergiotarxz 

> I think the ebook-edit gets interrupted because a worker dies on KeyboardInterrupt, concretely `src/calibre/gui2/tweak_book/completion/worker.py`
> 
> Apparently it is a widespread problem among PyQt apps:
> 
> https://stackoverflow.com/questions/4938723/what-is-the-correct-way-to-make-my-pyqt-application-quit-when-killed-from-the-co

Nope, ebook-edit registers a signal listener but apparently that got forgotten for the Viewer :) 084a672f178cc8799749f604082ce1378c58b0df

PR #2686, opened 2025-02-28
Allow ignore kepubify for Tolino devices

beedaddy:
This is an attempt to allow that Tolino devices can ignore kepubify since they don't support the file format. This means that you are not forced to adjust the configuration if you connect a Kobo device one time and a Tolino device the next.


beedaddy:
See my [comment](https://www.mobileread.com/forums/showpost.php?p=4493561&postcount=228) on mobileread.
Please review thoroughly since I'm not that familiar with Python (and don't have a Kobo device on hand right now to cross-check).

PR #2691, opened 2025-03-09
This PR implements what we were talking about by email -- creating a …

cbhaley:
…folder containing metadata.db and .calnotes while in the GUI then passing that via the subprocess to the backend.


cbhaley:
Pushed a change that should fix the failing test.

PR #2704, opened 2025-03-23
Add codespell to pyproject.toml (and fix typo)

un-pogaz:
Adding codespell setting to `pyproject.toml` in order to make it use more easier in the future.

Unfortunately calibre will probably never fully compliant with codespell, so this setting is only to filtring a great range of false-positives, but not all (because their can be true typo error in other context).
At least codespell add a 'per-file-ignores' parameter, codespell can not be added in a regular workflow for calibre.

Also, fix various typo.


kovidgoyal:
I didn't review it in detail but from a quick glance it seems to be
change lets to let's unconditionally whether the lets means either let us or
allowed.

pyproject.toml
```diff
@@ -74,6 +74,93 @@ docstring-quotes = 'single'
...
+    "./src/calibre/ebooks/unihandecode/*",
+    "./src/calibre/ebooks/html_entities.h",
+    "./src/calibre/ebooks/html_entities.py",
+    "./src/calibre/utils/icu_test.py",
+    "./src/calibre/utils/search_query_parser_test.py",
+    "./Changelog.old.txt",
+    "./COPYRIGHT",
+    "./LICENSE",
+    "./LICENSE.rtf",
+    "./pyproject.toml",
```
eli-schwartz, line 159: I don't see any codespell reports from pyproject.toml if I remove this exclude.

pyproject.toml
```diff
@@ -74,6 +74,93 @@ docstring-quotes = 'single'
 inline-quotes = 'single'
 multiline-quotes = 'single'
 
+[tool.codespell]
+# calibre will probably never fully compliant with codespell
+# this setting is only to easily find common typo errors
+# by filtring a great range of false-positives, but not all
```
eli-schwartz, line 80: although perhaps it should.

```suggestion
# by filtering a great range of false-positives, but not all
```

un-pogaz:
@eli-schwartz Hum, yes inded. I swear that at some time codespell had analyzed `pyproject.toml`, but I can reproduct it.

@kovidgoyal codespell say only "let's", so a will check out and change at needed.

PR #2707, opened 2025-03-25
More codespell work

un-pogaz:
codespell include a feature to exclude typo error with a high precision and pretty easy to maintain (put the all line to exclude, I added the file path only for reference) and now it return 0 error/false-positive.

Now that calibre code is compliant with codespell, we could included it into the check workflow, but I'll let you decide if and how you want to add codespell to this.

Also slightly improved default options.

### DIFF
--- a/codespell.exclude
+++ b/codespell.exclude
@@ -1,0 +1,251 @@
+# ignored lines for codespell
+
+Changelog.txt
+- Arret sur images
+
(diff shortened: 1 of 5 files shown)


kovidgoyal:
I have never seen much value in codespell, so I think I'll pass on integrating it in check.

PR #2714, opened 2025-04-02
Add an "update" button to locked book details that updates the dialog…

cbhaley:
… with any new metadata for the book.

### DIFF
--- a/src/calibre/gui2/dialogs/book_info.py
+++ b/src/calibre/gui2/dialogs/book_info.py
@@ -19,6 +19,7 @@ from qt.core import (
     QLabel,
     QListView,
     QModelIndex,
+    QObject,
     QPalette,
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Is this needed because the locked dialogs dont track the current books and hence dont get updates for changes?

cbhaley:
Yes. I looked for a way to automatically update the locked windows, perhaps with a signal, but that seemed to require changes in lots of places. Thus the "button".

kovidgoyal:
Does the listened API in db.cache not work? The locked book details
could register itself as a listener via add_listener and remove itself
when closed.

kovidgoyal:
I suggest you use the listener I will take care fo wiring up events for
note/links changes.

PR #2748, opened 2025-05-10
Add validation of column keys for collections to the kobo driver.

cbhaley:
There are three commits for this change. I did this to (I hope) better explain what the changes were for.

### DIFF
--- a/src/calibre/customize/__init__.py
+++ b/src/calibre/customize/__init__.py
@@ -149,17 +149,35 @@ class Plugin:  # {{{
 
         from calibre.gui2 import gprefs
 
+        class ConfigDialog(QDialog):
+
+            def __init__(self, parent, config_widget):
+                super().__init__(parent)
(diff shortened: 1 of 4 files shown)


kovidgoyal:
I am pretty nervous about changing the semantics of do_user_config(). Currently a failure to validate just means that the data is not saved. It could be that some config widgets perhaps in third party plugins rely on this behavior.

I suggest gating the semantic change behind something like if getattr(config_widget, 'retry_on_validation_failure', False).

cbhaley:
Instead of re-execing the dialog, the validate() is optionally done in the accept() method. This is gated by
- the validate() method must exist in the config_widget
- the attribute self.validate_on_accept must exist in the config_widget and be True.

This implementation should be compatible with previous behavior.

cbhaley:
Pushed a commit to remove some print statements

PR #2766, opened 2025-05-29
Add context menu action to copy the source code for the currently …

cbhaley:
…selected template to the editor.


cbhaley:
This is for Enhancement #2111957: Copy stored template to template editor

PR #2769, opened 2025-06-07
Define _FILE_OFFSET_BITS=64 and replace usage of pread64

vimproved:
pread64 and its fellow LFS64 interfaces are no longer present on musl systems as of musl 1.2.4, as pread is always 64-bit on musl. Instead of conditionalizing usage of pread64 to glibc, use pread instead and pass _FILE_OFFSET_BITS=64 to ensure pread is 64-bit on glibc. This should not have an any effect on other platforms.


vimproved:
Thanks!

PR #2780, opened 2025-06-21
Allow native linux file dialogs on NixOS

zspher:
In NixOS, native file dialogs are not detected due to `getattr(sys, 'frozen', False)`. This patch removes the sys.frozen check to allow native file dialogs in NixOS.

Please tell me if there is a better way to do this.
for reference: https://github.com/NixOS/nixpkgs/pull/416597

### DIFF
--- a/src/calibre/gui2/__init__.py
+++ b/src/calibre/gui2/__init__.py
@@ -969,7 +969,7 @@ if iswindows and 'CALIBRE_NO_NATIVE_FILEDIALOGS' not in os.environ:
     from calibre.gui2.win_file_dialogs import is_ok as has_windows_file_dialog_helper
     has_windows_file_dialog_helper = has_windows_file_dialog_helper()
 has_linux_file_dialog_helper = False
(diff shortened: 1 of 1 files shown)


kovidgoyal:
On non frozen builds, calibre use the system Qt. The system Qt should
already be using native file dialogs automatically. That is why that
check is there in the first place.

zspher:
yes, but the native file dialog from Qt does not use the preset icon theme. And what if the user wants to use the set xdg-desktop-portal native file dialog that zenith or kdialog uses.

kovidgoyal:
Then you set the

QT_QPA_PLATFORMTHEME=xdgdesktopportal

env var. When built against system qt calibre will defer to qt for such
things, that's not going to change.

eli-schwartz:
Yeah, hard "no" in my opinion for making all software have custom code for this. Which is what it sounds like NixOS wants.

One package having custom code used only in prebuilt binaries, is reasonable so that prebuilt binaries can defer to system policy even though *embedded* Qt doesn't necessarily get built with arbitrary support for e.g. portals. But it's quite reasonable for that code to only activate for embedded builds.

PR #2793, opened 2025-07-06
Add tweak to permit displaying series_sort for series in the …

cbhaley:
…tag browser.


cbhaley:
Fixed problem with parentheses.

PR #2796, opened 2025-07-08
massive performance improvement for restoring database

quarz12:
by wrapping all the database interaction into a single transaction, this brings the time to restore my 5.5k book database from over 5 hours to only 5 minutes


kovidgoyal:
Using begin and commit is dangerous as they cannot be nested. Instead use savepoints. Dont know if that would still make a performance difference though. As of my latest commit you can use savepoints with

```py
with db.new_api:
   code to run inside transaction
```

PR #2798, opened 2025-07-09
format_duration() changes:

cbhaley:
1) use [ ] instead of { } so it can be used in single function mode.
2) use zero for empty values.


cbhaley:
Pushed a commit to fix a problem in the documentation.

PR #2799, opened 2025-07-09
Improve GPM `format_duration()`

un-pogaz:
Improve the GPM `format_duration` function by:
- remove the `largest_unit` parameter (it guessed from the template itself)
- improving pattern match
- allowing empty suffixes

### DIFF
--- a/src/calibre/utils/formatter_functions.py
+++ b/src/calibre/utils/formatter_functions.py
@@ -3512,22 +3512,24 @@ This can be useful to truncate a value.
 
         def repl(mo):
             fmt_char = mo.group(1)
-            suffixes = re.split(r'\|', mo.group(3) or '')
-            match len(suffixes):
-                case 1 if not suffixes[0]:
-                    zero_suffix = one_suffix = more_suffix = fmt_char.lower() + ' '
-                case 1:
-                    zero_suffix = one_suffix = more_suffix = suffixes[0]
-                case 2:
-                    zero_suffix = more_suffix = suffixes[0]
-                    one_suffix = suffixes[1]
-                case 3:
-                    zero_suffix = suffixes[0]
-                    one_suffix = suffixes[1]
-                    more_suffix = suffixes[2]
-                case _:
-                    raise ValueError(_('The group {} has too many suffixes').format(fmt_char))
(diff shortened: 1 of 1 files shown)


kovidgoyal:
@cbhaley please review

cbhaley:
I don't think the first commit 'Retrieve the ...' works correctly. If I leave out the [s] specifier then largest_unit doesn't get assigned. It also seems to be order dependent. Finally, I wanted that parameter there so I could specify (for example) largest_unit as 'd' but only use h, m, s in the template. I don't think this functionality should be removed. 

Regarding 'allow empty suffixes', this seems to be there to allow using '[w:]'. I don't see the need for that, as one can already say '[w]' to get an empty suffix. That said, I don't object to it.

Regarding 'improving pattern match', this seems to be a matter of taste and it increases complexity. As far as I can see it doesn't actually improve the behavior. But as above, I don't object to it.

If this commit or a follow-on is accepted then un_pogaz should be the one who maintains it.

un-pogaz:
I see what you mean toward the behavior of `largest_unit`. This seems to me to be a very specific use case, so I will make `largest_unit` an optional parameter that overrides automatic detection.
And it not order dependent since it iter and check *every* selectors inside the template.

Curently, `[w:]` output `'9w '`, so it append the default suffix like you don't added the `:`. We can use `[w: ]` to remove the suffixe, but a not very confort to force the append a space/character after the result number.

For improving pattern match, I agree that the current one in not the best for reading and I will change for '\[(.)(:(.*?))?\]'. The main idea it's to make the selection more strict, but I confess I may look too far.

cbhaley:
From my reading of the code, if there is no 's' selector then largest_unit doesn't get set. It also seems to require that the 's' be first.

cbhaley:
I see what you mean about removing the default suffix.

un-pogaz:
Okay, I push a new version of the PR.

'improving pattern match' was merged into the "main commit" because is way more easy to handel.

And after test, 's' don't need to be first. It work like your original code.
Apparently, python trick: if `largest_unit = ''` (empty string) then `largest_unit in 'dhms'` will return `True`, so the others greater match case/selector are corretly tested.

cbhaley:
@Kovid: I'm ok with the changes.

PR #2803, opened 2025-07-11
Allow custom-metadata comment rows to grow with the window.

Mikescher:
### Problem description

I'm using a few custom columns to track data such as:  
 - when I've read a book
 - my own comments/notes/etc

Especially the multiline custom-fields are a bit  annoying, because they have a fixed height and I can't resize them to be bigger.  
 
image

The "Comments" field under Basic-metadata behaves a bit better, and actually fills the available width of the window.

### Solution

I've changed all `is_comments` rows to have `row_stretch`.  
This way, multiline (= comment) rows grow with the window, and if there are multiple they all grow equally.

image

### DIFF
--- a/src/calibre/gui2/custom_column_widgets.py
+++ b/src/calibre/gui2/custom_column_widgets.py
@@ -977,6 +977,9 @@ def populate_metadata_page(layout, db, book_id, bulk=False, two_column=False, pa
         if is_comments:
             layout.addLayout(l, row, column, layout_rows_for_comments, 1)
             layout.setColumnStretch(column, 100)
+            # All multiline fields should grow with the window
+            for i in range(layout_rows_for_comments):
+                layout.setRowStretch(row + i, 100)
             row += layout_rows_for_comments
         else:
             layout.addLayout(l, row, column, 1, 1)


kovidgoyal:
Custom column widgets areused in multiple places. Off the top of my head, the three types (layouts) of the edit single metadata dialog and the bulk metadata dialog and the edit metadata tab of preferences->Look & feel. You need to check the effect of these changes in all those locations or, safer, only change the stretch in the single metadata dialog when using the normal layout.

Mikescher:
Hi - thanks for the fast answer *(even though I was already sleeping at that point :D)*,  

> Custom column widgets are used in multiple places [...] edit single metadata dialog

your are right - the method is also used in the bulk-edit dialog.  
it does not really matter in this dialog, because comment-fields are not displayed in the bulk-edit dialog.  
(the condition is in [custom_column_widgets.py#L957](https://github.com/kovidgoyal/calibre/blob/9e92686c9d1d3146a3ff8f30004e5289f0524da2/src/calibre/gui2/custom_column_widgets.py#L957)).

But you could definitely make an argument for only applying the rowStretch `if not bulk`. That way it only really applies on the single dialog I tested.  
If you want, I can change it?

> and the edit metadata tab of preferences->Look & feel.

Hmm, I can't seem to find the place you are referencing.  
Looking through the UI, I can't find a dialog under Look-and-feel which renders the custom-column widgets.  
And in-code the `populate_metadata_page` function is only called from two places: `metadata_bulk.py` and `metadata/single.py`.  

The one "special" case that exists is with single-column layout (instead of two-column layout).  
But that also works (input fields grow with the window - but keep their min height).  
I've added two screenshots to show those cases:


### EditMetadata [one-column layout]

image

### Edit-Metadata-in-Bulk (no multi-line fields)

image

PR #2810, opened 2025-07-16
MTP Kindle Serial not detected

dpecos:
When connecting my kindle and showing the device information, the serial number appears empty:

```
Kindle Paperwhite
Serial number: 
Manufacturer: Amazon
Model: Kindle Paperwhite
ids: (1, 17, 6473, 39297, 'xxxxx')
Device version: 
Storage:
[{'capacity': 12349591552,
  'freespace_bytes': 11801124864,
  'freespace_objects': 4294967295,
  'id': 65537,
  'name': 'Internal Storage',
  'removable': False,
  'rw': True,
  'volume_id': '-10001'}]
```

causing that the persisted config is shared across all the MTP devices, as they are not properly identified (`device-`):

```
{
  "blacklist": [],
  "device-": {
    "apnx": {
...
```

With this change, I'm assuming the last value of the list of IDs is the actual serial device, and thus using it when the default serial number is empty, fixing the scenarios stated above.

### DIFF
--- a/src/calibre/devices/mtp/unix/driver.py
+++ b/src/calibre/devices/mtp/unix/driver.py
@@ -274,7 +274,7 @@ class MTP_DEVICE(MTPDeviceBase):
         self.current_friendly_name = self.dev.friendly_name
         if not self.current_friendly_name:
             self.current_friendly_name = self.dev.model_name or _('Unknown MTP device')
-        self.current_serial_num = snum
+        self.current_serial_num = snum or self.dev.ids[-1]
         self.currently_connected_dev = connected_device
 
     @synchronous


kovidgoyal:
What OS are you on that the serial number is empty?

kovidgoyal:
It might make sense to store this preference not per device/serial number, but instead globally so the user does not have to turn it on for each Kindle device individually.

dpecos:
> What OS are you on that the serial number is empty?

Linux (arch)

dpecos:
> It might make sense to store this preference not per device/serial number, but instead globally so the user does not have to turn it on for each Kindle device individually.

The code is actually trying to use the serial number when writing the config, so if a shared config is your preference, the code should be amended, otherwise you’ll have kind of erratic behavior.

dpecos:
Also ‘current_serial_number’ is used in multiple places, not sure the consequences of having an empty value there.

🤷‍♂️

kovidgoyal:
I have already fixed the empty serial number issue, see the commit that
closed this PR. The question is whether to make the APNX settings global
or not. Probably store them in the config used for the USBMS Kindle
driver and use that in the MTP driver as well.

kovidgoyal:
And I have changed the MTP driver to use the APNX settings from the KINDLE2 driver. The settings are now common across all kindles and can be configured via either KINDLE2 or MTP.

dpecos:
> And I have changed the MTP driver to use the APNX settings from the KINDLE2 driver. The settings are now common across all kindles and can be configured via either KINDLE2 or MTP.

Nice!

dpecos:
I've just pulled your changes, and the device serial number is still blank:

image

And I've checked from the device details in the kindle, and indeed the serial number is the last of the IDs in the screenshot.

kovidgoyal:
The changes are in C code which means you need to actually build it to
see them.

PR #2814, opened 2025-07-21
New FFML "_to_transifix" functionality and syntax error fixes

cbhaley:


cbhaley:
Added two new commits.

kovidgoyal:
Hmm, thinking about it some more, this is going to be a big performance
killer. Uploading modified strings to transifex means that the
translations are keyed off the modified strings, which means that
the _() function in formatter_functions() is going to actually have to
do the translation to transifex format and then lookup the actual
translation using that modified string. Since these are all class
attributes and the like, it means simply importing
formatter_functions.py will become quite slow.

cbhaley:
Won't it happen only on the first import (module import)? 

formatter_functions already has its own version of _(), which could do the FFML processing before instantiating TranslatedStringWithRaw() class already there.

kovidgoyal:
On Tue, Jul 22, 2025 at 02:18:36AM -0700, Charles Haley wrote:
> cbhaley left a comment (kovidgoyal/calibre#2814)
>
> Won't it happen only on the first import (module import)?

Yes, but that affects startup time. It will need some benchmarking and
analysis. Maybe make them all inherit from a base class that defines
method based access to the doc attributes.

cbhaley:
I see no performance difference if I change _() in formatter functions to be

```
def _(txt):
    from calibre.utils.ffml_processor import FFMLProcessor
    txt = FFMLProcessor().document_to_transifex(txt, '', safe=True)
    translated = xlated(txt)
    return TranslatedStringWithRaw(txt, translated, txt, translated)
```

cbhaley:
Do you want me to make the above change to _() or are you doing it? I'm happy to as I have another change to push to fix the template documentation display problem that @Michele pointed out.

kovidgoyal:
On Tue, Jul 22, 2025 at 02:52:21AM -0700, Charles Haley wrote:
> cbhaley left a comment (kovidgoyal/calibre#2814)
>
> I see no performance difference if I change _() in formatter functions to be

It adds between 5 and 10 ms to import time of that module on my machine.
Measured with:

hyperfine calibre-debug -c 'import calibre.utils.formatter_functions'
```
Current:
Time (mean ± σ):     375.7 ms ±   4.8 ms    [User: 328.6 ms, System: 37.3 ms]
With changes:
Time (mean ± σ):     386.3 ms ±   6.5 ms    [User: 336.2 ms, System: 40.6 ms]
```

kovidgoyal:
I will do the integration of the new translation code. Doing that change alone is not sufficient, the code to generate the POT file also has to be changed. And I want to see if I can mitigate the performance impact.

cbhaley:
So what I need to do is move the FFML from template_general_info.py to formatter_functions.py. You will do whatever you decide to do with "_()" etc.

kovidgoyal:
On Tue, Jul 22, 2025 at 03:10:16AM -0700, Charles Haley wrote:
> cbhaley left a comment (kovidgoyal/calibre#2814)
>
> So what I need to do is move the FFML from template_general_info.py to formatter_functions.py. You will do whatever you decide to do with "_()" etc.

yes.

PR #2815, opened 2025-07-22
FFML related bug fixes and changes

cbhaley:


cbhaley:
FWIW: I ran my "identical output" tests after these changes. Raw and transifex'ed doc strings produce identical rst and html.

kovidgoyal:
The current code (not including this PR) currently is add a trailing
space to some lines when translating to transifex format. Trailing
spaces are bad for translators since again they dont know if they are
significant and they are anyway hard to spot. An example is
builtinSublist its docs when passed through document_to_transifex result
in a trailing space on all the lines in the [LIST]

cbhaley:
It appears that if in python code I reference formatter function BuiltinSublist .doc I now get the transfexed version instead of what is shown in the code. Is this true? Is this what we want?

cbhaley:
To be clear, it appears that there isn't a way to get the text as shown in formatter_functions.py. This makes editing the documentation with the built-in FFML editor difficult because there is no way to reverse the processing.

kovidgoyal:
No, that should no longer be the case as per my latest commit.

PR #2816, opened 2025-07-22
Two FFML-related changes:

cbhaley:
1) Eliminate the space at the end of some lines such as list items.
2) Remove a spurious backslash from the FFML documentation.

NB: removing the space means that the generated rst and html are slightly different from what they were before the change. The changes are white space and don't appear to change what one sees in the interpreted html and rst.

The html and rst output remains identical with original input or rst-formatted input.


cbhaley:
I need to take your latest commit into consideration.

kovidgoyal:
I am done working on this so I shouldnt be stepping on your changes anymore.

PR #2817, opened 2025-07-22
Resubmission of two FFML changes:

cbhaley:
1) Eliminate the space at the end of some lines such as list items.
2) Remove a spurious backslash from the FFML documentation.

NB: removing the space means that the generated rst and html are slightly different from what they were before the change. The changes are white space and don't appear to change what one sees in the interpreted html and rst.

The html and rst output remains identical with original input or rst-formatted input.

### DIFF
--- a/src/calibre/gui2/dialogs/template_general_info.py
+++ b/src/calibre/gui2/dialogs/template_general_info.py
@@ -66,8 +66,9 @@ a space before the closing \`\` characters. Trailing blanks in the code text are
   Example: \:guilabel\:\`Preferences->Advanced->Template functions\`. For HTML the produced text is in a different font, as in: :guilabel:`Some text`
 [*][B]Empty lines[/B], indicated by two newlines in a row. A visible empty line in the FFML
 will become an empty line in the output.
-[*][B]URLs.[/B] The syntax is similar to BBCODE: ``[URL href="http..."]Link text[/URL]``.\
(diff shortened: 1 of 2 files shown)


un-pogaz:
I want report a bug, and since all this stuf is in-dev, not sur to use the launchpad:

Extranous new line is not suppresed and remplaced by a space before a CODE_TEXT or REF block inside a regular text line. This could be affected others type of blocks.

Example of function affected: `connected_device_name()`, `list_count()` and `extra_file_names()`

cbhaley:
This is true, and hard to fix without major changes to the FFML processor  effectively introducing context. That newline doesn't hurt anything beyond making a few lines shorter.

cbhaley:
To be precise: the problem is specifically caused by markup such as code text at the beginning of a line. For example, for list_count(), moving the "by" from the end of the first line to the beginning of the second line avoids the problem. For connected_device_name() moving the word "are" from the end of line 3 to the beginning of line 4 avoids the problem.

To "fix" it the FFML processor would need to backtrack to see how the previous TEXT block ended. That is hard and IMO not needed.

un-pogaz:
Ah, know issue, okay.
And yeah, it such a minor case that don't deserve the headcache to fix it.

cbhaley:
While sleeping I worked out how to fix this. Instead of using look behind with context during output generation, I can use look ahead while building the parse tree. This is much more straightforward. I'll make a PR once I've tested it sufficiently.

PR #2826, opened 2025-07-29
Keep emphasis style on ebook conversion

blastrock:
Hi,

I have been converting a few Japanese ebooks from EPUB to AZW3, and I found that the property `-epub-text-emphasis` wasn't translated to `-webkit` for the Kindle.

This style is often used in Japanese to emphasis text:

image

This PR fixes this issue.

### DIFF
--- a/src/calibre/ebooks/oeb/stylizer.py
+++ b/src/calibre/ebooks/oeb/stylizer.py
@@ -185,9 +185,17 @@ class StylizerRules:
                 size = 'xx-small'
             if size in FONT_SIZE_NAMES:
                 style['font-size'] = f'{self.profile.fnames[size]/float(self.profile.fbase):.1f}rem'
-        if '-epub-writing-mode' in style:
-            for x in ('-webkit-writing-mode', 'writing-mode'):
-                style[x] = style.get(x, style['-epub-writing-mode'])
+        UNPREFIXED_PROPERTIES = (
+            ['writing-mode'] +
+            [f'text-emphasis{x}' for x in ('', '-color', '-position', '-style')]
+        )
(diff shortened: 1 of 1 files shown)


kovidgoyal:
That's just the shorthand version, you have to handle the individual ones as well as documented in: https://www.w3.org/TR/epub-33/#sec-css-prefixed-text-decoration

And only set the unprefixed property if it is not already present. Do we actually need the -webkit- prefied version? Are there any reading programs that support that and not the unprefixed one?

Finally you should delete the -epub- prefixed ones after conversion to unprefixed.

blastrock:
> That's just the shorthand version, you have to handle the individual ones as well as documented in: https://www.w3.org/TR/epub-33/#sec-css-prefixed-text-decoration

Indeed, fixed now.

> And only set the unprefixed property if it is not already present

That should be handled by the `style.get()` call which will only take the `-epub` property as a fallback. I kept the style of `writing-mode`, just above, but I can write an explicit `if` if you prefer.

> Do we actually need the -webkit- prefied version? Are there any reading programs that support that and not the unprefixed one?

I just tested, and my Kindle 5.17.1.0.3 does not seem to support the unprefixed one. Also, the official Amazon ebooks I have use the `-webkit` one, this is how I found the fix.

> Finally you should delete the -epub- prefixed ones after conversion to unprefixed.

I am not familiar at all with calibre's code, but I understand that the stylizer rules are used for other conversions as well, and in both direction, right? I am not sure how I can easily remove the -epub ones only when needed.

Also, the previous code with `writing-mode` didn't bother with that. And official Amazon ebooks do contain `-epub` properties as well, which are probably just ignored. Is it worth trying to clean up those properties?

blastrock:
Thank you!

PR #2827, opened 2025-07-30
Update economist.recipe

unkn0w7n:


kovidgoyal:
Deleting economist_free.recipe will break downloads for people that have scheduled it. And what's the motivation for changing the recipe? The existing recipe is working fine.

unkn0w7n:
`Deleting economist_free.recipe will break downloads`
hmm... i mean it's just a duplicate, it's been there for so long, why not make them use the main one?

the old graphql code no longer works, so had to remove all that code.
`And what's the motivation for changing the recipe?`
this recipe supports interactive articles using new graphql requests. People were complaining about 1843 articles not having content.

I thought of keeping in all the old code and just adding more lines, but it was too much unused code.

kovidgoyal:
On Tue, Jul 29, 2025 at 09:28:01PM -0700, unkn_wn wrote:
> unkn0w7n left a comment (kovidgoyal/calibre#2827)
>
> `Deleting economist_free.recipe will break downloads`
> hmm... i mean it's just a duplicate, it's been there for so long, why not make them use the main one?

I dont like to break user workflows, not to mention that if I do that,
it will cause endless bug reports I have to deal with, so please restore
it.

>
> the old graphql code no longer works, so had to remove all that code.
> `And what's the motivation for changing the recipe?`
> this recipe supports interactive articles using new graphql requests. People were complaining about 1843 articles not having content.

I havent looked at the interactive articles but do they actually work on
eink devices? Doesnt it require JS for the interactive bits?

unkn0w7n:
`I havent looked at the interactive articles but do they actually work on
eink devices? Doesnt it require JS for the interactive bits?`
no the graphql version provides html content for in-app viewing, keeping the text and images.
You can see this with the interactive articles of 1843 in the latest summer issue.

PR #2830, opened 2025-07-31
Update economist_world_ahead.recipe

unkn0w7n:


unkn0w7n:
how is `def publication_date(self): ` different from from `self.timefmt` 

I set timefmt to the publication date of the magazine.

kovidgoyal:
timefmt controls how the publication date is formatted.

PR #2835, opened 2025-08-08
TTS: gracefully handle missing piper support

eli-schwartz:
Third party redistributors might choose to skip distributing this for a couple reasons:
- missing dependencies
- lack of interest in TTS as a feature

Lay some groundwork for handling this with fewer error message popups. In particular note that speechd / flite depend on PyQt6 being built with it, so support *may* appear dynamically after calibre is installed, and available_engines queries Qt to see what is available. Piper is built as part of calibre though, and if it has been patched out or skipped via `setup.py build --only=xxx` we can at least avoid claiming it's there.

Entrypoints into TTS eventually tend to consolidate into creating the backend. This gives us one consistent place to raise errors for missing backends... which however doesn't handle forcing a backend name. A forced backend that is unavailable ended up hitting the "no prefs" fallback code to use the default engine, which returned a different backend than the one which is *forced*, and later a KeyError when tweak_book attempted to access the backend name it forced but which didn't exist.

Instead, raise an immediate "TTS engine piper is not available" error dialog box, preventing any further confusing tracebacks.

### DIFF
--- a/src/calibre/gui2/tts/types.py
+++ b/src/calibre/gui2/tts/types.py
@@ -234,11 +234,18 @@ def available_engines() -> dict[str, EngineMetadata]:
             ), True)
         elif x == 'speechd':
             continue
-    ans['piper'] = EngineMetadata('piper', _('The Piper Neural Engine'), _(
-        'The "piper" engine can track the currently spoken sentence on screen. It uses a neural network '
-        'for natural sounding voices. The neural network is run locally on your computer, it is fairly resource intensive to run.'
-    ), TrackingCapability.Sentence, can_change_pitch=False, voices_have_quality_metadata=True, has_managed_voices=True,
-    has_sentence_delay=True)
+
+    try:
+        import calibre_extensions.piper
+    except ImportError:
+        pass
+    else:
+        ans['piper'] = EngineMetadata('piper', _('The Piper Neural Engine'), _(
+            'The "piper" engine can track the currently spoken sentence on screen. It uses a neural network '
+            'for natural sounding voices. The neural network is run locally on your computer, it is fairly resource intensive to run.'
+        ), TrackingCapability.Sentence, can_change_pitch=False, voices_have_quality_metadata=True, has_managed_voices=True,
+        has_sentence_delay=True)
+
     if islinux:
         try:
             from speechd.paths import SPD_SPAWN_CMD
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Gentoo currently doesn't package libonnxruntime. Nor piper, but that used to "gracefully" fallback at runtime. :P 

I *think* I also may want an option for `python setup.py build` to exclude building a single extension, so I can run `build` + `install` omitting piper. But I'm not sure how such an option should be shaped...

kovidgoyal:
While this PR may be worth merging on its own merits, note that calibre
no longer uses any piper code. Basically the only required tts related
deps are now onnxruntime and espeak-ng, I don't think expecting Linux
distributors to have those is an undue burden. There are some features
in calibre that *require* the ability to use piper (aka have onnxruntime
and espeak-ng), the rest of the backends are entirely optional already
and fallback to piper automatically.

kovidgoyal:
On Thu, Aug 07, 2025 at 10:50:51PM -0700, Eli Schwartz wrote:
> eli-schwartz left a comment (kovidgoyal/calibre#2835)
>
> Gentoo currently doesn't package libonnxruntime. Nor piper, but that used to "gracefully" fallback at runtime. :P

Why not? It's fast becoming a common runtime that will likely be used by
multiple "local AI" using applications in the future, which is an
outcome we should all encourage, rather than going out to the cloud for
all "AI" usage. And man I hate the term AI.

>
> I *think* I also may want an option for `python setup.py build` to exclude building a single extension, so I can run `build` + `install` omitting piper. But I'm not sure how such an option should be shaped...

You cant do that so easily, you'd have to patch/exclude the tests as
well. At which point you might as well just patch calibre to not build
it.

eli-schwartz:
I am not sure why it isn't packaged yet, I plan to ask around and see if our machine learning team can pull it in. However it's currently an issue with packaging 8.8.0 today, since I can't package the piper extension without libonnxruntime, so...

My major goal here is actually just to land code that looks reasonably useful "regardless". It's possible for users to build a Gentoo system with global `USE="-speech"` so it would be nice to support that even once onnx is packaged, and ideally the resulting error message if you try to do that is "you haven't installed with support for XXX" rather than "KeyError". :D this bothered me a bit before 8.8 as well, but I felt motivated to do a bit of that work tonight, so there we have it.

I know that optional dependencies are kinda jank and not fun to support gracefully. It's one of those "one day Eli will finally fix this" things, probably.

eli-schwartz:
> Why not? It's fast becoming a common runtime that will likely be used by multiple "local AI" using applications in the future, which is an outcome we should all encourage, rather than going out to the cloud for all "AI" usage. And man I hate the term AI.

Yes, but at least synthesizing voices for text to speech is one of the rare and very few unambiguously good use cases!

eli-schwartz:
(The other reason I can't package the piper speech backend is because espeak-ng doesn't have a stable release with the necessary API!)

kovidgoyal:
On Thu, Aug 07, 2025 at 11:19:30PM -0700, Eli Schwartz wrote:
> eli-schwartz left a comment (kovidgoyal/calibre#2835)
>
> (The other reason I can't package the piper speech backend is because espeak-ng doesn't have a stable release with the necessary API!)

You just need this pretty trivial commit:
https://github.com/espeak-ng/espeak-ng/commit/2108b1e8ae02f49cc909894a1024efdfde6682fd

PR #2838, opened 2025-08-10
feat: Add LLM tab to Lookup panel

amirthfultehrani:
Dear Kovid, may this pull request find you very well!

Following our discussion between each other and peers on [MobileRead](https://www.mobileread.com/forums/showthread.php?t=369298), I have implemented the proposed LLM integration as a tab in the lookout panel. It is fully modular (all logic isolated in llm.py) and lazy-loaded (nothing loads until the LLM tab is clicked; zero cost for non-users). It also persists the last-open tab, just [as you suggested](https://www.mobileread.com/forums/showpost.php?p=4530010&postcount=18).

Images of the feature:
image
image
image

Also, some additional ideas I had:
- Allowing the user to ask follow-up questions about the same piece of selected text (e.g., user could "Summarize" then "Elaborate" then "Translate" all without having to go to a separate window)
- Adding a small "Add as note" button next to the LLM's response - thus, the AI-generated response can be appended to the "Notes" field of the original highlight
- Optional real-time cost estimate for the session (instead of just the # of API calls)

Thank you for all your time and support in this matter, for this opportunity to contribute, and for all that you have done for Calibre! I look forward to your feedback, Kovid!

---
Integrates LLM query interface as a new tab in the existing Lookup panel (within Calibre's E-Book Viewer), following feedback from Kovid Goyal on MobileRead forums. The interface is implemented in a modular way + is lazily loaded on first use to ensure zero performance cost when inactive.

### DIFF
--- a/src/calibre/gui2/viewer/llm.py
+++ b/src/calibre/gui2/viewer/llm.py
@@ -1,0 +1,641 @@
+# License: GPL v3 Copyright: 2025, Amir Tehrani and Kovid Goyal
+
+import json
+from threading import Thread
+from urllib import request
+from urllib.error import HTTPError, URLError
+from urllib.parse import urlparse, parse_qs
+
+from qt.core import (
+    QAbstractItemView, QComboBox, QDialog, QDialogButtonBox, QEvent, QFormLayout,
+    QGridLayout, QGroupBox, QHBoxLayout, QIcon, QLabel, QLineEdit, QListWidget,
+    QListWidgetItem, QMessageBox, QPushButton, QSizePolicy, Qt, QTextBrowser,
+    QTextEdit, QVBoxLayout, QWidget, pyqtSignal
+)
+
+from calibre.gui2.viewer.config import vprefs
+from calibre.utils.localization import _
+from polyglot.binary import as_hex_unicode, from_hex_unicode
+
+# --- Backend Abstraction & Cost Data ---
+MODEL_COSTS = {
+    # Anthropic
+    'anthropic/claude-3-haiku': (0.25, 1.25),
+    'anthropic/claude-3.5-sonnet': (3.00, 15.00),
+    'anthropic/claude-3.7-sonnet': (3.00, 15.00),
+    'anthropic/claude-sonnet-4': (3.00, 15.00),
+
+    # DeepSeek
+    'deepseek/deepseek-chat-v3-0324': (0.18, 0.72),
+
+    # Google
+    'google/gemini-1.5-flash': (0.075, 0.30),
+    'google/gemini-1.5-pro': (1.25, 5.00),
+    'google/gemini-2.0-flash-001': (0.10, 0.40),
+    'google/gemini-2.5-flash': (0.30, 2.50),
+    'google/gemini-2.5-flash-lite': (0.10, 0.40),
+    'google/gemini-2.5-pro': (1.25, 10.00),
+
+    # Meta
+    'meta-llama/llama-3.1-8b-instruct': (0.015, 0.02),
+    'meta-llama/llama-3.1-70b-instruct': (0.10, 0.28),
+
+    # Mistral
+    'mistralai/mistral-7b-instruct': (0.028, 0.054),
+    'mistralai/mistral-nemo': (0.008, 0.05),
+
+    # MoonshotAI
+    'moonshotai/kimi-k2': (0.14, 2.49),
+
+    # OpenAI
+    'openai/gpt-4.1-mini': (0.40, 1.60),
+    'openai/gpt-4o': (2.50, 10.00),
+    'openai/gpt-4o-mini': (0.15, 0.60),
+    'openai/gpt-5': (1.25, 10.00),
+    'openai/gpt-5-mini': (0.25, 2.00),
+    'openai/gpt-oss-120b': (0.072, 0.28),
+
+    # Qwen
+    'qwen/qwen3-coder': (0.20, 0.80),
+
+    # ZhipuAI
+    'z-ai/glm-4.5': (0.20, 0.80),
+
+    # Default Fallback
+    'default': (0.50, 1.50)
+}
+
+API_PROVIDERS = {
+    'openrouter': {
+        'url': "https://openrouter.ai/api/v1/chat/completions",
+        'headers': lambda api_key: {
+            'Authorization': f'Bearer {api_key}',
+            'Content-Type': 'application/json',
+            'HTTP-Referer': 'https://github.com/kovidgoyal/calibre',
+            'X-Title': 'Calibre E-book Viewer'
+        },
+        'payload': lambda model_id, messages: {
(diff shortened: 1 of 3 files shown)


kovidgoyal:
cool, I'll review this when I next have some free time.

eli-schwartz (commented):
This feels like something significantly better suited to be a third-party Viewer plugin. I wonder if it would be better off to make those possible instead?

kovidgoyal:
Unfortunately, creating a framework for viewer plugins is a *lot* of
work (because the viewer is a mix of rapydscript and python and runs in
a standalone viewer program as well as in browser). That is why I have
not made one yet.

kovidgoyal:
I definitely think allowing the user to continue the conversation is useful. In my own use of LLMs I tend to often ask followup questions, being able to do so in the same window will be useful. Add as note is fine, I can see that coming in handy. As for cost estimate probably have both cost and number of calls.

To save some space move the titles like "Quick actions" and "Custom Prompt" into the group box border using QGroupBox::setTitle.

Using OpenRouter is good but we should keep the backend design flexible so in the future we can change providers if and when needed since this space is so full of churn at the moment. 

Unfortunately I am deep in the middle of something else at the moment so it will be a little while before I can do a more thorough review.

kovidgoyal:
Another useful thing might be to send the book title and author as part
of the prompt to the LLM, so that it has more context about the query.

amirthfultehrani:
Dear Kovid & Eli (and other friends who may be viewing this thread): may you all be well! 

Incorporating Kovid's feedback, the LLM integration feature is now, I hope, pretty complete, stable, and polished. I have updated my pull request in reflection.

In particular:

- Users can now ask follow-up questions about the same selected text
image
- A main "Save as Note" button saves the latest AI response, while individual [Save] links next to each message provider finer control (i.e., saves only that specific AI response, along with the conversation history leading up to it)
image
image
image
image

- A setting has been added for the highlight color used for new notes
image

- The UI, in addition to already displaying the number of API calls, now also provides an estimated session cost; `llm.py` now leverages a MODEL_COSTS dictionary containing per-million-token pricing (input and output) for various popular AI models, reflecting their cost structures on OpenRouter as of 2025-08-20.
image
image


- The book's title & author are now sent to the LLM for more relevant responses (`llm.py`)
image



Also, a good portion of time was dedicated to highlight creation & selection signals. I figured I note the following for future reference just in case:

- Creating a highlight + adding a note in two separate steps was prone to race conditions -> implemented a state machine in `ui.py` in which system now enters a pending_note_for_next_highlight state; thus, when the highlights_changed signal arrives (confirming the highlight was created in the JavaScript runtmie), the system identifies the new highlight by its UUID (universally unique identifier) and attaches the pending note
- The selection_changed signal from `web_view.py `(calibre\src\calibre\gui2\viewer\web_view.py) was unreliable in my testing; sometimes it sent incomplete data (e.g., just a UUID string instead of a full data object which should have contained the selected text, note, color, and other highlight metadata) when clicking an existing highlight -> made selected_text_changed (in `lookup.py`) intercept these inconsistencies: if only a UUID arrives, it actively queries ui.py's master highlights list to retrieve the full data object, effectively "healing" the signal at its earliest point.


Ideas I had for future enhancements:

- A user allowing the LLM to process the entire current "section" (however this may be defined) or chapter as context -> could be offered as checkbox/separate action - would allow powerful queries like "Given the events of X, what is Y's primary motivation?"
- LLM's response being streamed into the text box token-by-token (instead of waiting for the entire reply to finish) - leads to a more interactive and "live" feeling; I think this may contrast with Calibre's style, however, which seems to be more sober and deliberate in its presentation
- AI identifying key terms, names, and/or places that, when clicked, could trigger a search for _that_ term throughout the rest of the book (or even the user's entire Calibre library)
- On the easier side (I think): action to automatically create question-and-answer pairs from the selected text and export them as an Anki-compatible .csv file (massive value-add for students/researchers) + allow users to generate summary for entire chapter without needing to manually select text (one-click "create chapter summary" button)
- Ability to leverage multi-modal model to, for instance, right click an image/diagram in a book and ask an LLM to explain it, describe what is depicted, convert its data into a text-based summary, etc.,


Thank you again Kovid (and Eli) for your time and guidance. I look forward to your feedback and shall be ready to make any necessary adjustments. A beautiful day to you both!


*P.S. The screenshots use text from a Bible translation purely for demonstration. Its dense, cross-referenced nature made it an ideal candidate for testing text analysis and annotation features.*

kovidgoyal:
Looks good, I will probably review and merge (assuming no major issues)
next week.

kovidgoyal:
This is now merged into the llm branch where I will be refactoring it. Once I feel it's ready it will be merged into master. Thanks for your contribution. After it is merged into master we can discuss further enhancements.

amirthfultehrani:
Dear @kovidgoyal, thank you for both the update and the good news! I shall be awaiting for your next message.

May you have a wonderful day and rest of your week, Kovid!

kovidgoyal:
This is now merged into master, I have completely refactored it, hopefully improving it considerably. It currently supports OpenRouter and Gemini backends but has a plugin architecture for adding more backends in the future. It also renders the responses from the AI much more nicely, rendering the Markdown, showing the results as they are generated using streaming and allowing users to see the reasoning process of the AI.

amirthfultehrani:
What fantastic news, @kovidgoyal! Thank you very much for taking the time to not just consider my LLM feature request, but also in having taken your time to build a production-ready, improved version of it!

The improvements you made - especially the plugin architecture for backends + streaming Markdown responses, are awesome and beyond what I had initially built. I truly appreciate you taking the concept and building it into a robust feature.

If there is anything I can do to help with testing or adding new backends in the future (related to this or anything else), do please let me know!

I am very excited to see this in Calibre - thank you again Kovid, and a wonderful day to you!

kovidgoyal:
You are welcome! One thing that would help me is if you (or anyone)
could finish the OpenAI backend. For some reason OpenAI is failing to
verify me so I cant access their advanced models for testing and
development, via their API. Alternatively if you trust me enough, email
me a verified OpenAI API key I can use for development. In particular
the web search and multi-turn chat features need some developing.
OpenAI has a very different API than the other backend, something they
call "Responses API".

Note that if you do decide to work on the backend, I would appreciate if
you could commit to maintaining it in the future, as there are likely
going to be new APIs added to all backends to support things like
generating covers, finding what to read next, TTS, grammar and style fixing
in the editor and possibly metadata download.

Currently I am working on adding an OLLAMA backend plugin, once that is
done I think this feature is mostly ready.

Another thing that would help is testing of the feature, across
different models/backends, possibly improving the system prompt.
For instance, maybe add some instructions to the model to produce output
in Markdown. If it works well, then the heuristics currently used to detect markdown
become unneccessary and we can unconditionally use it.

margaretjoanmiller:
Disgusting.

PR #2843, opened 2025-08-17
Enhancement 2120768: up/down arrows for setting metadata edit order

cbhaley:
I changed all the row move buttons in Preferences to use the Shift (move 5), Ctrl (move 10), and Ctrl-Shirt (move to end). This involved a fair bit of refactoring. It is possible that I missed the buttons somewhere.


cbhaley:
Note also that the shortcut Ctrl-Arrow is now generalize across preferences, moving the item up or down.

All of this makes the rest of preferences have the same behavior as add your own columns.

PR #2846, opened 2025-08-21
[pdf] Add classes to display different headers and footers on the first pages of top level sections

manvithn:
In most printed books, it's common to have an empty header on the first page of each chapter, where the chapter title is the focus. This diff introduces the classes `first-section-page` and `not-first-section-page` to support this functionality.

### DIFF
--- a/manual/conversion.rst
+++ b/manual/conversion.rst
@@ -871,6 +871,12 @@ header template will show the title on odd pages and the author on even pages::
         <div class="odd-page"><i>_TITLE_</i></div>
     </header>
 
+In addition there are some more classes you can use in the headers and footers,
+documented below:
+
+  * ``first-section-page`` - include only on the first page of top level sections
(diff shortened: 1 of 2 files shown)


kovidgoyal:
You can do this already using javascript and _TOP_LEVEL_SECTION_PAGENUM_

Something like
```html
<div id="page1">...</div><div id="page2+">...</div>
<script>document.getElementById(_TOP_LEVEL_SECTION_PAGENUM_ !== 1 ? "page1" : "page2+").style.display="none";</script>
```

manvithn:
Ah, you're right, I was able to get the same behavior with:
```
<header>
  <div class="not-section-first-page">...</div>
  <div class="section-first-page">...</div>
  <script>document.currentScript.parentNode.querySelectorAll(_TOP_LEVEL_SECTION_PAGENUM_ === 1 ? ".not-section-first-page" : ".section-first-page").forEach(el => { el.style.display = "none"; });</script>
</header>
```

PR #2854, opened 2025-08-31
update economist recipes

unkn0w7n:
Increased download delay to 2 seconds to mitigate throttling.


unkn0w7n:
i dont understand what adding `canonicalize internal links` does differently than just using the `resolve internal links`.
resolve internal links seems to works as expected, as i use it in most of my recipes.
should i use canonicalize with every recipe when i add resolve links?

kovidgoyal:
You only need canoncalize if the internal links need to be modified to match article urls, see the docs of resolve_internal_url.

kovidgoyal:
I dont see nay throttling with the current delay of 1 lets wait and see if we get any more reports of that before making the recipe half as fast for everyone.

PR #2855, opened 2025-09-04
Allow using templates to define custom tooltips for the books list. …

cbhaley:
…The action is on the column context menu. It might be good to add a "calibre action" to do the same thing so the it can be put on book context menu etc.


cbhaley:
The risk of this change is low because the code path(s) don't change if no custom tooltip exists.

PR #2865, opened 2025-09-15
Fix for problem described in …

cbhaley:
… https://www.mobileread.com/forums/showthread.php?t=369898, can't sort by custom columns


cbhaley:
Added a commit to fix the help text.

PR #2868, opened 2025-09-17
Add support of pre-commit & calibre developement toolkit docs

un-pogaz:
Add support of [pre-commit](https://pre-commit.com).
This required to edit `setup/check.py` to support passing a file(s) as argument and checking only them.

The section "Pre-commit Checks" could be renamed to "Extended Workflow Checks" in the future if you append some other tools into the dev optional-dependencies.

Additionally, I have added a section "calibre developement toolkit" in the manual for the use of `setup.py`.

PS: I am not very confident in the quality and comprehensibility of my writing of the manual part, so yeah double check it, thanks.

### DIFF
--- a/manual/develop.rst
+++ b/manual/develop.rst
@@ -453,6 +453,16 @@ win64                   Build the 64-bit windows calibre installer
 xwin                    Install the Windows headers for cross compilation
 ====================    ========
 
+Pre-commit Checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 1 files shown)


kovidgoyal:
Why do we need a third party tool for this? Look at the existing git commit hooks in setup/git_*hook.py.

You can create another one that does check and symlink it into .git/hooks. No need for a tool. 

Regarding the manual changes, two comments:

1) setup.py is meant to be run by system python not calibre-debug as calibre-debug may not yet exist.

2) No need to reproduce the full docs of setup.py --help there. Just say that run ./setup.py --help for assistance. As setup.py changes keeping the manual docs in sync will not happen.

un-pogaz:
Okay, I will look to see how install you own pre-commit hook, add a assotiated sub-command in the setup.py, and extend the git_pre_commit_hook.py if is not too reinvented the wheel.

PR #2877, opened 2025-09-28
Add the possibility of having a library description shown in the …

cbhaley:
…title bar using a new option in Look & feel / Main interface. Also small clean ups of that tab's layout.

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -629,6 +629,7 @@ class DB:
         defs['edit_metadata_ignore_display_order'] = False
         defs['fts_enabled'] = False
         defs['column_tooltip_templates'] = {}
+        defs['library_description'] = ''
 
         # Migrate the bool tristate tweak
         defs['bools_are_tristate'] = \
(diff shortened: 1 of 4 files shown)


cbhaley:
The use case: it can be difficult to encode the purpose of the library in a file name, especially when name lengths are limited as on windows.

kovidgoyal:
Having the setting for this in Preferences->Look & feel seems rather
inappropriate to me. Instead it should be in the switch/create library
dialog and the rename this library dialog. And if we are going to have
library descriptions then people will want them to show up in the quick
switch and library switch menus, after all the whole rationale for this
is that short library names are insufficient to distinguish between
libraries. Distinguishing between libraries is a principal need in
library picking contexts.

cbhaley:
You're right, the description should go in all the places where one can change libraries, including the content server. I can't think of a way to do this reliably without requiring opening the db whenever the description is needed, which is a very bad idea. Caching should help but then the descriptions would be config dir specific, which isn't good. I'm closing this. If I think of some way to do it I will do another PR.

PR #2879, opened 2025-09-30
add --with-library to calibre-debug

un-pogaz:
Add support of `--with-library` to calibre-debug, can be handy to have the full log in the console and the safety to run on a specific library.


kovidgoyal:
calibre-debug -g -- --with-library /whatever

un-pogaz:
oh, the subtility of optparse.
I thought that `--` only work for scripts `-e`. I would recommend adding such statement in the manual because I do not find it to be the most obvious.

PR #2882, opened 2025-10-01
fix liberation.fr

unkn0w7n:


kovidgoyal:
There seem to be some unrelated comment only changes to nytimes and wsj?

unkn0w7n:
yes, the user agent might be useful in the future for nytimes if archive bot header doesn't work, thought i'd save it to the recipe.
WSJ doesnt work anymore but the code contains data that can fetch TTS content at least.

PR #3008, opened 2026-02-03
Add OAuth 2.0 support for email sending

shpala:
This adds OAuth 2.0 authentication for Gmail and Outlook, making it easier to set up email sending without dealing with app passwords or "less secure apps" settings.

**Why**
My mother uses calibre to send books to her Kindle, but she's not technical and has struggled with configuring SMTP credentials. OAuth simplifies this to just clicking "Authorize" and logging into her Google account. Our whole family has been using calibre for years with various e-readers and we're really happy this project exists - wanted to contribute something back.

**What's included**
- OAuth 2.0 with PKCE flow for Gmail (Outlook support ready but disabled)
- XOAUTH2 SMTP authentication
- Simple setup dialog that auto-detects provider from email domain
- For custom domains the UI lets the user select the provider manually. 
- Automatic token refresh

**For the maintainer**
I've registered a Google OAuth app for testing, but the client ID and secret in the code are only for demonstration. You should replace them with your own:

**Gmail setup:**
1. Go to https://console.cloud.google.com/
2. Create a project (or use existing)
3. Enable Gmail API
4. Go to Credentials → Create Credentials → OAuth client ID
5. Select "Desktop app"
6. Copy client ID and secret into GmailOAuth2Provider in oauth2.py

**Outlook setup (optional):**
1. Go to https://portal.azure.com/ -> App registrations
2. New registration, select "Accounts in any organizational directory and personal Microsoft accounts"
3. Add redirect URI: http://localhost:8080/oauth2callback (Web type)
4. Copy Application (client) ID into OutlookOAuth2Provider

Alternatively, delete the hardcoded values and users can set CALIBRE_GMAIL_CLIENT_ID, CALIBRE_GMAIL_CLIENT_SECRET, and CALIBRE_OUTLOOK_CLIENT_ID environment variables.

Please let me know if you want me to make any changes?

### DIFF
--- a/src/calibre/gui2/dialogs/oauth_reauth.py
+++ b/src/calibre/gui2/dialogs/oauth_reauth.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+# License: GPLv3 Copyright: 2026, Kovid Goyal <kovid at kovidgoyal.net>
+
+
+from qt.core import QDialog, QDialogButtonBox, QGridLayout, QIcon, QLabel, QPushButton, QSizePolicy, Qt
+
+from calibre.utils.resources import get_image_path as I
+
+
+class OAuthReauthMessage(QDialog):
+
+    def __init__(self, parent=None, title=None, provider=None):
+        QDialog.__init__(self, parent)
+        self.gui = parent
+        self.setWindowTitle(_('Email Authorization Required'))
+        self.setWindowIcon(QIcon(I('mail.png')))
+
+        l = QGridLayout(self)
+        self.setLayout(l)
+
+        la = QLabel()
+        la.setPixmap(QIcon(I('dialog_warning.png')).pixmap(64, 64))
+        la.setSizePolicy(QSizePolicy.Policy.Fixed, QSizePolicy.Policy.Fixed)
+        l.addWidget(la, 0, 0, Qt.AlignmentFlag.AlignTop)
+
+        pname = {'gmail': 'Google', 'outlook': 'Microsoft'}.get(provider, provider or 'your email provider')
+        msg = _(
+            '<h3>Email Authorization Expired</h3>'
+            '<p>Your {0} authorization has expired or been revoked. '
+            'This can happen if you changed your password, revoked access, '
+            'or the authorization expired due to inactivity.</p>'
+            '<p>Click <b>Re-authorize</b> to set up email again.</p>'
+        ).format(pname)
+        if title:
+            msg = _('<p>Failed to email: {0}</p>').format(title) + msg
+
+        self.msg = QLabel(msg)
+        self.msg.setWordWrap(True)
+        l.addWidget(self.msg, 0, 1)
+
+        bb = QDialogButtonBox()
+        self.reauth_btn = QPushButton(_('&Re-authorize'))
+        self.reauth_btn.setIcon(QIcon(I('config.png')))
+        self.reauth_btn.clicked.connect(self.do_reauth)
+        bb.addButton(self.reauth_btn, QDialogButtonBox.ButtonRole.AcceptRole)
+        bb.addButton(QDialogButtonBox.StandardButton.Close)
+        bb.rejected.connect(self.reject)
+        l.addWidget(bb, 1, 0, 1, 2)
+
+        self.resize(450, 250)
+
+    def do_reauth(self):
+        self.accept()
+        if self.gui is not None:
+            try:
+                self.gui.iactions['Preferences'].do_config(
+                    initial_plugin=('Sharing', 'Email'), close_after_initial=True)
+            except Exception:
+                pass
+
+
+if __name__ == '__main__':
+    from calibre.gui2 import Application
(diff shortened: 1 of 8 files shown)


shpala:
Almost forgot to mention, the scope I've set in "my" google cloud app, is ".../auth/gmail.send" only, which is what I recommend as it only allows sending an email without allowing reading emails associated to the user account

jwallrab:
This is a great PR, and it could have an even broader impact if it were generalized to enable OIDC for both email and user authentication.

First, make the OIDC endpoint generic to support any ID provider (e.g., google, outlook, keycloak, etc.):
- CALIBRE_ISSUER_URL
- CALIBRE_CLIENT_ID
- CALIBRE_CLIENT_SECRET

Then, enable an OIDC option for:
- Email authentication
- User login to the server (instead of username/password)

shpala:
@jwallrab Thanks for the feedback!
A generic OIDC provider using discovery (/.well-known/openid-configuration) is a good idea and I can look into that, but I think this can be added later on in a separate PR.
One complication for email: SMTP scopes are provider-specific and not part of the OIDC standard:
- Gmail: https://mail.google.com/
- Outlook: https://outlook.office.com/SMTP.Send
- Keycloak/generic: no standard SMTP scope (at least not one I know of)

So for email auth, we'd either need to keep some provider-specific knowledge or let users configure the scope manually.

Furthermore, if the problem is that a user might use a custom domain which actually is handled by Gmail or Microsoft, then there's a simple workaround for that. 
If the UI in current PR, won't recognize that it's supposed to be handled by a specific provider in case of a custom domain, then it will let the user manually select the right one.

kovidgoyal:
My big problem with OAuth is it requires periodic re-authentication, which can only be done in a browser. This means it is unsuitable for background tasks like sending email, where there may be no browser available. From glancing through your code I see you just fail the job in this case. Which will be utterly mysterious to most people. At the least, raise a special exception type in this case and popup a dedicated dialog explaining what needs to be done when this error occurs. Similar to what happens when the user tries to convert a DRMed book, see the job_exception() method for where this is done.

shpala:
@kovidgoyal Thanks for the fast response and for pointing out the right place in the code.
I've added code to handle authentication errors to be handled in a way similar to DRMed book conversion attempt error handling.

kovidgoyal:
Cool, will review this in detail when I have some time, unfortunately am
utterly swamped at the moment.

shpala:
Thank you!.
I'm building a qemu build environment for a windows build.
Will create a build off this branch and send it to my mom. She's got a new batch of books she wants to send to her kindle.
She'll be the first alpha-tester :)
So far Linux version works fine for me.

kovidgoyal:
OK, I reviewed the code and merged and pushed some changes to the oauth branch. But major concerns remain. 

1. gmail and outlook require client id and client secrets. This cant work with calibre since it is open source and runs on end user computers. There is no way to have a secret there. As usual with these horrible web companies their APIs are only intended to work with webapps with server backends not desktop software.

2. testing this with gmail gave me errors about the application needing verification. I am absolutely not going to be doing a verification dance with these silly email providers. 

3. This relies on patching calibre's private smtplib. This code is not used/maintained. You need to implement this using stdlib smtplib. You should use subclassing/monkeypatching for this with minimal changes.

4. the gmail create token flow complained about a missing privacy policy, screenshot attached. 
screenshot

5. You have some secrets of your google app client in the source code. I have deleted them in a new commit in my branch, but I suggest you switch to my branch, squash all commits so that the secret is not present and then do further work. I will be deleting my oauth branch in a few days anyway to remove the secret.

CDBridger:
I know this PR is a little bit old but I would much prefer oauth2 method than app password method as well so I was hoping this would get to a mergeable point. That being said I'm slightly confused about some of the points you made (I'm genuinely hoping for clarification @kovidgoyal ).

1. "gmail and outlook require client id and client secrets. This cant work with calibre since it is open source and runs on end user computers. There is no way to have a secret there. As usual with these horrible web companies their APIs are only intended to work with webapps with server backends not desktop software." - Doesn't having an app password (or any secret) necessitate the same thing? The only issue here is that this PR has a default isn't it? (which it absolutely shouldn't, each user should be setting their own secrets as some sort of env variable). Also @shpala you need to deal with this ASAP. 

2. ( and 4.) "testing this with gmail gave me errors about the application needing verification. I am absolutely not going to be doing a verification dance with these silly email providers." This is up to the user, to setup their own google console application and configuring it correctly, nothing to do with this PR. It just so happened that the google project/application was set up not fully configured no? Same with 4. I don't think this PR is suggesting there is a global calibre google app that everyone uses, it suggesting users set up their own and point calibre at it (just as you can point calibre at your own gmail account with an app password).

kovidgoyal:
Setting up a google app and getting it verified is waaaaaaaay harder than simply using a password with a non broken SMTP provider. I am not comfortable with a feature that expects users to jump through that many hoops just to send an email.

PR #3086, opened 2026-04-13
[codex] Add OpenAI-compatible AI provider

lazydao:
This change adds a new built-in `OpenAI compatible` AI provider for calibre's chat features and follows through on the integration issues found while testing it on Windows.

The user-facing problem is that the built-in AI provider list currently covers only a small set of services. Many gateways, self-hosted deployments, and third-party vendors expose an OpenAI-compatible chat API, but calibre had no generic provider for that shape of service. That meant users could not plug those services into the existing AI conversation UI without writing a dedicated provider.

The root cause was that calibre only shipped provider implementations for a few named services. In addition, the initial pass on the generic provider exposed a couple of integration gaps during real usage: on Windows, `run-local` did not reuse the installed calibre runtime and therefore failed to import required binary extensions such as `winutil`; the first version of the provider also mishandled missing custom headers and over-normalized base URLs, which broke services such as Volcengine Ark that use `/api/v3` rather than `/v1`; and the chat UI re-rendered unfinished markdown on every streamed chunk, which caused visible jitter while responses were arriving.

This patch adds a new built-in provider under `src/calibre/ai/openai_compatible/` with a configuration UI for base URL, API key, optional headers, model selection, timeout, and temperature. The backend uses the OpenAI-style `GET /models` and `POST /chat/completions` flow for text chat, while keeping URL handling generic enough to work with standard `/v1` servers and providers like Volcengine Ark. The provider is registered in the built-in provider list so it appears alongside the existing AI services.

The patch also updates `run-local` on Windows to delegate to an installed calibre runtime while setting `CALIBRE_DEVELOP_FROM`, matching the documented Windows development workflow. That keeps source changes active while still loading the compiled runtime dependencies required by the application. Finally, the AI conversation UI now throttles streamed re-renders and shows in-progress output as plain pre-wrapped text until completion, which removes most of the visible shaking during streaming without changing the final rendered response.

Validation was done with targeted checks. I ran `python -m py_compile` across the modified Python files, exercised `run-local` through `calibredb --version` and `calibre-debug -c ...` on Windows, verified that the new provider is discoverable from the built-in plugin list, checked the Volcengine Ark URL normalization path, and ran the new backend unit tests for the provider logic in isolation. I did not run a full end-to-end GUI automation pass, but the feature was manually iterated against real local usage reports until the provider could be configured and used successfully.

### DIFF
--- a/run-local
+++ b/run-local
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 import importlib
 import os
+import shutil
+import subprocess
 import sys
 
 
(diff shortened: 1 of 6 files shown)


kovidgoyal:
Split this into 3 separate PRs for run-local, throttling and the generic provider.

PR #3088, opened 2026-04-14
[codex] Fix Windows run-local runtime resolution

lazydao:
This change fixes `run-local` on Windows so that a source checkout can be exercised against an installed calibre runtime, which is the workflow documented by calibre itself.

The user-facing problem is that `python run-local calibre` and related entry points failed immediately on Windows with import errors such as `cannot import name 'winutil' from 'calibre_extensions'`. In practice that made it impossible to use the local checkout for testing changes on Windows, including GUI and AI provider work.

The root cause is that the previous `run-local` script assumed a source-style runtime on every platform. On Windows, however, calibre depends on compiled runtime extensions such as `winutil.pyd` that come from the installed application bundle rather than from the source tree. The old script pointed Python at the checkout and never delegated to the installed executables, so the import machinery could not resolve the required binary modules.

This patch updates `run-local` to detect Windows, set `CALIBRE_DEVELOP_FROM` to the checkout's `src` directory, locate an installed calibre runtime, and then invoke the corresponding installed executable for the requested entry point. It supports explicit overrides via `CALIBRE_RUNTIME_PATH` or `CALIBRE_WINDOWS_INSTALL_ROOT`, falls back to executables on `PATH`, and finally tries the default `C:\Program Files\Calibre2` install root. Non-Windows behavior is unchanged.

Validation was targeted. I ran `python run-local calibredb --version` on Windows from the source checkout and verified that it now delegates successfully to the installed runtime while loading Python code from the checkout.

### DIFF
--- a/run-local
+++ b/run-local
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 import importlib
 import os
+import shutil
+import subprocess
 import sys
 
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This is LLM nonsense. The correct and documented way to develop on
windows is CALIBRE_DEVELOP_FROM. run-local is basically a private script
I use myself on Linux.

https://manual.calibre-ebook.com/develop.html#windows-development-environment

lazydao:
Understood. I will drop this change. For Windows development I'll stick to the documented CALIBRE_DEVELOP_FROM workflow instead of changing run-local.

PR #3089, opened 2026-04-14
[codex] Throttle AI chat streaming re-renders

lazydao:
This change smooths the AI chat experience while a response is streaming by reducing unnecessary re-render churn in the conversation view.

The user-facing problem is that streamed responses visibly shook or jittered as text arrived. The effect was especially noticeable with markdown-like content, because partially emitted formatting markers caused the browser content to keep reflowing while the response was still incomplete.

The root cause is that the chat UI rebuilt and re-rendered the streamed response too aggressively. The in-progress assistant message was repeatedly converted through the normal display path and then pushed through a full HTML re-render on every small update. That meant the widget was constantly laying out unstable markup and resetting the displayed content as the stream advanced.

This patch makes two focused changes in `src/calibre/gui2/llm.py`. First, it renders in-progress streamed content as escaped pre-wrapped text instead of running unfinished output through the normal markdown-oriented display conversion. Second, it adds a short single-shot timer so intermediate streamed updates are coalesced rather than repainting the full conversation on every chunk. Completed responses still go through the existing final rendering path, so the stable end result is unchanged.

Validation was targeted. I ran `python -m py_compile src\calibre\gui2\llm.py` and loaded the updated module through the installed calibre runtime with `CALIBRE_DEVELOP_FROM` set to the checkout, confirming the new helper and import path load correctly on Windows.

### DIFF
--- a/src/calibre/gui2/chat_widget.py
+++ b/src/calibre/gui2/chat_widget.py
@@ -11,6 +11,7 @@ from qt.core import (
     QFrame,
     QHBoxLayout,
     QIcon,
+    QLabel,
     QPalette,
     QSize,
     QSizePolicy,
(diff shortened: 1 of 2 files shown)


kovidgoyal:
This analysis makes no sense to me. Partial markdown formatting might
cause issues at the trailing edge of content, it wont cause flicker
overall and Qt's HTML rendering is certianly fast enough to deal with
streaming updates.

Create a fake content injection loop for the widget that demonstrates
the issue and whether your fix works.

lazydao:
Good point. I changed the patch to keep the normal markdown rendering during streaming and focus only on batching the full-document re-renders.

I also added a fake content injection loop in `src/calibre/gui2/llm.py` via `develop(auto_close=True)`. It runs the same synthetic streaming response through two widgets side-by-side: one with immediate re-rendering on every chunk and one with the 50 ms batching in this PR.

On my Windows machine I ran:

`calibre-debug -c "from calibre.gui2.llm import develop; print(develop(auto_close=True))"`

and got:

`immediate=788 throttled=202`

So the markdown conversion itself is not the point of the change. The measurable difference is the number of full `setHtml()`/scroll cycles while the content is growing.

lazydao:
I found another contributor to the visible bottom-edge jumping: the disclaimer text at the end of the chat view was part of the scrolling HTML document, so every full re-render and scroll-to-bottom was anchoring on that block as well.

I moved the disclaimer out of the QTextBrowser HTML and into a fixed label below the browser. That keeps the streamed content area and the scroll anchor focused on the answer body only.

The fake streaming demo still shows the same immediate-render count, but the batched variant dropped from 202 full renders to 172 after this change.

PR #3091, opened 2026-04-14
Clean from source try catch

un-pogaz:
clean some remaining "running from source" and "pyton import" try catch


un-pogaz:
Hmm, Arch test fail because `tzlocal.get_localzone_name()` return a None, but only on Arch. Other CI work fine.
Will try to figure out how to fix on my side, but I don't consider the PR to fail since it seem more a edge case that calibre don't handle correctly than a strict "import failure" that the PR target.
EDIT: Fixed by adding a `or 'UTC'`

PR #3094, opened 2026-04-16
fixed Read Aloud highlights skips a word causing tts to show end of c…

C4N-6:
https://bugs.launchpad.net/calibre/+bug/2148341

the macos tts expects for the inputed text to be ascii

### DIFF
--- a/src/calibre/gui2/tts/qt.py
+++ b/src/calibre/gui2/tts/qt.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # License: GPLv3 Copyright: 2024, Kovid Goyal <kovid at kovidgoyal.net>
 
+from unicodedata import normalize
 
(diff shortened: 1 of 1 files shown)


C4N-6:
I noticed that my orignal code will still output a UTF-8 character if the character was unable to be converted to utf-8, I have now fixed this and it should ignore any utf-8 characters that are left after being normalized

kovidgoyal:
Your code is not converting to ASCII, it's converting to unicode normal
form and stripping out accent characters and other combining characters
that dont have a combined normal form.

converting to normal form is acceptable but stripping out combining
characters is not. Generally speaking the Qt TTS backends are severely
broken when it comes to Unicode, look at the _saying_word function for
details, I recommend using the default Piper engine instead.

C4N-6:
thank you this was good enough for me, and should fix problems I had with the TTS in the futer.

PR #3101, opened 2026-04-21
Fix security vulnerabilities and code quality issues

collaed:
## Summary

Code review identified several security vulnerabilities and quality issues. This PR fixes the high and medium severity findings.

### High Severity Fixes

1. **`srv/content.py:309`** — Typo `normapth` → `normpath` in the `/reader-background/` endpoint. This caused an `AttributeError` on every request, making the endpoint completely non-functional.

2. **`library/catalogs/epub_mobi.py:357,368`** — Replaced `eval()` with `ast.literal_eval()` for parsing `--prefix-rules` and `--exclusion-rules` CLI options. `eval()` allows arbitrary code execution; `ast.literal_eval()` safely parses Python literals only.

3. **`gui2/__init__.py:908`** — `FunctionDispatcher.dispatch()` silently swallowed all exceptions, returning `None` to the caller with no way to distinguish a crash from a legitimate `None` return. Now logs the traceback via `traceback.print_exc()`.

### Medium Severity Fixes

4. **`ebooks/oeb/base.py` — DirContainer path traversal** — `read()`, `write()`, and `exists()` did not validate that resolved paths stay within `self.rootdir`. A crafted ebook with `../../` paths could read/write arbitrary files. Added `os.path.abspath()` + `startswith()` checks.

5. **`gui2/comments_editor.py:200`** — XPath injection in `merge_contiguous_links()`. Href values from the document were interpolated into XPath via f-string. A double-quote in an href would break the query. Fixed by using lxml XPath variables (`$h` parameter).

6. **`library/database2.py:110`** — SQL injection in `library_id` setter. UUID was interpolated into SQL via f-string inside `executescript()`. Replaced with parameterized `execute()` calls.

7. **`utils/serialize.py:122`** — Added safety comment to `pickle_loads()` documenting that it should only be used for calibre's own serialized data.

### What was tested

- Verified lxml XPath variable syntax works correctly
- All changes are minimal and targeted — no refactoring or behavior changes beyond the fixes

### DIFF
--- a/src/calibre/ebooks/oeb/base.py
+++ b/src/calibre/ebooks/oeb/base.py
@@ -595,12 +595,16 @@ class DirContainer:
     def read(self, path):
         if path is None:
             path = self.opfname
-        path = os.path.join(self.rootdir, self._unquote(path))
+        path = os.path.abspath(os.path.join(self.rootdir, self._unquote(path)))
+        if not path.startswith(os.path.abspath(self.rootdir)):
+            raise ValueError(f'Path {path!r} is not inside {self.rootdir!r}')
         with open(path, 'rb') as f:
             return f.read()
 
     def write(self, path, data):
-        path = os.path.join(self.rootdir, self._unquote(path))
+        path = os.path.abspath(os.path.join(self.rootdir, self._unquote(path)))
+        if not path.startswith(os.path.abspath(self.rootdir)):
+            raise ValueError(f'Path {path!r} is not inside {self.rootdir!r}')
         dir = os.path.dirname(path)
         if not os.path.isdir(dir):
             os.makedirs(dir)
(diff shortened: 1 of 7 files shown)


kovidgoyal:
None of these are security issues, with the possible exclusion of
DirContainer, though IIRC DirContainer isnt meant to be a security
boundary, it relies on other parts of the conversion pipeline for that.
That said, I dont see any harm in hardening it.

Note that database2.py is only used for testing and eval() is used on
trusted data only.

qykth-git:
@kovidgoyal 
I have planed backport this fix to older Debian package.
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1136055
* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1136058

Do you have any plans to publish [Calibre GHSA](https://github.com/kovidgoyal/calibre/security) (or CVE etc.) about this issue?

kovidgoyal:
No, as I said I dont really consider these security issues. DirCaontainer is the only one that *might* be one, but given its fixed anyway I am not that fussed about it. That said if you want a security advisory for credit feel free to make one for the dircontainer issue and I will publish it.

PR #3134, opened 2026-05-18
bbc_news: emit <h1> with article title when none survives

claybdavis:
The JSON-success path in `parse_article_json` already emits `<h1>` from `article['headline']`. But when `parse_raw_html` fails to find the `window.__INITIAL_DATA__` sentinel it silently falls through to raw HTML, and `keep_only_tags = [dict(name='article')]` often leaves no `<h1>` in body — so those articles ship with no visible title in the EPUB.

`preprocess_html` now checks for a body `<h1>`; if none, it injects one from the page `<title>` (stripping the trailing `" - BBC News"` / `" - BBC Sport"` / `" - BBC"` suffix). JSON-success articles are unaffected (the guard sees the existing `<h1>` and no-ops).

### DIFF
--- a/recipes/bbc.recipe
+++ b/recipes/bbc.recipe
@@ -330,6 +330,22 @@ class BBCNews(BasicNewsRecipe):
             placeholder.decompose()
         for img in soup.findAll('img'):
             img.attrs = {'src': img.get('src', '')}
+        # If no <h1> made it through (JSON parse fell back to raw HTML, or
+        # headline was empty), inject one from <title> so the article opens
+        # with a visible heading rather than starting at the byline/body.
+        body = soup.find('body')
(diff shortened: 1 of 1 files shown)


claybdavis:
Withdrawing — this doesn't fix the actual current bug. After testing across BBC's section types: News/Future/Culture migrated to Next.js and the source `<h1>` is already preserved by `keep_only_tags`, so the injection here is a no-op. The real missing-headline issue is in Sport articles, where BBC still emits `window.__INITIAL_DATA__` but restructured the JSON so `article['headline']` is empty and the headline lives in a `headline` block-type that the recipe's parse loop doesn't handle. Will open a properly-scoped Sport fix instead.

PR #3147, opened 2026-05-22
pages count: support fixed-layout

un-pogaz:
If the ePub is a fixed-layout, use the html count as proper page count.

Adding a `fixed_layout_data()` to retrive the various data for a fixed-layout rendering.
The 'original-resolution' property seem not official, but could be usefull if commonly use.

I also change the algorithm version, but I let you decide if is a soft change (only on request by user) or a hard one (forced recount the all library).
Probably not worst it a hard for such niche case.


kovidgoyal:
Yeah I dont think this requires a forced rescan. Fixed layout EPUB is
generally not supported in calibre anyway.

PR #3148, opened 2026-05-23
Protect rooted file paths from traversal

M-Hassan-Raza:
This adds a small shared helper for keeping user-provided paths inside an expected root, then uses it for served local files and extra-file operations.

It rejects absolute paths, parent-directory traversal, drive/UNC paths, and sibling-directory escapes without adding compatibility shims.

Tests cover the shared helper, DB extra-file paths, and an authenticated content-server data-file escape case.

### DIFF
--- a/src/calibre/db/backend.py
+++ b/src/calibre/db/backend.py
@@ -62,11 +62,12 @@ from calibre.utils.filenames import (
     atomic_rename,
     copyfile_using_links,
     copytree_using_links,
-    get_long_path_name,
     hardlink_file,
     is_case_sensitive,
     is_fat_filesystem,
+    is_path_inside,
     make_long_path_useable,
+    path_from_root,
     remove_dir_if_empty,
     samefile,
 )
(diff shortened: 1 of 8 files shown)


kovidgoyal:
This has performance implications. realpath() does filesystem I/O not
all the code paths you are replacing do or need filesystem I/O. For
example when checking paths inside the calibre library directory one can
assume there are no symlinks as calibre does not create them and calibre
is not designed to be robust against attacker controlled library
directories.

M-Hassan-Raza:
> This has performance implications. realpath() does filesystem I/O not all the code paths you are replacing do or need filesystem I/O. For example when checking paths inside the calibre library directory one can assume there are no symlinks as calibre does not create them and calibre is not designed to be robust against attacker controlled library directories.

You're right, thenks. I over-scoped the helper by making realpath() part of the default containment check.

I pushed an update (after rebase) that makes the shared helper lexical: abspath() + commonpath(), so it does not add filesystem I/O to the library path checks. realpath(strict=True) is now only used in is_existing_subpath(), which already had existing-path symlink semantics before this PR.

The PR still replaces the unsafe prefix checks with component-aware containment, but without adding symlink resolution to these paths.

If you're feeling good about this code now then we can proceed otherwise I'm open to closing it as well. Again, whatever you suggest we can do that :D

PR #3158, opened 2026-05-27
proposal: Firefox-based stealth fallback for automate Browser

feder-cr:
opening as draft since issues and discussions are disabled on this repo.

the current `Browser` class in `src/calibre/web/automate/browser.py` subclasses `AsyncCamoufox` directly. some target sites (newer Amazon metadata pages, some publisher product pages) flag camoufox's UA combination by default.

asking if it would be in scope to add an optional sibling class wrapping feder-cr/invisible_playwright, which drives a patched Firefox 150 binary (feder-cr/invisible_firefox, MPL-2, same license as Firefox upstream, fingerprint patches at the C++ source level). same async API surface as `AsyncCamoufox`, so `Warmup` and call sites would not change. selected only via the existing automate config when explicitly opted into.

if accepted, a follow-up PR adds the sibling class plus a config switch. issues against the backend route to feder-cr/invisible_playwright. closing this without noise if not in scope.

### DIFF
--- a/src/calibre/web/automate/INVISIBLE_FIREFOX_FALLBACK.md
+++ b/src/calibre/web/automate/INVISIBLE_FIREFOX_FALLBACK.md
@@ -1,0 +1,13 @@
+# Firefox-based stealth fallback (proposal)
+
+> Draft proposal, opened because issues and discussions are disabled on this repo.
+
+The current `Browser` class in `browser.py` subclasses `AsyncCamoufox` directly. Some target sites (notably newer Amazon metadata pages and some publisher product pages) flag camoufox's UA combination by default.
+
+This proposal asks if it would be in scope to add an optional sibling class wrapping `invisible_playwright` (https://github.com/feder-cr/invisible_playwright), which drives a patched Firefox 150 binary (https://github.com/feder-cr/invisible_firefox, MPL-2, same license as Firefox upstream, fingerprint patches at the C++ source level so no JS shims to detect).
+
+Same async API surface as `AsyncCamoufox`, so the `Warmup` and call sites would not change. Would be selected only when explicitly opted into via the existing automate config.
+
+If accepted, follow-up PR adds the sibling class plus a config switch. Issues against the backend route to feder-cr/invisible_playwright, not here.
+
+Closing this without noise if not in scope.


kovidgoyal:
In my testing camoufox works fine with amazon (as of a couple of months
ago anyway when I last tried it), why do you feel
otherwise? And does it really matter which variant of firefox is
used for this? They are all based on modifying firefox to make its
signature harder to detect.

kovidgoyal:
Oh and looking at your releases page there dont seem to be macOS builds. calibre needs all three major platform builds. Is macOS supported at all? And do you plan to have macOS binaries in the future?

feder-cr:
fair on the first question: for amazon product pages camoufox is fine today, agreed. where the C++ patch path matters more is on pages that score visitor risk via newer fingerprint suites (FingerprintJS Pro, recent Cloudflare turnstile, recaptcha v3 scoring), where the JS-injected stealth surface starts showing up as tampered. for the use case in `automate/browser.py` (metadata sites like amazon, google books, isbndb, librarything) the gap is small. positioned as a fallback option, not a replacement.

on macOS: honest answer is no macOS builds today, only linux x86_64 and windows x86_64. macOS support is on the roadmap (open issue feder-cr/invisible_playwright#5 from january, PR #7 has draft work but is not merged yet because there is no hidden-display backend on macOS).

so for calibre's cross-platform requirement the proposal is not ready today. happy to close the PR if it is blocking other reviews, and reopen once macOS builds exist if you want to revisit then.

kovidgoyal:
Yes please let me know when the macOS builds are ready then I will be
happy to have a fallback to camoufox. Until then closing this PR as
without macOS it's a no go for calibre's use case.

feder-cr:
macOS builds are up now. the latest release ships mac (arm64 + x86_64) next to the linux and windows ones, so all three majors are covered. if the camoufox fallback angle still makes sense on your side i'm happy to put together a PR that wires it in the same way, no rush. wanted to close the loop either way since you asked to be pinged.

kovidgoyal:
If that includes true macOS headless rather than with head and hidden window happy to merge your PR. hidden window is a bit too hacky for calibre.

feder-cr:
the hidden window isn't a hack so much as the whole point: a true-headless browser renders and behaves differently enough that detection can tell it apart from a real one, so for the fingerprint case we run a normal headed firefox and just keep it off-screen, which stays identical to a real user's browser. on macos there's no xvfb-style virtual display, so if calibre needs to run with no gui session at all, real headless is the only path there and you'd trade off some of that realness. i can add a real-headless mode for exactly that case so the macos build runs displayless, and you'd pick per use case. want me to wire it into the PR?

kovidgoyal:
True headless is needed on macOS and Linux. calibre has command line tools that are supposed to be able to run in contexts without a gui. On windows there is currently no headless support. Why would the presence of a GUI affect fingerprinting? In a properly implemented headless mode the set of fonts etc. should all be the same as with a GUI. At best some performance characteristics might differ.

feder-cr:
@kovidgoyal the window matters because true-headless firefox has its own fingerprint that detection keys on, so we run it headed and hidden to stay realistic. for calibre's no-gui case that tradeoff doesn't apply, you want it to just run. i'll add a real-headless mode so it runs without a display on macos and linux. when that's in, the build fits calibre's cli tools. i'll ping you on this PR.

PR #3159, opened 2026-05-27
Add browse field settings

M-Hassan-Raza:
Adds a Browse tab to settings so users can choose which browse sections appear in the sidebar.

Visibility is stored in user settings per library. The server still owns the list of valid browse fields.

### DIFF
--- a/TODO
+++ b/TODO
@@ -3,10 +3,6 @@ a virtual library used to filter the shown books. Calling it Virtual
 library will mean calibre users will be more likely to understand what
 it is.
 
-There should be a way to select which items appear under browse in the sidebar.
-A tab in settings and maybe long tap/right click (contextmenuevent)
-on the category in the side bar to have an option to hide it.
-
 The scrollbars in dark mode have very low contrast
 
 Viewer UI needs to be re styled to match overall server UI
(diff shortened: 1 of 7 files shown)


kovidgoyal:
Why are there two identical values in library init data called
browse_fields and browse_field_options? Also in the preferences tab add
a line saying control which categories are visible in the sidebar above
the list of categories. And I dont think we need the subtitle saying
"Browse all series" etc.

Finally, in both the sidebar and the preferences the category list
should be sorted to have the same order as the Tag browser in calibre,
which is controlled via Preferences->Look & feel->Tag browser. You can
pass it to the server via interface-data endpoint.

The default set of visible fields should be the ones form the calibre
setting, until the user configures it on the server.

M-Hassan-Raza:
> Why are there two identical values in library init data called browse_fields and browse_field_options? Also in the preferences tab add a line saying control which categories are visible in the sidebar above the list of categories. And I dont think we need the subtitle saying "Browse all series" etc.
> 
> Finally, in both the sidebar and the preferences the category list should be sorted to have the same order as the Tag browser in calibre, which is controlled via Preferences->Look & feel->Tag browser. You can pass it to the server via interface-data endpoint.
> 
> The default set of visible fields should be the ones form the calibre setting, until the user configures it on the server.

Can you please check now? all the comments should be resolved hopefully faithfully

PR #3164, opened 2026-05-29
Make pdf-odd-even-offset not affect the page width

karl-zylinski:
My issue: I'm using calibre to convert an epub book to a print-ready PDF. I tried to use `--pdf-odd-even-offset 10` with ebook-convert CLI in order to shift my PDF content away from the gutter (common operation in order to make the gutter a bit bigger for both even and odd pages). However, when doing that, the page size became smaller: I used `--custom-size 6x9`, but with the odd-even-offset the width became something like 5.86 inch.

My fix: These changes shift the cropbox and mediabox so that the size stays the same.

### DIFF
--- a/src/calibre/ebooks/pdf/html_writer.py
+++ b/src/calibre/ebooks/pdf/html_writer.py
@@ -1213,11 +1213,21 @@ def convert(opf_path, opts, metadata=None, output_path=None, log=default_log, co
     if opts.pdf_odd_even_offset:
         for page_num in range(1, pdf_doc.page_count() + 1):
             margins = page_margins_map[page_num - 1]
-            mult = -1 if page_num % 2 else 1
-            val = opts.pdf_odd_even_offset
-            if abs(val) < min(margins.left, margins.right):
+            val = abs(opts.pdf_odd_even_offset)
+            if val < min(margins.left, margins.right):
+                mbox = list(pdf_doc.get_page_box('MediaBox', page_num))
                 box = list(pdf_doc.get_page_box('CropBox', page_num))
-                box[0] += val * mult
+                # Expand MediaBox to the right by val
+                mbox[2] += val
+                pdf_doc.set_page_box('MediaBox', page_num, *mbox)
+                # Odd pages: CropBox at left side of expanded MediaBox
+                # Even pages: CropBox at right side of expanded MediaBox
+                if page_num % 2:  # odd
+                    box[0] = mbox[0]
+                    box[2] = mbox[2] - val
+                else:  # even
+                    box[0] = mbox[0] + val
+                    box[2] = mbox[2]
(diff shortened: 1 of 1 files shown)


karl-zylinski:
@kovidgoyal I think I'm going about this the wrong way. It seems hard to make a fix that actually works. For now I'll just increase the `--custom-size` by however much the odd-even-offset takes away. But that aside: Do you think this should be addressed?

kovidgoyal:
I am travelling so cant look at this at the moment, but as I recall this
option simply shifts page contents, it doesnt affect the page size and
it does so by changing the CropBox, this is AFAIK the way this feature
is typically done in PDFs designed for printing.

karl-zylinski:
Thanks for the reply. It does change the page size. I set it to 6" wide using `--custom-size 6x9`. The pages end up being 5.86" when I use odd-even-offset of 10 pt (offset (10/72 = 0.138889, so 6-0.14 = 5.86).

kovidgoyal:
How are you measuring the page size? PDF has 5 different things that are
various page boxes, as I said this feature adjusts the CropBox. The
actual physical page size is set by the MediaBox and is unaffected IIRC.

karl-zylinski:
Firefox and Chrome PDF viewer says that it is 5.86x9" with `--custom-size 6x9` and `--pdf-odd-even-offset 10`. Perhaps they are looking at the "wrong" number. I'll try printing a few pages and measure physically where the edges end up.

kovidgoyal:
Browsers are not even good for one thing, browsing, let alone previewing
PDF. Use a proper PDF tool preferably something aware of print workflows
as CropBox is a print workflow feature of PDF.

karl-zylinski:
Using Adobe Acrobat Reader it says that the page width is 5.86 as well, the print dialog also gives similar numbers. I will try to research what the printer actually needs and in which contexts the different boxes pop up.

karl-zylinski:
Alright, I looked more at this now. The current way calibre does this is indeed using the cropbox. However, when uploading a PDF that uses odd-even-offset to Amazon KDP for printing, then Amazon seems to crop using the cropbox. This has the effect that the shift is in practice removed, since Amazon will then just re-center the cropped content. So this method for odd-even-offset does not work when publishing the PDF on amazon.

Perhaps this feature needs to be implemented using some per-page translation of the content rather than a shift of a box?

For now, I have worked around this by doing the even-odd pages as a post-processing step using pypdf (I also made sure to shift any annotations) and just not using the odd-even-offset setting in calibre.

kovidgoyal:
I am not that interested in catering to Amazon's limitations. As far as
I know cropbox is *the* correct way to do this. If you want to implement
it in some other fashion, a PR is welcome, adding that as an option.

PR #3174, opened 2026-06-01
Centralize viewer chrome styling

M-Hassan-Raza:
Move shared viewer chrome colors into a small helper.
Move main overlay CSS and class names into a dedicated style module.
Add RapydScript coverage for the shared styling contract.


M-Hassan-Raza:
The CI fail seems to be unrelated to my work, it looks like the usual GH actions crashing and burning. Please re-trigger actions if possible, cuz it seems like a transient error not a deterministic one

PR #3177, opened 2026-06-04
Share webui theme variable contract

M-Hassan-Raza:
Unifies webui theme variable generation so the book list and standalone viewer use the same CSS variable contract.

Light-mode primary tints now come from the saved primary color instead of reading from the DOM, and invalid primary colors fall back to the default.

Adds RapydScript tests for generated CSS, DOM-applied variables, invalid color fallback, exact light tint values, and dark mode.

Validation: git diff --check upstream/webui...HEAD; RapydScript compile for test.pyj, viewer-main.pyj, and srv.pyj.


kovidgoyal:
This is merged into master.

PR #3183, opened 2026-06-12
Support HTML Export of Book Highlights and Annotations

yogat3ch:
## Goal / Context
This PR integrates rich, interactive HTML export functionality for highlights and annotations in Calibre, allowing users to export their notes with a responsive, styled sidebar, outline navigation, and real-time filtering options. It updates both the main library annotations dialog and the standalone E-book viewer highlights dialog.

---

### 1. Library Annotations Manager (`src/calibre/gui2/library/annotations.py`)

- **HTML Generation Engine (`format_annotations_to_html`)**:
  - Implemented the conversion logic that takes raw JSON annotations and markdown formatted highlights and outputs a fully stylized HTML page.
  - Parsed headings to construct an interactive navigation sidebar outline using custom outline parsing helpers.
  - Injected responsive inline CSS styling that supports a collapsible sidebar (responsive layout under `900px`), CSS Grid and Flex layouts for notes, and highlight color tokens.
  - Appended client-side Javascript to handle:
    - Sidebar expansion/collapsing (and backdrop handling on mobile viewports).
    - Active heading highlighting based on scroll position using `scroll` event listeners and `offsetTop` bounds.
    - Smooth scrolling when outline links are clicked.
- **Helper Functions**:
  - **`slugify`**: Generates url-safe slugs from section headers.
  - **`get_unique_id`**: Ensures header anchors are unique to avoid collision.
  - **`generate_outline_html`**: Builds a nested semantic tree layout for the table of contents sidebar.
- **Export UI and Data Flow (`Export` class)**:
  - Added the **HTML** option to the format combobox dropdown in `setup_ui`.
  - Refactored `exported_data()` to abstract markdown generation into a local helper function `_generate_markdown()`. This allows both standard Markdown export and HTML export to leverage the same document structures.
  - In the `html` branch, the method constructs a JSON representation of the collection along with the generated Markdown, then invokes `format_annotations_to_html()`.

---

### 2. Viewer Highlights Dialog (`src/calibre/gui2/viewer/highlights.py`)

- **Export Compatibility**:
  - Modified the viewer's `Export` dialog to handle the `html` format similarly.
  - Refactored the core Markdown generation routine into a local `_generate_markdown()` function within `exported_data()`.
  - Dynamically imports `format_annotations_to_html` from the library annotations module when the user chooses to export annotations or highlights as HTML from the E-book viewer.

### DIFF
--- a/src/calibre/gui2/library/annotations.py
+++ b/src/calibre/gui2/library/annotations.py
@@ -6,7 +6,7 @@ import json
 import os
 import re
 from functools import lru_cache, partial
-from urllib.parse import quote
+from urllib.parse import quote, unquote
 
 from qt.core import (
     QAbstractItemView,
@@ -88,6 +88,14 @@ def render_highlight_as_text(hl, lines, as_markdown=False, link_prefix=None):
 
 
 def render_bookmark_as_text(b, lines, as_markdown=False, link_prefix=None):
+    """Render a bookmark as text.
+
+    Args:
+        b (dict): The bookmark data.
+        lines (list): The list to which the text lines will be appended.
+        as_markdown (bool): Whether to render as markdown.
+        link_prefix (str): The prefix for location links.
+    """
     lines.append(b['title'])
     date = render_timestamp(b['timestamp'])
     if as_markdown and link_prefix and b['pos_type'] == 'epubcfi':
@@ -100,6 +108,34 @@ def render_bookmark_as_text(b, lines, as_markdown=False, link_prefix=None):
     else:
         lines.append('───')
     lines.append('')
+
+
+def sanitize_color(color):
+    """
+    Extract the hex code from a hexadecimal color
+    Args:
+        color: A hexadecimal color string (e.g., "#rrggbb").
+
+    Returns:
+        The hexadecimal color string with the #.
+    """
+    return re.sub(r'[^a-zA-Z0-9-]', '', color)
+
+
+def color_contrasting(hex_color):
+    """
+    Generates a contrasting hexadecimal color.
+
+    Args:
+        hex_color: A hexadecimal color string (e.g., "#rrggbb").
+
+    Returns:
+        A contrasting hexadecimal color string.
+    """
+    hex_color = hex_color.lstrip("#")
+    rgb = tuple(int(hex_color[i:i+2], 16) for i in (0, 2, 4))
+    contrasting_rgb = tuple(255 - c for c in rgb)
+    return "#{:02x}{:02x}{:02x}".format(*contrasting_rgb)
 
 
 class ChapterGroup:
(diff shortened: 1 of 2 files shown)


kovidgoyal:
Some feedback:

1) Use single quotes for string literals to match the calibre codebase
convention

2) I am not sure why you are first converting the annotation to markdown
and then converting it to HTML? Why not render as HTML directly? And if
for some reason you do want to convert via markdown, why not use the
existing python markdown library that calibre includes to do the
conversion?

yogat3ch:
Hi @kovidgoyal,
Thanks for the feedback!

1. I've updated the previously submitted changes to use single quotes as per convention. 
2. a. This HTML export enhancement evolved from a small node.js application that used the Markdown and the calibre_highlights as inputs to create an HTML file. The logic parses the Markdown formatting as an overall structure for the HTML document, and then enhances it with HTML. It was just the simplest way to implement it.
   b. I don't load any python markdown libraries that calibre doesn't already include. There's the `_generate_markdown` helper function which is just a wrapper for the existing logic when the format is `md`.  I've eliminated some redundancy in the code to make that more clear. If you're referring to external library references I'm not seeing, could you be more specific?

kovidgoyal:
2b: I mean why do you have custom code to convert markdown to HTML? Why not  use the existing markdown to html conversion in calibre.

2a: This seems pretty fragile to me. If the markdown format generated changes one has to update all that custom markdown parsing code. IMO its better to go from the original parsed annotations to HTML via a render_as_html() method in ChapterGroup() That HTML can then be placed inside a larger HTML template used for your features.

yogat3ch:
Hi @kovidgoyal,
Can you direct me to where the "existing markdown to html conversion in calibre" is?
I searched for `render_as_html` and returned no results...
image
I checked out the `ChapterGroup` class and I don't see a method for rendering highlights at HTML. Perhaps this was something planned but never included, or deprecated at some point?

If the Markdown changes at some point in the future, why don't we just cross that bridge when we get there?

---
**Update**
I think I might have misread your message as indicating that there's already a method for rendering as html that I should use, but I think you're saying that I should create a `render_as_html` method somewhere [here](https://github.com/kovidgoyal/calibre/blob/add52bcf391b25cd7b560bf5ce3f870dfd1f943e/src/calibre/gui2/library/annotations.py#L88) and use that to generate the HTML output?

yogat3ch:
Hi @kovidgoyal,
I think this implementation is more along the lines of what you were suggesting. Can you take a look and let me know if this is what you meant?

Here's a walkthrough of the changes:


## Changes Made

1.  **Refactored Style Logic**:

    - Created the `get_annotation_style_classes` helper function in `src/calibre/gui2/library/annotations.py` to deduce the appropriate HTML tag (e.g., `blockquote`, `u`, `s`) and CSS classes (`bq-[color]`, `decor-[style]`) from an annotation's style dictionary.

2.  **Updated Highlight/Bookmark Renderers**:

    - Created `render_highlight_as_html` to use the new helper function for dynamic styling of highlights. It now outputs the exact HTML markup previously assembled via regex parsing.

    - Added a corresponding `render_bookmark_as_html` to handle bookmark items specifically.

3.  **Direct HTML Generation via `ChapterGroup`**:

    - Implemented `ChapterGroup.render_as_html()`, which serves as the new entry point for HTML export.

    - This method recurses through the hierarchical `ChapterGroup` tree to:

      - Automatically inject `<h{level}>` tags for chapter headings and generate an outline sidebar.

      - Render individual highlights and bookmarks in their proper hierarchical locations.

      - Assemble and inject the inline CSS `<style>` block containing dynamic styles for only the colors and decorations present in the tree.

      - Integrate the existing layout wrappers and JavaScript behavior (sidebar toggle, smooth scrolling).

4.  **Replaced Markdown Dependency**:

    - Updated `src/calibre/gui2/viewer/highlights.py` to directly construct the `ChapterGroup` tree and call `root.render_as_html(link_prefix)` when exporting as HTML.

    - Removed `format_annotations_to_html` from `src/calibre/gui2/library/annotations.py` completely since the Markdown-based workaround is no longer needed.

yogat3ch:
@kovidgoyal fantastic, excited to see this in the next release!

PR #3197, opened 2026-06-20
feat(sort): natural sort for title and series

tachibana-shin:
- Implement natural order sorting algorithm
- Use natural sort for title and series fields
- Add natord_sort_key utility in icu module

Before PR
image

After PR
image

I believe that a book manager should have a way to "organize" books accurately, rather than using the default sort function


kovidgoyal:
I have no idea what "natural sort " is. If you mean you want to detect numbers inside titles and series strings use the existing numeric_sort_key for that. A pure python implementation will never be performant enough for this. Sorting columns is extremely performance sensitive.

PR #3199, opened 2026-06-26
Improve layout of the Preferences overview

majklvi:
Wrap preference item labels and give all overview buttons a uniform width, so translated labels line up more predictably. Also make the category sections more compact and size the overview dialog to fit its contents when screen space allows.

Before:
original Screenshot 2026-06-26 094312

After:
polished Screenshot 2026-06-26 094215

### DIFF
--- a/src/calibre/gui2/preferences/main.py
+++ b/src/calibre/gui2/preferences/main.py
@@ -11,16 +11,15 @@ from collections import OrderedDict
 from functools import partial
 
 from qt.core import (
-    QApplication,
     QDialog,
     QDialogButtonBox,
     QFont,
-    QFrame,
     QHBoxLayout,
     QIcon,
     QKeySequence,
     QLabel,
     QPainter,
+    QPalette,
     QPointF,
     QPushButton,
     QScrollArea,
@@ -44,6 +43,14 @@ from calibre.gui2.dialogs.message_box import Icon
 from calibre.gui2.preferences import AbortCommit, AbortInitialize, get_plugin, init_gui
 
 ICON_SIZE = 32
+PREFERENCE_BUTTON_WIDTH = 112
+PREFERENCE_BUTTON_TEXT_WIDTH = 13
+PREFERENCES_OVERVIEW_PADDING = 12
+PREFERENCES_SCREEN_MARGIN = 80
+
+
+def wrap_preference_button_text(text):
+    return textwrap.fill(text, PREFERENCE_BUTTON_TEXT_WIDTH, break_long_words=False)
 
 
 # Title Bar {{{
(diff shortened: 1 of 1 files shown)


JSWolf:
That new Preferences UI is very ugly. Please put  it back to the way it was when it looked good.

Terisa:
Sorry, that's one of those unnecesary changes than only makes worse the original screen (at least for me).

Do you think this is really an improvement in a computer?

Preferences

It looks like it's designed for a mobile phone screen.... but it's a computer tool.

majklvi:
Thank you for your feedback.

When I created this change, I tried to make this screen polished and the icons be aligned to grid. Maybe the dialog should be automatically resized in width, so that there's no empty space in the right side. I still think this looks better than the original one, especially, when it's sized properly:

image

kovidgoyal:
The problem with resizing it to be the "correct" width is that the rest
of the preferences screens work better with a wider dialog thus the
ideal width forth ehome screen cant beused to set the width of the
dialog.

JSWolf:
Can we please go back to what we had prior to this mess?

Terisa:
I don't think this looks better, really. And based on the picture I've shown, it doesn't look like it has been tested in different setting to see how it works.

Dunhill69:
With this labeling adjustment, it seems the length of the texts translated into other languages ​​(for example, Spanish) wasn't taken into account, as the longer text appears with periods in between. I've also noticed excessive space on the right side of the window.
from macos

JSWolf:
@majklvi please pull your code. It's ugly as sin and it breaks foreign languages. Put back the code that was there. Nobody wants it to look like this except you.

majklvi:
> With this labeling adjustment, it seems the length of the texts translated into other languages ​​(for example, Spanish) wasn't taken into account, as the longer text appears with periods in between. I've also noticed excessive space on the right side of the window. 

Thanks a lot for your feedback, this definitely needs more work a polishing, I am working to make this work.

JSWolf:
Do not work on making this work. It doesn't work. It won't work. PUT IT BACK THE WAY IT WAS BECAUSE THAT'S WHAT WORKS! IT'S UGLY AND IT WON'T GET ANY BETTER!

All your doing is polishing a turd. You can't polish a turd so it's no longer a turd.

majklvi:
> Do not work on making this work. It doesn't work. It won't work. PUT IT BACK THE WAY IT WAS BECAUSE THAT'S WHAT WORKS! IT'S UGLY AND IT WON'T GET ANY BETTER!
> 
> All your doing is polishing a turd. You can't polish a turd so it's no longer a turd.

Thank you for your honest opinion. Please let the author decide if he wants to have this change in the project. I am not the project maintainer, I can't revert it in this repository. I am just doing my best to make it work for different systems, locales and GUI sizes, only after that, I am about to provide another patch.

kovidgoyal:
@majklvi just ignore JSWolf. See if you can address the genuine issues,
if not I will revert.

@JSWolf: moderate your tone or you will be blocked.

majklvi:
@kovidgoyal 

Both the items and their captions in the Preferences dialog use the space more effectively now. The captions wrapping has been improved. Please see the screenshot with Spanish settings below. Before I send the complete PR, I will also test MacOS version and few Linux environments.

Windows:
image


MacOS HiDPI:
image


MacOS standard DPI:
image

Debian Linux (trixie) with Gnome:
image

Terisa:
Yes, it looks more balanced now, thank you.

majklvi:
@Dunhill69 May I ask you what system and settings you use? I would like to test my recent modifications in that environment, thank you.

JSWolf:
It would be nice to have the lines back if you don't mind. Thanks.

Dunhill69:
> @Dunhill69¿Puedo preguntarle qué sistema y configuración utiliza? Me gustaría probar mis modificaciones recientes en ese entorno. Gracias.

The screenshot I sent is from a friend who uses macOS M5. I use Linux Mint, and what I've noticed is the excessive space on the right. The preferences screen was always the same size as all the other settings screens.

majklvi:
I've made the lines more visible and the PR is almost ready, please see the screenshots from Windows, Linux and MacOS:

image

image

image

kovidgoyal:
Looks OK to me. Please note that I have reverted your original PR so you will likely need to rebase/create a fresh PR.

PR #3201, opened 2026-06-26
Make main toolbar buttons use a uniform width

majklvi:
This proposes a small layout adjustment for the main toolbar.

While I understand and respect the original layout, I wanted to try a change similar in spirit to the Preferences overview: making the top toolbar items use a uniform width, so the GUI feels more orderly and visually aligned.

The implementation does not change the user-selected icon size and does not force text wrapping. It computes the shared width from the widest current toolbar button. IconOnly mode is preserved, and the custom Donate button keeps its existing behavior while participating in width alignment.

### DIFF
--- a/src/calibre/gui2/bars.py
+++ b/src/calibre/gui2/bars.py
@@ -20,6 +20,7 @@ from qt.core import (
     QPalette,
     QPropertyAnimation,
     QSize,
+    QSizePolicy,
     Qt,
     QTimer,
     QToolBar,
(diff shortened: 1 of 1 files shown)


kovidgoyal:
This would lead to insanely wide buttons with lots of wasted space if
the user happens to say install a plugin that has a long name for its
toolbar button. Unlike with preferences, the toolbar has to accomodate
buttons from plugins which cannot be QAed.

PR #3205, opened 2026-06-29
Allow typed AI prompts to reference selected viewer text

lazydao:
### Motivation

When text is selected in the ebook viewer, the AI panel already includes that selection when the user clicks one of the predefined quick actions. However, if the user types a free-form question directly into the AI chat box, the selected text was not included in the model context.

This makes typed prompts behave differently from quick actions even though both are initiated from the same selected viewer text.

### Description

This change allows free-form AI chat prompts in the viewer to reference the current selected text explicitly with the `{selected}` placeholder.

This matches the existing custom action behavior:

- If a typed prompt contains `{selected}`, it is replaced with the currently latched viewer selection.
- If a typed prompt does not contain `{selected}`, it is sent unchanged.
- Unknown placeholders are preserved instead of raising formatting errors.
- The viewer help text now documents `{selected}` for free-form prompts.

### Testing

- `python -m py_compile src/calibre/gui2/viewer/llm.py`
- `git diff --check`

---
> ### Description
> 
> This change makes free-form AI chat prompts include the currently latched viewer selection when they start a new conversation.
> 
> - Add `LLMPanel.add_selected_text_reference()` to append the selected text to a prompt with a clear localized label: `Quoted text from the current selection:`.
> - Override `LLMPanel.run_custom_prompt()` so typed prompts include the selected text only when starting a new conversation.
> - Do not repeat the selected text on follow-up prompts in the same conversation, since the existing conversation history already contains the initial quoted context.
> - Pass `uses_selected_text=True` only when the typed prompt actually starts a new conversation with selected-text context.
> 
> ### Testing
> 
> - `python -m py_compile src/calibre/gui2/viewer/llm.py`
> - `git diff --check`

### DIFF
--- a/src/calibre/gui2/viewer/llm.py
+++ b/src/calibre/gui2/viewer/llm.py
@@ -19,6 +19,38 @@ from calibre.utils.localization import ui_language_as_english
 from polyglot.binary import from_hex_unicode
 
 
+def selected_text_for_prompt(selected_text: str) -> str:
+    if not selected_text:
+        return ''
+    probably_has_multiple_words = len(selected_text) > 20 or ' ' in selected_text
+    what = 'Text to analyze: ' if probably_has_multiple_words else 'Word to analyze: '
+    return prompt_sep + what + selected_text
+
+
+def uses_field(template: str, field_name: str) -> bool:
+    try:
+        for _, fname, _, _ in string.Formatter().parse(template):
+            if fname == field_name:
+                return True
+    except ValueError:
+        pass
+    return False
+
+
+class SafeFormatDict(dict):
+
+    def __missing__(self, key):
+        return '{' + key + '}'
+
+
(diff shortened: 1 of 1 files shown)


kovidgoyal:
As is noted in the dialog to create custom actions if you want the selected text to be included in the prompt use the {selected} placeholder in your prompt. If you want selected text to be present in the typed in free form prompt I suggest you implement it in the same way by checking for the {selected} placeholder in the typed prompt and replacing it when present.

kovidgoyal:
Also, make a note of the use of {selected} in the help text starting with Or, type a question to the AI below, for example:

lazydao:
> Also, make a note of the use of {selected} in the help text starting with Or, type a question to the AI below, for example:

Thank you for your reply. I have attempted an iteration. Please review it.

PR #3213, opened 2026-07-09
feat(content server): Add support for content server plugins

jgoguen:
Enable support for third-party calibre plugins to register HTTP routes on the content server, using the existing @endpoint decorator for route metadata and auth-control. A new `ContentServerPlugin` base class provides a `content_server_endpoints()` function that returns decorated endpoint functions; these are discovered in `Handler.__init__()` after built-in routes are loaded, so both standalone and embedded servers pick them up. Third-party plugins are not allowed to modify existing routes, they may only add new ones. This does mean that a third-party plugin will not be able to register a route taken by a third-party plugin loaded earlier. Plugin import order will determine which plugin gets to register the route in the event of a collision.

### DIFF
--- a/src/calibre/customize/__init__.py
+++ b/src/calibre/customize/__init__.py
@@ -894,3 +894,23 @@ class AIProviderPlugin(Plugin):  # {{{
             return model_id
         return self.builtin_live_module.human_readable_model_name(model_id)
 # }}}
+
+
+class ContentServerPlugin(Plugin):  # {{{
+
+    type = _('Content server')
+    supported_platforms = ['windows', 'osx', 'linux']
+    # minimum version when support for this plugin type was added
+    minimum_calibre_version = (9, 12, 0)
+
+    def content_server_endpoints(self):
+        '''
+        Return endpoint functions decorated with @endpoint().
+        These are registered with the content server's router at startup.
+
+        The default returns an empty tuple. Override to provide routes.
+        Endpoints with auth_required=False will be publicly accessible.
+        Endpoints with a route matching an existing route will not be added.
+        '''
+        return ()
+# }}}
--- a/src/calibre/customize/ui.py
+++ b/src/calibre/customize/ui.py
@@ -14,6 +14,7 @@ from calibre.constants import DEBUG, ismacos, numeric_version, system_plugins_lo
 from calibre.customize import (
     AIProviderPlugin,
     CatalogPlugin,
+    ContentServerPlugin,
     EditBookToolPlugin,
     FileTypePlugin,
     InterfaceActionBase,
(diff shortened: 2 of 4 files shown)


kovidgoyal:
The doc string for the contentserverplugin class says it replaces
builtin routes. That should presumably be changed. And maybe add some
text there detailing the actual behavior.

Also, the function to iterate over plugins of the type should be moved
into customize/ui.py to match how other plugin classes are handled.

Finally, what's the use case you have in mind for this?

jgoguen:
The use case I have in mind is a plugin to let KOReader sync progress to/from calibre. I can already use calibre's OPDS feed to get books into KOReader, it would be better if I could also do progress sync but that seems like it might be too specific to put in calibre core.

kovidgoyal:
This requires doing what exactly? Getting and setting values(s) from
custom column(s)?

jgoguen:
There will be two custom columns, a float for percentage progress and text for XPath location. 

The more interesting problem will actually be computing the partial MD5 checksums KOReader sync uses and mapping them back to a book ID. A calibre book might change after it's added to KOReader, which would change its partial MD5, but you would expect progress to properly sync anyway, so that probably means a table of `(book_id, format, hash)` triplets where the whole triplet is the unique key. I'm digging around to see what I might be able to hook into without changes that would be specific to this use case, and I'm expecting to need to keep a separate DB for that but I'll happily create a custom table in the main DB if I can safely do that so I can have a foreign key constraint and cascade delete on the book ID.

kovidgoyal:
I meant from the perspective of the content server routes what's needed.
If it is something generic like get/set custom column values I am happy
with just adding a route for that to core.

jgoguen:
Ahh, on the backend it's as simple as read/write custom columns, plus the extra bit to calculate partial MD5 checksums in the background, but the KOReader sync protocol expects specific HTTP endpoints under a URL prefix:

- POST /users/create (create a user, which for this I would say we disable and use the content server user DB instead)
- GET /users/auth  (verify credentials)
- GET /syncs/progress/{partial_md5} (get the progress recorded on the server for the book)
- PUT /syncs/progress (send the device's progress to the server, which includes a device name and either XPath selector for formats like ePub or a float percentage for formats like PDF)

So I would decorate a function with `@endpoint("/kosync/users/auth", auth_required=False, postprocess=json)` and the KOReader base URL would be `https://calibre.example.net/kosync`.

Its auth  is different from what calibre does too. The endpoints will need to declare themselves as "auth disabled" and get the username from the `X-Auth-User` header and the MD5 hash of the password from the `X-Auth-Key` header. I think I saw that calibre stores the user password in the DB in plaintext so that's more of an annoyance but also needs a plugin README note that you must enable TLS in calibre or put it behind a TLS terminating reverse proxy.

kovidgoyal:
OK make the changes I requested and I will merge.

PR #3214, opened 2026-07-09
Add `__builtins__.pyi`

un-pogaz:
This PR attempt to add a `__builtins__.pyi` file, used by type checker to provide additional builtins.

This PR also extend the ruff check to the PYI files.

Revert the commit 9f746298 since their is no longuer need to explicitly import to avoid spurious errors from type checkers. The second revert was some similar things on the same period.


kovidgoyal:
But wont this cause type checker to ignore the rest of the python builtins? Or is it additive? Also I think I actually prefer to have explicit imports, it makes making queries like all files with user visible strings easier.

un-pogaz:
No it additive, it don't overide python builtins declaration. That all the interest of this file.

And I can drop the two revert if you want.

un-pogaz:
Ah, I had forgotten to parsing pyi files into setup.py check

un-pogaz:
drop it.
If you prefer to have explicit imports, their is no longuer need to use a `__builtins__.pyi`
Will open a other PR for setup check specific.

PR #3220, opened 2026-07-13
Fix Economist weekly edition recipes

lazydao:
## What

- Load the current weekly edition index from The Economist's existing GraphQL endpoint.
- Support date-specific weekly editions through `findEditionByDate`.
- Keep the regular and free Economist recipes in sync.

## Why

The weekly edition web page is now protected by a Cloudflare managed challenge, so the recipe cannot read `__NEXT_DATA__` and reports that no articles were found. Changing the browser user agent does not get past that challenge.

The recipes already use The Economist's GraphQL endpoint to download article bodies. Using the same endpoint for edition metadata avoids the challenged web index and provides the cover, sections, and article list directly.

## Impact

- Restores the default digital-edition download path.
- Preserves the existing optional web-edition behavior.
- Keeps support for downloading an edition by date.

## Checks

- Compiled both recipe files successfully.
- Downloaded the current edition in test mode with both recipes.
- Downloaded the 2026-07-04 edition through the date-specific option.
- Downloaded the complete 2026-07-11 edition: 20 sections, 76 articles, no failed-download placeholders.

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -282,13 +282,21 @@ class Economist(BasicNewsRecipe):
 
     def publication_date(self):
         edition_date = self.recipe_specific_options.get('date')
+        if isinstance(edition_date, str):
+            edition_date = edition_date.strip()
+        if edition_date:
+            return datetime.strptime(edition_date, '%Y-%m-%d').replace(tzinfo=local_tz)
         try:
-            if not isinstance(edition_date, str):
-                edition_date = None
-            edition = self.economist_get_edition(edition_date)
+            url = self.browser.open('https://www.economist.com/printedition').geturl()
+            date_str = url.split('/')[-1]
+            return datetime.strptime(date_str, '%Y-%m-%d').replace(tzinfo=local_tz)
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


kovidgoyal:
The current recipe does not actually use graphql to fetch articles, see
from_web precisely because the eco servers tend to block graphql queries

The current parse_index() also works fine other than close to Fridays
when the blocking gets more stringent presumably because of the large
number of requests.

If you do want to use graphql for parse_index() then I suggest keeping
the existing code and using both methods, trying one the the other to see
which succeeds.

lazydao:
Thanks. I updated the PR to preserve the existing website index parser (including its UA retry) as the primary path and use the GraphQL edition query only when the website parser returns no index or raises an exception. The publication-date lookup follows the same website-first fallback, and the existing `from_web` / `de` article behavior is unchanged.

I verified both routing branches with a small local harness. In a live run the website failed with both UAs, then the GraphQL fallback successfully produced both the current edition and the date-specific 2026-07-04 edition. A complete current-edition run produced 20 sections and 76 articles with no failed-download placeholders. Both Economist recipes were tested and remain identical.

PR #3230, opened 2026-07-20
Add AirDop context menu option on MacOS for sharing books

PeterWarrington:
Pull request adding an AirDrop context menu option on MacOS.

Screenshots:

Context Menu screenshot

Choose Format screenshot

AirDrop modal screenshot

### DIFF
--- a/src/calibre/customize/builtins.py
+++ b/src/calibre/customize/builtins.py
@@ -10,7 +10,7 @@ from calibre.ai.lm_studio import LMStudioAI
 from calibre.ai.ollama import OllamaAI
 from calibre.ai.open_router import OpenRouterAI
 from calibre.ai.openai_compatible import OpenAICompatible
-from calibre.constants import numeric_version
+from calibre.constants import ismacos, numeric_version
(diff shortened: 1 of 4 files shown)


kovidgoyal:
You should not be sharing paths within the calibre library folder. Instead export the formats to PersistentTemporaryDirectory() (and update the metadata in them while exporting) and then share those exported copies. This is how all export from calibre tools work such as save to disk, send to device, content server, etc.

PeterWarrington:
> You should not be sharing paths within the calibre library folder. Instead export the formats to PersistentTemporaryDirectory() (and update the metadata in them while exporting) and then share those exported copies. This is how all export from calibre tools work such as save to disk, send to device, content server, etc.

Thank you, I now export books to a PersistentTemporaryDirectory with a progress dialog before the airdrop_share call - I presume shutil.copy2 also copies the metadata.

kovidgoyal:
Note I have changed the export code to actually update metadata in the shared book files, so please test.

PeterWarrington:
Thank you for your effort fixing my slightly dodgy code! Lovely to see this merged.

PR #3241, opened 2026-07-24
Run CI format checker in last

un-pogaz:
Run the CI format checker in last, after Test calibre.
This ensure to not skip calibre tests because simply of some format error that don't impact on code execution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,9 +73,6 @@ jobs:
             - name: Run linter
               run: ./setup.py check
 
-            - name: Run format checker
-              run: ./setup.py fmt --check-only
-
             - name: Test calibre
               env:
                 PYTHONWARNINGS: error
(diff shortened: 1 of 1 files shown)


kovidgoyal:
The reason all the quality checks run first is so that one does not have
to wait for the full test suite to finish to get a failure. quality
chekcs are much faster than the test suite. The tests arent being
skipped, they simply wont run until the quality issues are fix first.

un-pogaz:
Agree for the ruff check and type check, but this PR is only about the *formating*. Even if it important, I think it's better to run the calibre tests rather to don't run them for something that, by nature, trully don't impact on code execution.
Interrupting the CI at this step could mask much more serious and real errors.

I think it really matter to ensure the calibre tests because it quite hard to execute them localy, when the CI github workflows is garanted.

kovidgoyal:
You can run calibre tests locally with

calibre-debug -t all

PR #3248, opened 2026-07-29
update econ and others

unkn0w7n:


unkn0w7n:
looks like this work around got blocked too, worked for me yesterday and today.
`This website is using a security service to protect itself from online attacks. The action you just performed triggered the security solution. There are several actions that could trigger this block including submitting a certain word or phrase, a SQL command or malformed data`

PR #3249, opened 2026-07-29
update econ

unkn0w7n:
although i didnt include browser_type in the process this somehow fixed the recipe.

### DIFF
--- a/recipes/economist.recipe
+++ b/recipes/economist.recipe
@@ -236,7 +236,7 @@ class Economist(BasicNewsRecipe):
     resolve_internal_links = True
 
     delay = 2
-    browser_type = 'webengine'
+    # browser_type = 'webengine'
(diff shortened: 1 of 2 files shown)


kovidgoyal:
See https://www.mobileread.com/forums/showthread.php?p=4601088#post4601088

unkn0w7n:
i think this might work because i pulled these headers and link query from the latest android application.

I keep testing and it still works. I think the _browser_type = web engine_ was used when fetching images which created the issue.. especially after it tries loading the cover image after loading parse_index.

we can update the recipe to use these headers even for fetching images.

kovidgoyal:
well no harm in trying i guess, but then dont you also need to set articles_are_obfuscated = True? Also one person doing the download doesnt trigger the block.

PR #3260, opened 2026-08-18
Allow plugin to add actions in comment editor

un-pogaz:
It a bit of nich thing, but here we go.

Context: In my plugin [Comments Cleaner](https://www.mobileread.com/forums/showthread.php?t=333861), I decided to add the action of my plugin into the toolbar of the comments editor.

(Since their is no real support, it a bit hacky)

So here a PR to a formal support for extend the actions in the comments editors.

To extend the action, the plugin must add a function to the set `EditorWidget.plugin_actions` (class attribut) that will work as generator. The resulting action (or groups of actions) will be append to the toolbar and the "Advanced" submenu.

The Note editor inherited from the same actions, it a intended behavior.

### DIFF
--- a/src/calibre/gui2/comments_editor.py
+++ b/src/calibre/gui2/comments_editor.py
@@ -378,7 +378,8 @@ class EditorWidget(QTextEdit, LineEditECM):  # {{{
 
     # allow plugins to extend the actions in the toolbar and the "Advanced" submenu
     # each function take the EditorWidget and the QMenu as arguments and return a single or list of QAction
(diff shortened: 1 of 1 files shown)


kovidgoyal:
You should also allow plugins to add separators, maybe by using None
instead of a QAction

un-pogaz:
Good idea. I don' flaten `plugin_actions_items` because we still need to separet the behavior betwen the toolbar and the menu.
If we were to add a separator by default between each group of actions (with most of them that will probaly contain only one), it would quickly look messy. So, it to the plugin to determin if their want to add separator to the menu to create visualy a groupe.

PR #3265, opened 2026-08-22
The Golfer's Journal Recipe Updates

ben-ng:
Good day @kovidgoyal. I thought I created #3264 in draft mode and was surprised when it was merged yesterday. Sorry about the slip-up, I wasn't done testing it on all issues yet, and discovered that it didn't work on issues 26 and earlier. This pull request fixes that, and makes a bunch of other quality-of-life fixes.

**Support older issues**: Issues No. 26 and earlier use the site's original layout (linked feature-story blocks instead of a toc-block), and Nos. 27-28 have a table of contents with no links at all, requiring the article URLs to be constructed from the titles' WordPress slugs. With these fallbacks the full back catalogue from No. 1 onwards is downloadable. Verified against all 32 currently published issues.

**Keep issue covers consistent**: Adds special-casing to handle the inconsistencies in how issue covers look between issues so that they are always the same aspect ratio. Also add a highly visible issue number in the lower right corner so the reader can pick out the issue they want from a thumbnail.

**Streamline formatting**: Removes UI elements that don't work or make sense on an e-reader, like Light/Dark mode toggles, so only the content remains. Adds a working table of contents so readers can jump to the article they care about more quickly.

### DIFF
--- a/recipes/golfers_journal.recipe
+++ b/recipes/golfers_journal.recipe
@@ -11,6 +11,8 @@ contents.
 """
 
 import re
+from contextlib import closing
+from io import BytesIO
 
 from calibre.web.feeds.news import BasicNewsRecipe, classes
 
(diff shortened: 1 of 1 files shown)


kovidgoyal:
no worries, its my bad I didnt notice the draft status (I most interact
with github over email)

PR #3268, opened 2026-08-29
docs: fix typo definately -> definitely

vaibhav8a:
Corrects the misspelling `definately` -> `definitely` in `src/calibre/devices/mtp/unix/upstream/music-players.h`.

Documentation only, no behaviour change.


vaibhav8a:
Closing this — my mistake. `src/calibre/devices/mtp/unix/upstream/` holds files vendored from libmtp (separate copyright, kept under `upstream/` so they can be re-synced), so a typo fix here would be reverted on the next sync and belongs upstream in libmtp instead. Sorry for the noise.

PR #3282, opened 2026-09-17
Add a "grayscale jpeg" option in the Compress Images window of the ebook-editor

bechapte36:
Adds a "grayscale jpeg" option in the Compress Images window of the ebook-editor.

This is useful for
A) documents with jpegs that should be grayscale but are not
B) readers with a grayscale display (dropping colors on-the-fly during the export would be even better in that case)

From the "jpegtran -grayscale" documentation:
The luminance channel is preserved exactly
This switch is particularly handy for fixing a monochrome picture that was mistakenly encoded as a color JPEG. (In such a case, the space savings from getting rid of the near-empty chroma channels won't be large; but the decoding time for a grayscale JPEG is substantially less than that for a color JPEG.)

It was trivial to implement, and worked, in version 8. I cannot test it in the current version as I fail to build that (not related to this pull request).
Judging by the code base it should work flawlessly though.

It is possible that the color won't be dropped because the new jpeg is not smaller then the old one, mostly due to arithmetic coding.
That can be handled that in an additional pull request if wanted (Using PIL.Image.open(filename).mode).


bechapte36:
Please squash the commits during merge, as I could not do that in the web UI. Neither editing multiple files in a single commit. Sorry for that.

kovidgoyal:
I'm afraid this is too niche a use case for core calibre but the editor
has a plugin framework you can use to create a simple plugin to do this
for you.

